Issue #25, opened 2026-01-27
[Explorer] Collector Pipeline


jaydeluca:
See parent issue #23  for more information / context.

## Notes

Collector components have different schema than Java Agent (no target_versions,
different status structure, distribution/type taxonomy).

Collector has multi-dimensional classification:
- **Distributions**: core, contrib, third-party
- **Component Types**: receiver, processor, exporter, extension, connector

## Requirements

- Implement Collector registry reader
  - Read from `ecosystem-registry/collector/{version}/`
  - Parse metadata.yaml format from each component type directory
  - Handle receivers/, processors/, exporters/, extensions/, connectors/
  - Extract README/markdown content from component directories
  - Extract distribution and type information
- Implement Collector transformation logic
  - Map registry data to output schema
  - Include distribution and type metadata in each component
- Implement Collector output writer
  - Create `collector/` directory structure
  - Write content-addressed component files
  - Write content-addressed markdown files
  - Generate version manifests with optional markdown references
  - Generate index.json with taxonomy summary
  - Generate versions.json

Ensure all component types are handled

Issue #24, opened 2026-01-27
[Explorer] Java Pipeline


jaydeluca:
See parent issue #23  for more information / context.

## Requirements

- Implement Java Agent registry reader
  - Read from `ecosystem-registry/java/instrumentation/{version}/`
  - Parse javaagent.yaml format
  - Extract README/markdown content from source paths
- Implement Java Agent transformation logic
  - Map registry data to output schema
  - Preserve all existing fields
  - Handle instrumentation-specific metadata (target_versions, configurations, etc.)
- Implement Java Agent output writer
  - Create `javaagent/` directory structure
  - Write content-addressed component files
  - Write content-addressed markdown files
  - Generate version manifests with optional markdown references
  - Generate index.json (flat list of instrumentations)
  - Generate versions.json

Issue #23, opened 2026-01-27
Content-Addressed Data Transformation Pipeline


jaydeluca:
The Ecosystem Explorer (website) will use a [content-addressed storage pattern](https://en.wikipedia.org/wiki/Content-addressable_storage) to store versioned metadata. Since this data changes infrequently and often has large overlaps between versions, content addressing enables significant space savings and efficient caching.

We may evolve this strategy in the future, but this allows us to generate a low overhead, static file "database" to power the site.

## Objective

Implement an `ecosystem-automation` component called `explorer-db-builder` that:

1. Reads raw metadata from `ecosystem-registry`
2. Transforms it into content-addressed format
3. Outputs transformed data directly into `ecosystem-explorer` static site directory

This mirrors the POC's [data-processing-v2/](https://github.com/jaydeluca/instrumentation-explorer/tree/main/data-processing-v2) pipeline which generated data into `frontend/public/data/`.

## Components

```
ecosystem-registry/          # Raw metadata (inventory only)
  java/
    javaagent/
      v2.12.0/
        instrumentation.yaml
      ...

ecosystem-automation/
  explorer-data-transform/   # Transformation pipeline (this component)
    - Reads from ecosystem-registry
    - Applies content addressing
    - Generates manifests and indexes
    - Writes to ecosystem-explorer

ecosystem-explorer/
  public/data/               # Content-addressed output (static site data)
    javaagent/
      index.json
      versions.json
      versions/
        2.24.0.json
      components/
        akka-http-10.0-737fb17f9652.json
      markdown/
        aws-sdk-1.11-48c8b39bee75.md
    collector/
      index.json
      versions.json
      versions/
        0.95.0.json
      components/
        otlp-abc123def456.json
      markdown/
        ...
```

### Output Storage Structure

The transformation pipeline outputs to `ecosystem-explorer/public/data/`:

```
ecosystem-explorer/
  public/
    data/
      javaagent/
        index.json            # Lightweight index for javaagent (browsing/search)
        versions.json         # List of available javaagent versions
        versions/
          2.24.0.json        # Version manifest: {component-id: content-hash}
          2.23.0.json
          ...
        components/
          akka-http-10.0-737fb17f9652.json
          aws-sdk-1.11-48c8b39bee75.json
          ...
        markdown/
          aws-sdk-1.11-48c8b39bee75.md    # Content-addressed READMEs
          ...

      collector/
        index.json            # Lightweight index for collector
[...]

PR #52, opened 2026-02-04
Content addressed database - Java agent

jaydeluca:
Sorry for the massive PR 😬 , there's a lot of files changed, but 246 of them are just data files.

This is part of #23 
Completes #24
Related to #53

It takes the registry data, converts it to json, and writes it into a content-addressed database in the explorer project. We hash the contents of the instrumentation metadata and then for each release of the java agent we create an index pointing to the relevant entry by hash. Since most of the metadata doesn't change release to release, it keeps our overall overhead pretty low.

This will eventually run nightly or be triggered after a new release is detected.

On the frontend we will load thing as needed and use [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API) to cache it for performance. Down the line we will likely re-evaluate this approach, but this will get us started and is easy to manage for a static site without us needing to maintain a backend.

The implementation focuses on the java agent, but I will make it a bit more generalized when I tackle #25

------

Running locally `uv run explorer-db-builder --clean`:

```
============================================================
Explorer DB Builder
============================================================

Cleaning database directory: ecosystem-explorer/public/data/javaagent
Database directory cleaned
Processing 1 release versions
Processing Java Agent version: 2.24.0
Found 244 libraries
Wrote version index for 2.24.0 with 244 instrumentations
Wrote version list with 1 versions (latest: 2.24.0)

Database Statistics:
  Files written: 246
  Total size: 649,662 bytes (0.62 MB)

✓ Database build completed successfully
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,6 @@ node_modules/
 # Collector Watcher temporary repositories
 tmp_repos/
 
+metadata-issues.md
+
 *.DS_Store
--- a/ecosystem-automation/README.md
+++ b/ecosystem-automation/README.md
@@ -8,6 +8,7 @@ Automation tools for the OpenTelemetry Ecosystem Explorer project.
 ## Components
 
 - **collector-watcher**: Collects and aggregates metadata from OpenTelemetry Collector components
+- **java-instrumentation-watcher**: Collects and aggregates metadata from the OpenTelemetry Java Instrumentation project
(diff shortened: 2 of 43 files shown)


jaydeluca:
> Wow. RSLGTM on this one 😆 

Yes it's a bit abstract, once all the pieces are in place I'll do a live overview/review during a comms meeting. I'm also working on some documentation as reference as well

PR #85, opened 2026-02-12
Java Agent Instrumentation List page

jaydeluca:
Adds the agent list page where you can search and filter the instrumentations. Will be adding onto this with things like semantic conventions, but will handle that in a followup.

image


ecosystem-explorer/src/features/java-agent/components/instrumentation-card.tsx
```diff
@@ -0,0 +1,94 @@
+import type { InstrumentationData } from "@/types/javaagent";
```
jaydeluca: Preview of this component:

image

vitorvasc (approved):

PR #105, opened 2026-02-21
feat(explorer): group instrumentations by display name on list page

### LINKED ISSUE #98: [Explorer] Group instrumentations on instrumentation list
We have many instrumentations that are for the same library, but might target different versions or artifacts. In the POC we grouped them together on the list page, which you could click to expand and it would then provide links to the individual "sub" instrumentations:

Image

Image

Requirements:

- Instrumentations that have the same display name should be grouped into a combined card on the list page
- When the card is clicked to expand, it provides sub-cards / links to the individual instrumentation page

s1v4-d:
Instrumentations sharing the same display name are now grouped into expandable cards on the Java Agent instrumentation list page.

- Add groupInstrumentationsByDisplayName utility that groups and sorts instrumentations by their resolved display name
- Add InstrumentationGroupCard component with expand/collapse behavior, aggregated badges, version count, and sub-item links
- Integrate grouping into JavaInstrumentationListPage using a filter-then-group approach so filters apply before grouping
- Singletons render as regular InstrumentationCard (no visual regression)
- Add comprehensive tests for grouping utility (10) and group card (17)

Closes #98

### DIFF
--- a/ecosystem-explorer/src/features/java-agent/components/instrumentation-badges.test.tsx
+++ b/ecosystem-explorer/src/features/java-agent/components/instrumentation-badges.test.tsx
@@ -40,6 +40,7 @@ describe("TargetBadges", () => {
     );
     expect(screen.getByText("Agent")).toBeInTheDocument();
     expect(screen.getByTitle("Java Agent")).toBeInTheDocument();
+    expect(screen.getByLabelText("Has Java Agent target")).toBeInTheDocument();
   });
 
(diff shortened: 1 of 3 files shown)


jaydeluca (approved):
Thank you @s1v4-d for your contribution. As our first external contributor, you’ve helped mark an important milestone for this project 🎉  

We are grateful for your involvement!

PR #106, opened 2026-02-22
Add contributor survey

jaydeluca:
Copied from: https://github.com/open-telemetry/opentelemetry.io/blob/main/.github/workflows/survey-on-merged-pr.yml

### DIFF
--- a/.github/workflows/survey-on-merged-pr.yml
+++ b/.github/workflows/survey-on-merged-pr.yml
@@ -1,0 +1,50 @@
+name: Survey on Merged PR by Non-Member
+
+on:
+  pull_request_target:
+    types: [closed]
+
+permissions:
+  contents: read
+
+env:
+  PR_NUM: ${{ github.event.pull_request.number }}
+  SURVEY_URL:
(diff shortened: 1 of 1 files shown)


.github/workflows/survey-on-merged-pr.yml
```diff
@@ -0,0 +1,50 @@
...
+    types: [closed]
+
+permissions:
+  contents: read
+
+env:
+  PR_NUM: ${{ github.event.pull_request.number }}
+  SURVEY_URL:
+    https://docs.google.com/forms/d/e/1FAIpQLSf2FfCsW-DimeWzdQgfl0KDzT2UEAqu69_f7F2BVPSxVae1cQ/viewform?entry.1540511742=${{
+    github.repository }}
```
maryliag: for future reference, this line here make sure when the survey open is already selected to the correct repo, but that means the form itself needs to be updated to add the new entry.
I just updated the form to include this repo, so it will work correctly now, but if anyone is looking into this in the future, don't forget to add the repo as an option to the form (you can ask me or any other GC member to make the change, because the form is owned by the admin account)

PR #148, opened 2026-03-16
feat: added the seperate watcher job's for nightly-registry

### LINKED ISSUE #142: Make Nightly registry updates more resilient
Right now we run [both of the watchers in a single job](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/blob/main/.github/workflows/nightly-registry-update.yml#L68-L74), and if one of them fails, the nightly registry update won't create a PR. For example, the java file format changed and therefore we didn't get updates for the collector.

We are also seeing occasional rate limits for the github API. We should ensure we have retry logic in place, and that we are appropriately using the github token to try and avoid this happening.

example error from [this job](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/actions/runs/23094949805/job/67085781977):

```
Traceback (most recent call last):
  File "/home/runner/work/opentelemetry-ecosystem-explorer/opentelemetry-ecosystem-explorer/ecosystem-automation/java-instrumentation-watcher/src/java_instrumentation_watcher/java_instrumentation_client.py", line 64, in get_latest_release_tag
    response.raise_for_status()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/runner/work/opentelemetry-ecosystem-explorer/opentelemetry-ecosystem-explorer/.venv/lib/python3.14/site-packages/requests/models.py", line 1026, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 403 Client Error: rate limit exceeded for url: https://api.github.com/repos/open-telemetry/opentelemetry-java-instrumentation/releases/latest
```

Requirements:
- Setup workflow notifications for each of the watchers that run (collector, java) so that an issue is created separately if either fail (for example if they both failed, there would be two issues created)
- Update the nightly job so that a PR is still created as long as one of the watchers succeeds and there is a diff (don't fail the job when either fail)
- Ensure github client will retry in the case of failures on things like get_version()

pratik50:
Fixes #142

### Changes

- Split collector and java watchers into separate independent jobs so one failing does not block the other
- Pass \`GITHUB_TOKEN\` to watcher steps for authenticated API calls which will increase the Rate limit
- Create separate failure issues per watcher
- Ensure PR is created if at least one watcher succeeds

### DIFF
--- a/.github/workflows/nightly-registry-update.yml
+++ b/.github/workflows/nightly-registry-update.yml
@@ -31,7 +31,7 @@ jobs:
             echo "is_primary=false" >> $GITHUB_OUTPUT
           fi
 
-      - name: Configure git 
+      - name: Configure git (primary repository)
         if: steps.repo_check.outputs.is_primary == 'true'
         run: |
           .github/scripts/use-cla-approved-bot.sh
@@ -43,7 +43,7 @@ jobs:
           app-id: ${{ vars.OTELBOT_APP_ID }}
           private-key: ${{ secrets.OTELBOT_PRIVATE_KEY }}
 
-      - name: Configure git 
+      - name: Configure git (forked repository)
         if: steps.repo_check.outputs.is_primary == 'false'
         run: |
           git config user.name "github-actions[bot]"
--- a/.github/workflows/reusable-workflow-notification.yml
+++ b/.github/workflows/reusable-workflow-notification.yml
@@ -40,7 +40,7 @@ jobs:
             ISSUE_TITLE="Workflow failed: $GITHUB_WORKFLOW (#$GITHUB_RUN_NUMBER)"
           fi
 
-          number=$(gh issue list --search "in:title \"$ISSUE_TITLE\"" --limit 1 --json number -q .[].number)
+          number=$(gh issue list --search "in:title $ISSUE_TITLE" --limit 1 --json number -q .[].number)
 
           if [[ -n "$FAILURE_DETAILS" ]]; then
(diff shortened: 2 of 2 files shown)


.github/workflows/nightly-registry-update.yml
```diff
@@ -11,8 +11,91 @@ permissions:
   pull-requests: write
 
 jobs:
-  synchronize-inventory:
+  # Watcher 1: Collector
```
jaydeluca: what if instead of creating separate jobs and managing uploading/merging state etc, we kept everything in the same job, but used  `continue-on-error: true`  with somethign like

```
      - name: Run collector-watcher
        id: collector_watcher
        run: |
          uv run collector-watcher
        continue-on-error: true

      - name: Run java-instrumentation-watcher
        id: java_instrumentation_watcher
        if: always()
        run: |
          uv run java-instrumentation-watcher
        continue-on-error: true
        
      - name: Watcher failure notification
        if: steps.collector_watcher.outcome == 'failure'
     ......
```

This helps avoid having to upload/download/merge etc, which also adds a bunch of additional action dependencies.

I really appreciate the screenshots of the test runs you do, thanks for doing that. It would also be a good test to try emulating a failure, like by undoing your fix the other day (#139), and then running the workflow to see if it correct creates an issue

pratik50: undoing the fix from #139 to verify separate issues are created per watcher. Will try that out and let you know.

PR #217, opened 2026-04-09
Update renovate configs

jaydeluca:
The current strategy of grouping all of these updates along with requiring the 14 day wait period was resulting in some PRs hanging around for a long time, blocking other updates.

This change will split things out with some more granular rules so that we can better stay on top of and review the dependency updates, as well as reducing the wait time from 14 days to 7


chalin (approved):
RSLGTM

PR #254, opened 2026-04-16
Add YAML output generator (#178)

### LINKED ISSUE #178: Configuration builder YAML Output Generator
## Context
This converts the form state from #177 into a valid OpenTelemetry declarative
configuration YAML file that users can download and use with the Java agent.

## Requirements
- Use a YAML library that preserves comments and formatting (recommend js-yaml)
- Output must validate against OpenTelemetry configuration schema
- Keys should be sorted alphabetically for consistency
- Include helpful comments for each section
- Handle edge cases (empty sections, default values)

## Notes
- See POC implementation at #151 
- Wait until after #177 is completed before this one

lucacavenaghi97:
Closes #178

Adds `generateYaml(state, schema, options?)` in `src/lib/yaml-generator.ts`. Pure transform from `ConfigurationBuilderState` to a downloadable OpenTelemetry SDK declarative configuration YAML string.

`file_format` is pinned first via `toFileFormatVersion` (e.g., `1.0.0` -> `1.0`). Other top-level sections are emitted alphabetically with banner comments from schema labels and descriptions. Empty values are stripped recursively, with a single-entry heuristic that preserves plugin_select discriminator shorthand like `{tracecontext: null}` and `{batch: {}}`.


ecosystem-explorer/src/lib/yaml-generator.ts
```diff
@@ -0,0 +1,156 @@
...
+  return [
+    "# OpenTelemetry SDK Configuration",
+    `# Generated by Ecosystem Explorer on ${date}`,
+    `# Schema version: ${version}`,
+    "# Docs: https://opentelemetry.io/docs/specs/otel/configuration/",
+    "",
+  ].join("\n");
+}
+
+function toFileFormatVersion(version: string): string {
```
jaydeluca: I'm not sure if we should alter the version number. The intent is that we'll always point at the latest released version (so we don't need to worry about `-SNAPSHOTS` suffix), and I think it would be good to have the patch value included. What do you think?

ecosystem-explorer/src/lib/yaml-generator.test.ts
```diff
@@ -0,0 +1,446 @@
...
+
+    const defaultOutput = generateYaml(emptyState, emptySchema);
+    expect(defaultOutput).toContain("# OpenTelemetry SDK Configuration");
+    expect(defaultOutput).toContain(`# Generated by Ecosystem Explorer on ${today}`);
+    expect(defaultOutput).toContain("# Schema version: 1.0.0");
+
+    const overridden = generateYaml(emptyState, emptySchema, { header: "# custom" });
+    expect(overridden.startsWith("# custom\n")).toBe(true);
+    expect(overridden).not.toContain("# OpenTelemetry SDK Configuration");
+  });
```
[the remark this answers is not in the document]
jaydeluca: I think its probably sufficient to jus remove the assertion for the date in this test as opposed to using a fake timer, it's not that important IMO

jaydeluca:
just a few small comments, otherwise looks good!

ecosystem-explorer/src/lib/yaml-generator.ts
lucacavenaghi97: I based this on the [official SDK config example](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/otel-sdk-config.yaml) which uses `file_format: "1.0"` (major.minor only), so I assumed that was the expected format. Checked the schema though and there's no pattern constraint, any string is valid. So happy to drop the truncation and use the full version, makes the code simpler too.

jaydeluca: about 30 minutes ago i was reviewing that same exact page and noticed that too, and I found it a little strange they truncated it. I think we can remove it for now and add it back later if we decide it would be useful

lucacavenaghi97: sounds good, done in the latest commit

jaydeluca (approved):

PR #256, opened 2026-04-16
fix: snapshot cleanup runs after successful fetch and parse

### LINKED ISSUE #255: Nightly registry update should not remove files on failure
In #252 the java-instrumentation-watcher failed due to a new schema version being released. This is expected. What is not expected is that in the PR it then removed the latest SNAPSHOT data for java-instrumentation (ecosystem-registry/java/javaagent/v2.26.2-SNAPSHOT/instrumentation.yaml).

We should figure out why and ideally prevent that from happening in the future. If a watcher fails, it should be a no-op for that particular ecosystem.

lucacavenaghi97:
`cleanup_snapshots()` was called before fetch and parse in all three watchers. If parsing failed, the old snapshot was already deleted but the new one was never written, causing the nightly update to commit the deletion.

Moved cleanup to after fetch and parse succeed so a failed run is a no-op.

Fixes #255

### DIFF
--- a/ecosystem-automation/collector-watcher/src/collector_watcher/collector_sync.py
+++ b/ecosystem-automation/collector-watcher/src/collector_watcher/collector_sync.py
@@ -285,9 +285,9 @@ class CollectorSync:
         Update or create the SNAPSHOT version for a distribution.
 
         This:
-        1. Cleans up old snapshots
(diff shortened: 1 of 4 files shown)


jaydeluca (approved):
thank you @lucacavenaghi97 !

PR #274, opened 2026-04-22
feat: add feature flags and wrap java config builder on one

### LINKED ISSUE #223: Hide in-progress features behind build-time feature flags
Currently, all changes merged into `main` are automatically deployed to production at [https://explorer.opentelemetry.io/](https://explorer.opentelemetry.io). While this enables fast iteration, it also means that incomplete or experimental features can become publicly accessible.

This can lead to a confusing user experience, especially when users encounter partially implemented pages without context that they are still under development.

A concrete example is the declarative configuration builder page:

* [https://explorer.opentelemetry.io/java-agent/configuration/builder](https://explorer.opentelemetry.io/java-agent/configuration/builder)

---

### Problem

We do not currently have a mechanism to prevent unfinished features from being exposed in production.

Relying on obscurity (i.e., assuming users won’t find the URL) is not sufficient, and removing features from `main` slows down development and collaboration.

---

### Proposed Solution

Introduce **build-time feature flags using environment variables** to control the visibility of in-progress features.

Since the site is statically built and deployed via Netlify, environment variables can be used to include/exclude features at build time.

---

### Implementation Details

1. **Define environment variable(s)** in Netlify (and locally via `.env`):

   ```bash
   REACT_APP_ENABLE_BUILDER=false
   ```

2. **Gate feature routes/components** in the React application:

   ```ts
   const isBuilderEnabled = process.env.REACT_APP_ENABLE_BUILDER === "true";
   ```

3. **Conditionally render or register routes**:

   ```tsx
   {isBuilderEnabled && (
     <Route
       path="/java-agent/configuration/builder"
       element={<BuilderPage />}
     />
   )}
   ```

   Alternatively, redirect or return `null` if disabled.

4. Ensure the flag is:

   * `false` in production
   * configurable for local development and deploy previews

---

### Expected Outcome

* In-progress features can be merged into `main` without being publicly visible
* Developers can continue working and testing features locally or via preview deployments
* Production site only exposes features that are explicitly enabled
* Introduce a centralized feature flag registry instead of scattered env checks
* For the declarative configuration builder, use a “coming soon” UI instead of fully hiding the page

---

### Acceptance Criteria

* [ ] Builder page is not accessible in production when flag is disabled
* [ ] Builder page is accessible locally when flag is enabled
* [ ] Route is fully excluded or gated (no partial rendering)
* [ ] Pattern is documented for future feature flags

eruizdechavez:
Resolves #223.

## Summary

In this pull request I am adding basic support for feature flags. The idea is that features that are still in progress, or simply not ready for general availability, can be hidden behind environment variables that are read at build time by Vite.

Since feature flags work as boolean conditions, they can be used for multiple purposes, such as hiding components, disabling routes, or gating other in-progress functionality.

## Implementation details

The original issue mentioned using React's `REACT_APP_` environment variable naming convention. After some initial testing, that was not working as intended (`'process' variable is null or undefined`).

After looking through the Bun and Vite documentation, I found that the Vite way to do this is to use the `VITE_` naming convention for environment variable names and access them through `import.meta.env`.

To make the intent of these variables explicit, I decided to use the prefix `VITE_FEATURE_FLAG_`. This makes it clear that these environment variables are meant to control feature availability, rather than represent other kinds of configuration values.

To avoid leaking Vite-specific details throughout the codebase, I wrapped this logic in a small utility function, `isEnabled`. Behind the scenes, it normalizes common environment variable values such as `true`, `1`, and `yes` into a boolean. If the environment variable is not defined, Vite returns an empty string.

## Testing

To test this locally, create a `.env.local` file in `ecosystem-explorer/` and define `VITE_FEATURE_FLAG_JAVA_CONFIG_BUILDER` with either `true` or `false`.

Setting it to `true` makes the Java Config Builder visible. Setting it to `false` hides it and disables the route.

## AI

This contribution was prepared with the assistance of an AI development tool.


ecosystem-explorer/src/lib/feature-flags.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+function normalizeFlagValue(value: string): boolean {
+  return ["true", "1", "yes"].includes((value ?? "").toLowerCase());
+}
+
```
jaydeluca, line 37: what do you think about if we added some type safety / autocomplete, and create a little registry that shows the available options and gives us a place to annotate them with comments describing their utility


So for example, we could also add another one for the [collector page](https://explorer.opentelemetry.io/collector):
 
```javascript
export type FeatureFlag = "JAVA_CONFIG_BUILDER" | "COLLECTOR_PAGE";

const FLAGS: Record<FeatureFlag, string | undefined> = {
    // Declarative Configuration Builder for Java - Still in development
    JAVA_CONFIG_BUILDER: import.meta.env.VITE_FEATURE_FLAG_JAVA_CONFIG_BUILDER,
    // Collector Page - Still in development
    COLLECTOR_PAGE: import.meta.env.VITE_FEATURE_FLAG_COLLECTOR_PAGE
};

export function normalizeFlagValue(value: string | undefined): boolean {
    return ["true", "1", "yes"].includes((value ?? "").toLowerCase());
}

export function isEnabled(flag: FeatureFlag): boolean {
    return normalizeFlagValue(FLAGS[flag]);
}
```

jaydeluca (commented):
first of all, great PR! I appreciate the info you provided in the description, thank you for that

Along with the suggestion I made about the implementation, I'd love to have at least the new flag enabled by default in local and deploy previews. Could we add a `.env.development` file that enables it? And then I believe in order to handle the deployment previews we can add [deploy contexts](https://docs.netlify.com/deploy/deploy-overview/#deploy-contexts-1) to `netlify.toml`: 

```
[context.deploy-preview.environment]
  VITE_FEATURE_FLAG_JAVA_CONFIG_BUILDER = "true"
[context.branch-deploy.environment]
  VITE_FEATURE_FLAG_JAVA_CONFIG_BUILDER = "true"
```

ecosystem-explorer/src/lib/feature-flags.ts
eruizdechavez: Thanks for the suggestions!

I definitely forgot about the Netlify part. I assumed it needed some sort of admin access to define those so I was hoping to get a comment about that 😝.

Your suggestion looks A LOT like my initial version of this feature, but I decided against it in favor of reducing the amount of code someone needs to write in order to get a flag in the code and use it. With the existing approach we only need to have them available in the environment and they "just work", but the downside is that there is no autocomplete or validation in place. If you don't mind about the extra code required to "register" a flag before being able to use it I do like the idea. Let me explore a little bit and see how it feels.

jaydeluca: yea I don't think the additional overhead is too bad, and I like having a "source of truth" codified to know what options might be possible

eruizdechavez: New version of the feature flags is in!

I decided to go with a slightly different version but it serves the same purpose. In this case, instead of a dictionary with the values, I am using a simpler approach with an array of strings. The values on the array are then used (with a bit of TS gymnastics) as the only allowed values which serves all the purposes of static checking, explicit definition, and autocomplete.

I also added the collector flag and used it in the templates where I think it makes sense.

Last but not least, added the flags to local environment and the suggested netlify environments.

ecosystem-explorer/.env.development
```diff
@@ -0,0 +1,2 @@
+VITE_FEATURE_FLAG_JAVA_CONFIG_BUILDER=true
+VITE_FEATURE_FLAG_COLLECTOR_PAGE=true
```
jaydeluca, line 2: looking at the test failure, we might need to add these to the github action build step environment too: https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/blob/main/.github/workflows/build-and-test.yml#L67

eruizdechavez: done

jaydeluca (approved):
great work, thank you @eruizdechavez !

eruizdechavez:
thanks @jaydeluca, I had a great experience with this contribution, the issue description was super helpful as it had all the required context, and your comments and suggestions helped a lot. definitely looking forward to helping with more contributions ;)

jaydeluca:
@eruizdechavez in case you are interested in a small followup, I think I'd like to go back to having the collector tile on the homepage and linking to the the "Coming soon" page like we have for the java agent config, so we'll probably revert that piece for now. Figured i'd let you do it if you have interest in another PR though, no pressure either way!

image

eruizdechavez:
Sure thing, I’ll send a small pull request with that change.

Issue #269, opened 2026-04-20
SDK Configuration Schema browser updates


jaydeluca:
Some notes and followups for #265 

These can be handled separately or in batches

- We should exclude the `distribution` configs for now, we'll need to create an issue to think through how to handle that in a followup. It's much lower priority though
- The Resource config should be promoted higher, maybe as a separate tab, so we'd have - "Resource" "SDK" "Instrumentation"
  - It might also make sense to explore other ways of grouping some of the similar components - Tracer, Logger, Meter providers for example
- The vertical distribution of information feels too big. Maybe we should collapse most of them by default (aside from the getting started components)
  - Also find a way to make each section more distinct. Perhaps instead of rendering them all at once we have ways to navigate and search for various components
- I talked to one of the maintainers of the configuration schema, and he indicated that we shouldn't use the patch version of the version... so `1.0` instead of `1.0.0`.. I'm sorry that I had you remove that in a previous pr
- When we go to add a processor, you get option "Batch, Simple", "Custom". If you choose custom, the helper text of the input reads "custom-pluin-key". We should call it "name" instead of "key", to be consistent with the language from the spec. (https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk.md#register-plugincomponentprovider)
- We should handle snippets like `otlp_http: {}` as `otlp_http:` (i.e. omit the empty object)
- We need to do something about maintaining line breaks and formatting from descriptions. A description with markdown lists and line breaks between list items is smashed onto one line, which reduces readability.
- We can remove the "instrumentation" component when under the SDK navigation, we'll have all that in a separate tab

lucacavenaghi97:
Hi @jaydeluca! Could you assign this to me?

Issue #250, opened 2026-04-15
Configuration Builder - Instrumentation Browser


jaydeluca:
The Configuration Builder (#162) has two main tabs:
1. **SDK Tab** - Configure SDK-level options (exporters, samplers, propagators, etc.)
2. **Instrumentation Tab** - Select instrumentations and configure their specific options

This issue covers the Instrumentation Tab, which allows users to:
- Browse available instrumentations for a specific Java agent version
- Select which instrumentations to include in their configuration
- Configure per-instrumentation options (where available)

Implement the Instrumentation tab of the Configuration Builder with:
1. A searchable list of available instrumentations
2. Selection mechanism (add/remove instrumentations)
3. Per-instrumentation configuration options (when available)
4. Integration with the YAML output generator

## Acceptance Criteria

- Instrumentation tab shows searchable list of instrumentations
- Search filters by name and description
- Users can select/deselect instrumentations
- Selected instrumentations appear in dedicated section
- "Add All" adds all visible (filtered) instrumentations
- "Clear All" removes all selections
- Each selected instrumentation shows available config options
- Config changes reflect in state and YAML output
- Version selector updates instrumentation list

lucacavenaghi97:
Heads up from #275 (open): that PR hides the SDK schema's `instrumentation/development` section from the SDK tab, since conceptually it belongs here.

When this tab is built, it will likely need to surface two data sources together:

1. **Java agent registry** (`ecosystem-registry/java/javaagent/v<version>/instrumentation.yaml`) — the library picker described above.
2. **SDK config schema** (`ecosystem-registry/configuration/v<version>/instrumentation.yaml`) — `general.*` cross-cutting knobs (HTTP semconv, captured headers, sanitization, stability opt-in) plus `<language>.*` per-library config.

The two are bridged by the `declarative_name` field on each Java agent library's `configurations`, which maps to a path under `instrumentation/development.*` in the SDK schema. Everything feeds into the same YAML output.

PR #275, opened 2026-04-22
Collapse empty YAML values and hide Instrumentation from SDK tab

lucacavenaghi97:
Addresses items 6 and 8 from #269.

**Item 6**: `js-yaml` was emitting `otlp_http: {}` and `always_on: null` for plugin-select discriminators and empty enabled groups. Added an `emptiesToNull` walker that converts empty objects to null before dumping, combined with `styles: { "!!null": "empty" }` so both cases render as `otlp_http:`. Matches the canonical form in the OTel configuration spec examples.

**Item 8**: Hid `instrumentation/development` from the SDK tab form. The filter now uses a named `HIDDEN_SDK_KEYS` set alongside the existing `file_format` exclusion. That content will surface in the Instrumentation tab once #250 lands.


jaydeluca (approved):
✅  nice!

PR #286, opened 2026-04-23
fix: restore collector page

eruizdechavez:
## Summary

Restore the collector page and links, while keeping the feature flag in place by showing a placeholder on the collector page.

## Implementation details

In my previous pull request (#274) I added feature flags to this project, including one that disabled the collector page and its links in production.

In this pull request I am restoring the collector page and links. Instead of removing the flag entirely, I am adding a placeholder on the collector page that is only displayed when the flag is enabled, mostly to avoid removing the flag and having someone else add it back when needed.

## Testing

Tested locally by verifying:

- the collector page route is available again
- the collector links are visible again in the header and home page
- the collector page shows the expected placeholder based on whether the flag is enabled


ecosystem-explorer/src/features/collector/collector-page.tsx
```diff
@@ -1,3 +1,20 @@
+/*
```
jaydeluca, line 1: looks like the copyright header got duplicated here

eruizdechavez: 🤦‍♂️ ... wonder why the pre-commit hook duplicated it 🤔

eruizdechavez: done. had to include another fix from the pre-commit hook 🤷‍♂️ .

jaydeluca (approved):
you rock, thanks @eruizdechavez !

PR #312, opened 2026-04-27
Remove cache version assertion

jaydeluca:
We now have automation that automatically increments the cache version, so having an assertion on the version is problematic, and not really valuable.


ecosystem-explorer/src/lib/api/idb-cache.test.ts
```diff
@@ -39,7 +39,6 @@ describe("idb-cache", () => {
       const db = await initDB();
 
       expect(db.name).toBe("otel-explorer-cache");
-      expect(db.version).toBe(4);
```
jaydeluca, line 42: this is the only change in this PR, the other ones were picked up by linting

maryliag (approved):

PR #321, opened 2026-04-29
[Explorer] Use has_javaagent field for Java Agent support

### LINKED ISSUE #303: [Explorer] Use has_javaagent field
The Java Agent metadata now includes an explicit `has_javaagent` boolean field, added in file_format 0.5 by the java-instrumentation-watcher. Previously, the frontend inferred whether an instrumentation had Java Agent support by checking whether `javaagent_target_versions` was a non-empty array.

This inference is unreliable because an instrumentation can have `has_javaagent: true` with an empty `javaagent_target_versions` array (e.g., cross-cutting agent capabilities like `external-annotations` or `jmx-metrics` that don't target a specific third-party library version range).

We want to replace anywhere we are inferring this to instead just use the `has_javaagent` field.

## Requirements

* Add `has_javaagent` to `InstrumentationData`
* Both `getBadgeInfo` and `getAggregatedBadgeInfo` in `ecosystem-explorer/src/features/java-agent/utils/badge-info.ts` currently infer javaagent support from `javaagent_target_versions`. Replace with the explicit `has_javaagent` field.
* On the `java-instrumentation-list-page` the `javaagent` target filter currently infers support from `javaagent_target_versions`, so update that
* On `ecosystem-explorer/src/features/java-agent/instrumentation-detail-page.tsx` the "Requirements" section currently shows when `javaagent_target_versions` is non-empty. Replace the outer guard with `has_javaagent`, while keeping `javaagent_target_versions` for the "Target Versions" sub-card (it is still valid data worth displaying when present).
* Across the test suite, fixtures that use `javaagent_target_versions` solely to simulate "this instrumentation has javaagent support" should be updated to use `has_javaagent: true` instead.
* Bump the `DB_VERSION` in `ecosystem-explorer/src/lib/api/idb-cache.ts` to invalidate previous caches where this field might not have been available

kajal-jotwani:
The frontend was inferring Java Agent support by checking if `javaagent_target_versions` was non-empty. This breaks for
instrumentations like `external-annotations` or `jmx-metrics` that have `has_javaagent: true` but an empty versions array.

This PR switches everything over to use the `has_javaagent` boolean field that was added in file_format 0.5.

Fixes: #303 

Changes:
- Added `has_javaagent` to `InstrumentationData` type
- `getBadgeInfo` and `getAggregatedBadgeInfo` in `badge-info.ts` now use `has_javaagent` directly
- Javaagent filter in `java-instrumentation-list-page` updated
- Requirements section in `instrumentation-detail-page.tsx` now guards on `has_javaagent`; `javaagent_target_versions` kept only for the Target Versions sub-card
- Test fixtures updated to use `has_javaagent: true` instead of a non-empty versions array
- Bumped `DB_VERSION` to invalidate stale caches


jaydeluca (approved):
nice job, thanks @kajal-jotwani !

Issue #326, opened 2026-04-30
Optimize website build


jaydeluca:
when running `bun run build`, we are seeing these errors:

```
6:25:03 AM [vite] warning: `esbuild` option was specified by "vite:react-swc" plugin. This option is deprecated, please use `oxc` instead.
vite v8.0.10 building client environment for production...
✓ 1870 modules transformed.
computing gzip size...
dist/index.html                   0.48 kB │ gzip:   0.31 kB
dist/assets/index-WHKwK23V.css   74.21 kB │ gzip:  11.11 kB
dist/assets/index-CHVrgGFq.js   503.67 kB │ gzip: 148.10 kB

[plugin builtin:vite-reporter] 
(!) Some chunks are larger than 500 kB after minification. Consider:
- Using dynamic import() to code-split the application
- Use build.rolldownOptions.output.codeSplitting to improve chunking: https://rolldown.rs/reference/OutputOptions.codeSplitting
- Adjust chunk size limit for this warning via build.chunkSizeWarningLimit.
[INEFFECTIVE_DYNAMIC_IMPORT] Warning: node_modules/js-yaml/dist/js-yaml.mjs is dynamically imported by src/hooks/use-configuration-builder.ts but also statically imported by src/lib/yaml-generator.ts, dynamic import will not move module into another chunk.
```

We should look into implementing this code splitting/chunking and fix the import issue, along with any other opportunities to optimize performance

snakefood3232:
The issue involves a warning about chunks exceeding 500 kB and an ineffective dynamic import with js-yaml. I'd handle this by implementing Vite's code-splitting features more effectively and resolving the static vs dynamic import conflict in your use-configuration-builder.ts and yaml-generator.ts files. I can refactor the imports and update the build configurations to split heavy modules into separate chunks, ensuring optimal loading. Can have a PR up in 2 days. I've built scalable apps with Next.js and vite in production environments, tackling similar optimization issues before.

jaydeluca:
hey @snakefood3232 , are you still planning on working on #302 ?

aabhinavvvvvvv:
@jaydeluca I've identified root cause and created this necessary fix.
[...]

PR #337, opened 2026-05-01
perf(explorer): optimize build with code splitting and vendor chunks

aabhinavvvvvvv:
## Summary

Fixes the build warnings and 849 kB monolith bundle reported in #326.

**Root causes:**
- All page components in `App.tsx` were eagerly imported, bundling every route into a single chunk regardless of which page the user visits. This also pulled `js-yaml` and the entire config builder into the main bundle for all users.
- The `await import("js-yaml")` in `use-configuration-builder.ts` was a no-op — `yaml-generator.ts` already statically imports it, so the dynamic import could never split into its own chunk (hence the `INEFFECTIVE_DYNAMIC_IMPORT` warning).
- No vendor chunk strategy was configured in `vite.config.ts`.

**Changes:**
- `App.tsx`: convert all page imports to `React.lazy()` + `Suspense`
- `use-configuration-builder.ts`: replace ineffective dynamic import with static `import { load } from "js-yaml"`
- `vite.config.ts`: add `build.rolldownOptions.output.manualChunks` to split `react/react-dom/react-router`, `js-yaml`, `react-markdown`, and `lucide-react` into separately cached vendor chunks

## Results

| | Before | After |
|---|---|---|
| Bundle | 849 kB single chunk | 32 chunks, largest 344 kB |
| Main entry | 849 kB | 190 kB |
| `js-yaml` | loaded on every page | loaded only for config builder |
| `INEFFECTIVE_DYNAMIC_IMPORT` warning | present | gone |
| Chunk size > 500 kB warning | present | gone |

## Note
use-configuration-builder.ts shows a large diff due to line ending normalization (CRLF → LF). The only functional change is replacing await import("js-yaml") with a static import { load as loadYaml } from "js-yaml"


ecosystem-explorer/vite.config.ts
```diff
@@ -24,4 +24,31 @@ export default defineConfig({
...
     },
   },
+  build: {
+    rolldownOptions: {
+      output: {
+        manualChunks(id) {
+          if (
+            id.includes("node_modules/react") ||
+            id.includes("node_modules/react-dom") ||
+            id.includes("node_modules/react-router")
```
jaydeluca: The check `id.includes("node_modules/react")` is a substring match, so it catches `node_modules/react-markdown` before the markdown condition is ever evaluated. `react-markdown` will get bundled into `vendor-react` instead of `vendor-markdown`, and `vendor-markdown` ends up containing only `remark-gfm`

```suggestion
            id.includes("node_modules/react/") ||
            id.includes("node_modules/react-dom/") ||
            id.includes("node_modules/react-router/")
```

jaydeluca (approved):
thanks @aabhinavvvvvvv !

PR #341, opened 2026-05-02
feat(explorer): Configuration Builder instrumentation browser

lucacavenaghi97:
First of two PRs for #250. Replaces the Instrumentation tab placeholder in the Configuration Builder.

Users can:
- Search and browse the Java instrumentations for the current agent version. Search matches name and description.
- Edit cross-cutting `instrumentation/development.general.*` settings via the existing schema renderer.
- Override a specific instrumentation's enabled state. Output emits only the diff from the agent default.

Reuses the SDK tab's 3-column shell, sidebar, and `PreviewCard`. The sidebar gains a search input and an Overridden filter that appears once there's at least one override. Two new components carry the new behavior: `InstrumentationBrowser` and `InstrumentationRow`.

Small fixes that landed alongside:
- `setValueByPath` on `useConfigurationBuilder`: array-form sibling of `setValue` so paths whose segments contain dots (e.g. `cassandra-4.4`) reach the reducer without going through `parsePath`.
- `BUILDER_GRID` switched to `minmax(0, 1fr)` for the middle column. Long instrumentation IDs were expanding the column past 1fr and pushing the `PreviewCard` off-screen.
- Section-flag mirroring uses a new `hasMeaningfulLeaf` helper from `state-hydrate` so empty-but-non-null subtrees don't keep the section enabled in the YAML.

Out of scope, follow-ups:
- Per-instrumentation \`configurations[]\` form.
- Bulk actions (Override all visible / Clear overrides).
- Multi-version override preservation across version switch.
- Category filter chips.

### DIFF
--- a/ecosystem-explorer/src/features/java-agent/configuration/components/instrumentation-browser.test.tsx
+++ b/ecosystem-explorer/src/features/java-agent/configuration/components/instrumentation-browser.test.tsx
@@ -1,0 +1,242 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { render, screen, fireEvent } from "@testing-library/react";
+import type { InstrumentationData } from "@/types/javaagent";
+import type { ConfigurationBuilderState } from "@/types/configuration-builder";
+
+const setValueByPath = vi.fn();
+const setEnabled = vi.fn();
+const removeMapEntry = vi.fn();
+
+let mockBuilderState: ConfigurationBuilderState;
+
+vi.mock("@/hooks/use-configuration-builder", () => ({
+  useConfigurationBuilder: () => ({
+    state: mockBuilderState,
+    setValueByPath: (...args: unknown[]) => setValueByPath(...args),
+    setEnabled: (...args: unknown[]) => setEnabled(...args),
+    removeMapEntry: (...args: unknown[]) => removeMapEntry(...args),
+    setValue: vi.fn(),
+    selectPlugin: vi.fn(),
+    addListItem: vi.fn(),
+    removeListItem: vi.fn(),
+    setMapEntry: vi.fn(),
+    resetToDefaults: vi.fn(),
+    enableAllSections: vi.fn(),
+    loadFromYaml: vi.fn(),
+    validateField: vi.fn(),
+    validateAll: vi.fn(),
+    clearValidationError: vi.fn(),
+  }),
+}));
+
+const useInstrumentationsMock = vi.fn();
+vi.mock("@/hooks/use-javaagent-data", () => ({
+  useInstrumentations: (version: string) => useInstrumentationsMock(version),
+}));
+
+import { InstrumentationBrowser } from "./instrumentation-browser";
+
+const FIXTURE: InstrumentationData[] = [
+  {
+    name: "cassandra-4.4",
+    display_name: "Cassandra Driver",
+    description: "Database client spans + metrics for the DataStax Cassandra Driver.",
+    disabled_by_default: false,
+    scope: { name: "io.opentelemetry.cassandra-4.4" },
+  },
+  {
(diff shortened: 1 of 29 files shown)


ecosystem-explorer/src/features/java-agent/configuration/components/instrumentation-browser.test.tsx
```diff
@@ -0,0 +1,242 @@
...
+    clearValidationError: vi.fn(),
+  }),
+}));
+
+const useInstrumentationsMock = vi.fn();
+vi.mock("@/hooks/use-javaagent-data", () => ({
+  useInstrumentations: (version: string) => useInstrumentationsMock(version),
+}));
+
+import { InstrumentationBrowser } from "./instrumentation-browser";
```
jaydeluca, line 23: can we move this to the top?

ecosystem-explorer/src/features/java-agent/configuration/components/instrumentation-browser.tsx
```diff
@@ -0,0 +1,249 @@
...
+import type { ConfigValue, ConfigValues } from "@/types/configuration-builder";
+import { useInstrumentations } from "@/hooks/use-javaagent-data";
+import { useConfigurationBuilder } from "@/hooks/use-configuration-builder";
+import { SectionCardShell } from "./section-card-shell";
+import { InstrumentationRow } from "./instrumentation-row";
+
+export interface InstrumentationBrowserProps {
+  version: string;
+  search: string;
+  statusFilter: "all" | "overridden";
```
jaydeluca, line 30: No need to change it right now, but Im wondering if "overridden" is intuitive enough for people to understand vs something like "customized"... just noting so we can continue brainstorming but we can leave it as is

ecosystem-explorer/src/features/java-agent/configuration/configuration-builder-page.tsx
```diff
@@ -88,22 +118,185 @@ function SdkTabContent({ schema, starter, version, activeTab }: SdkTabContentPro
...
+    []
+  );
+  const sectionKeys = useMemo(() => tocSections.map((s) => s.key), [tocSections]);
+  const sectionsContainerRef = useRef<HTMLDivElement>(null);
+  const { activeKey, scrollToSection } = useActiveSection(sectionKeys, sectionsContainerRef);
+
+  // The SDK schema declares `instrumentation/development.java` as a
+  // `key_value_map` (opaque), so the values-tree shape we write under
+  // `["instrumentation/development", "java", <id>, "enabled"]` round-trips
+  // verbatim through the YAML generator without a schema lookup.
```
jaydeluca: the `enabled` flags (and a few others) are actually handled in a new way now and are under a `distribution` block ([reference](https://opentelemetry.io/docs/zero-code/java/agent/declarative-configuration/#mapping-of-configuration-options))

```
distribution:
  javaagent:
    instrumentation:
      default_enabled: false # was otel.instrumentation.common.default-enabled
      enabled:
        - tomcat
        - spring_webmvc
      disabled:
        - armeria_grpc
```

ecosystem-explorer/src/features/java-agent/configuration/configuration-builder-page.tsx
```diff
@@ -88,22 +118,185 @@ function SdkTabContent({ schema, starter, version, activeTab }: SdkTabContentPro
...
+  generalNode: GroupNode | null;
+}
+
+// Renders the `instrumentation/development.general.*` subtree as a card
+// without the depth=0 SwitchPill that GroupRenderer would otherwise add.
+// The SwitchPill is wrong here because its enable key is the GroupNode's
+// own `key` (`general`), not the top-level YAML section
+// `instrumentation/development` — see InstrumentationTabBody's effect for
+// the section-flag mirroring.
+function InstrumentationGeneralCard({ generalNode }: InstrumentationGeneralCardProps): JSX.Element {
```
jaydeluca, line 257: we can add this to look into as a followup, but when you expand the settings in "general settings", they aren't indented or colored differently so its hard to know if you are looking at a subsection:

image

lucacavenaghi97: Yeah, I noticed that as well. I actually tried to address it in this PR, but I wasn’t able to find a clean solution that fit well with the current layout. I agree it’s something we should definitely revisit in a follow-up.

ecosystem-explorer/src/features/java-agent/configuration/components/instrumentation-browser.tsx
lucacavenaghi97: Yes I think that "Overridden" is more precise technically (you're overriding the agent's default) but "customized" would read more naturally for less technical users

ecosystem-explorer/src/features/java-agent/configuration/configuration-builder-page.tsx
lucacavenaghi97: Good catch, I missed that one sorry about that

jaydeluca (approved):

PR #342, opened 2026-05-02
Fix: prevent crash in ConfigurationCard when description is undefined

suhr25:
### SUMMARY

Fixes a crash in **ConfigurationCard** when a config item has no `description`. Adds a simple null check before calling `renderWithInlineCode`, keeping it consistent with existing checks in the file.

**File:**

* `ecosystem-explorer/src/features/java-agent/components/configuration-card.tsx`

---

### FIX

The issue occurs because `renderWithInlineCode` assumes it always receives a string and calls `.split()`, but in some cases `config.description` is `undefined` due to incomplete runtime data. This PR adds a guard to ensure the function is only called when `description` exists, preventing the crash. The change follows the same defensive pattern already used for `config.default`, making it consistent and low-risk.

---

### CHECKLIST

* [x] Prevents crash when `description` is missing
* [x] No API or type changes
* [x] Follows existing code patterns
* [x] Small, safe change

---

###  VERIFICATION

* Ran app locally
* Tested config **with and without** `description`
* No crashes, no console errors
* Existing behavior unchanged

### DIFF
--- a/ecosystem-explorer/src/features/java-agent/components/configuration-card.tsx
+++ b/ecosystem-explorer/src/features/java-agent/components/configuration-card.tsx
@@ -76,7 +76,7 @@ export function ConfigurationCard({ config, format }: ConfigurationCardProps) {
         </div>
 
         <p className="text-muted-foreground text-sm leading-relaxed">
-          {renderWithInlineCode(config.description)}
+          {config.description && renderWithInlineCode(config.description)}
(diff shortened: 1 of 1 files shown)


jaydeluca (approved):
LGTM, thanks @suhr25 ! out of curiosity, did you encounter this in a particular instrumentation module?

PR #348, opened 2026-05-04
fix: avoid empty description rendering and make configuration.description optional

suhr25:
## SUMMARY

Fixes extra spacing in `ConfigurationCard` caused by rendering an empty `<p>` when `description` is missing, and updates the type to reflect that `description` isn’t always present.

---

## FIX

Updated `ConfigurationCard` to render the `<p>` only when `config.description` exists, avoiding empty elements in the UI. Also changed the TypeScript type from `description: string` to `description?: string` so it matches real data and enforces proper null checks.

---

## CHECKLIST

* [x] No empty `<p>` rendered
* [x] Type updated to optional
* [x] Consistent with other components
* [x] No breaking changes

---

## VERIFICATION

Open the Configuration tab and check items without a description — they should no longer show extra spacing.

### DIFF
--- a/ecosystem-explorer/src/types/javaagent.ts
+++ b/ecosystem-explorer/src/types/javaagent.ts
@@ -56,7 +56,7 @@ export interface InstrumentationScope {
 export interface Configuration {
   name: string;
   declarative_name?: string;
-  description?: string;
+  description: string;
   type: "boolean" | "string" | "list" | "map" | "int" | "double";
   default: string | boolean | number;
   example?: string[];
--- a/ecosystem-explorer/src/types/javaagent.ts
+++ b/ecosystem-explorer/src/types/javaagent.ts
@@ -56,7 +56,7 @@ export interface InstrumentationScope {
 export interface Configuration {
   name: string;
   declarative_name?: string;
-  description: string;
+  description?: string;
   type: "boolean" | "string" | "list" | "map" | "int" | "double";
   default: string | boolean | number;
   example?: string[];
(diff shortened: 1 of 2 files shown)


suhr25:
Hi @jaydeluca,
Kindly review this PR.
Thanks!

ecosystem-explorer/src/types/javaagent.ts
```diff
@@ -56,7 +56,7 @@ export interface InstrumentationScope {
 export interface Configuration {
   name: string;
   declarative_name?: string;
-  description: string;
+  description?: string;
```
jaydeluca: thanks for another contribution @suhr25 !

Configuration options should always have a description, did you find a page or configuration that didn't?

suhr25: That makes sense, thanks! I will keep `description` required and update the UI to conditionally render it to avoid empty elements.

suhr25: I have made changes accordingly , kindly review

jaydeluca: thanks @suhr25. I was interested in knowing if you encountered this situation on the site, or if you just identified this in the code and that is what motivated the idea to submit a fix? I am only asking because If we have a config option without a description, I want to go and fix it at the source too

suhr25: I didn’t encounter this in the UI, I identified it while reviewing the code.
Since the description is rendered unconditionally, I thought it could lead to empty elements if the data is ever incomplete.
So I proposed a defensive UI fix while keeping description required.
Happy to adjust if you would prefer enforcing this strictly at the data/source level instead.

jaydeluca: thanks for the clarification, this change is good as-is, i just wanted to understand the motivation for making the change

jaydeluca (approved):

suhr25:
Thanks for the merge!

PR #357, opened 2026-05-05
fix: add mobile hamburger menu to header

mohanakatari119-bit:
The header navigation was only visible on `md` and above breakpoints (`hidden ... md:flex`). On mobile there was no fallback, so the nav links were completely inaccessible to anyone on a phone or narrow viewport.

This adds a hamburger toggle button that is visible only below the `md` breakpoint. Clicking it slides open a vertical nav panel with the same links. The icon swaps between `Menu` and `X` (from lucide-react, already a dependency). The panel closes automatically when a link is tapped so the user lands on the new page without having to dismiss the menu manually.

Both the toggle button and the nav panels carry appropriate `aria-label`, `aria-expanded`, and `aria-controls` attributes. The desktop `<nav>` now also has `aria-label="Main"` which was missing.

Three new test cases cover: hamburger button presence, open/close toggle state, and auto-close on link click.

demo

| Open | Closed |
|------|--------|
|   |   | reverse the image


jaydeluca (commented):
great contribution @mohanakatari119-bit thank you for addressing this! 

The two comments from copilot are valid, i confirmed the menu stays open when you navigate without pressing a button in the opened menu, so that would be good to address. The second comment about having a unified nav-item definition would be a nice improvement too if you wouldn't mind looking into that. 

Thanks again!

mohanakatari119-bit:
> great contribution @mohanakatari119-bit thank you for addressing this!
> 
> The two comments from copilot are valid, i confirmed the menu stays open when you navigate without pressing a button in the opened menu, so that would be good to address. The second comment about having a unified nav-item definition would be a nice improvement too if you wouldn't mind looking into that.
> 
> Thanks again!

thanks :)
i will start working on the copilot suggestions

mohanakatari119-bit:
@jaydeluca hey i just fixed the suggestions, can u review it : )

vitorvasc (commented):
Looks great! Thanks for the contribution 🙌🏼 

Two minor tweaks that could improve the experience:

- The menu currently stays open even when clicking outside of it
- We could add an opaque backdrop behind the menu t keep the focus on the navigation while it's open. This would also make it easier to handle the "click outside" behavior from the previous item

Let me know what you think :)

mohanakatari119-bit:
hey @vitorvasc, both make sense! the backdrop especially would make the close-on-outside-click much cleaner to handle. i'll work on both together since they pair well. will push an update soon

vitorvasc:
Hi @mohanakatari119-bit, just checking in - are you still interested in moving forward with this contribution?

vitorvasc:
FYI: I'm moving this PR to draft state so we can better track what's ready for review.

jaydeluca:
hey @mohanakatari119-bit just wanted to check in one more time to see if you plan on finishing this PR or if we should close it and allow another contributor to pick it up?

jaydeluca (approved):
thanks @mohanakatari119-bit !

Issue #40, opened 2026-02-01
[Automation] Track collector metadata schema versions


jaydeluca:
The schemas/formats of the collector metadata might change over time, and we need to be able to support processing multiple versions at once if needed, and do backfill jobs if we need to rebuild the registry.

We need a way to annotate our registry data with some schema identifier. Since there is no `file_version` associated with the [collector schema](https://github.com/open-telemetry/opentelemetry-collector/blob/main/cmd/mdatagen/metadata-schema.yaml), we could instead create a system where we hash their schema and use that as the version identifier, or come up with another approach. Better yet we should go and propose/contribute a `file_version` or other schema versioning identifier to their existing schema to help us out.

Perhaps it would make sense to also copy and store the schema definitions (where possible) in a `meta` directory within the registry, to make mapping/lookups easy having a local source of truth.

We also need to ensure that an issue is automatically opened when a schema evolution is detected so we can ensure it is handled properly.

ThyTran1402:
Hi,

I can take a look and work on this issue if it's not resolved yet.

Thank you!

jaydeluca:
hey @ThyTran1402 that would be great. This work will actually relate to a comment I left this morning on another PR: https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/346#discussion_r3181085836

The schemas created by this issue can be used to populate the schema we post for the LLMs.

Let me know if you want to talk anything through or if you have any questions along the way.

ThyTran1402:
Hi @jaydeluca,

I'm able to find the root causes of this issue are:
1. Collector schema has no version identifier field: The Java instrumentation watcher works well because its upstream YAML (instrumentation-list.yaml) carries a file_format field (0.1, 0.2, 0.3, 0.5). The collector's `metadata.yaml` has no equivalent field.
There's no file_format, no schema hash, nothing that lets the parser know which version of the schema it's reading. Historical registry files have no such annotation either, so a backfill job has no way to know which parser to route them through. This is exactly why the issue proposes contributing a `file_format` field upstream.
2. No local schema snapshot for the collector: The configuration watcher has a `SchemaCopier`. that copies the entire upstream schema/ directory into ecosystem-registry/configuration/vX.Y.Z/ on every release. This gives a versioned local record of every schema revision which evolution is detectable by diffing consecutive versions.
[...]

PR #358, opened 2026-05-05
feat/fixed collector metadata schema versions

ThyTran1402:
**Motivation**
The collector watcher ingests `metadata.yaml` files from upstream with no version identifier on the schema. When the upstream schema evolves; the parser silently drops new fields or misreads restructured data. There is no local copy of the upstream schema stored, so drift is undetectable without manually diffing the live URL, and no automated alert exists when it happens.

**Issue Resolved**:
- #40 

**Approach**
1. Schema snapshot: copy `metadata-schema.yaml` from the cloned opentelemetry-collector repo into ecosystem-registry/collector/meta/v{version}/ once per release. Stored distribution-independently since the file lives in the core repo and applies to both core and contrib.
2. Schema fingerprinting: annotate every generated component YAML with a schema_hash field (12-char SHA-256 of the raw schema bytes). This let backfill jobs and future consumers identify which schema generation produced a given file without inspecting the schema itself.

**Changes**
collector_watcher/schema_copier.py
Change: `CollectorSchemaCopier.copy_schema()` and `compute_schema_hash()`

collector_watcher/collector_sync.py
Change: `save_version()` computes hash, passes it to inventory, copies schema to `meta/` for core only

collector_watcher/inventory_manager.py
Change: Added meta_schema_path(version); `save_versioned_inventory()` writes `schema_hash`; `load_versioned_inventory()` defaults to "unknown" for backward compat

collector_watcher/metadata_parser.py
Change: `BaseMetadataParser`, `MetadataParserV1`, `MetadataParserFactory`, `parse_component_metadata()`; `MetadataParser` wrapper unchanged from the caller's perspective

.github/workflows/nightly-registry-update.yml
Added Detect collector schema drift, Open or update schema drift issue, Close schema drift issue if schema is stable steps

tests/test_schema_copier.py: 11 tests
tests/test_metadata_parser.py: Extended: factory, V1, convenience function, auto-detection    
tests/test_inventory.py: Extended: `schema_hash` round-trip, backward compat, `meta_schema_path` 
tests/test_collector_sync.py: Extended: hash absent/present, schema copied to `meta/`, not duplicated in distribution dir 

**Testing**
- 135 tests pass 
- Verify `schema-drift` label exists in the repository 
- Trigger a manual `workflow_dispatch` run of the nightly workflow to confirm drift detection runs without error
- Confirm `ecosystem-registry/collector/meta/` is populated after the next nightly run

### DIFF
--- a/ecosystem-automation/collector-watcher/src/collector_watcher/collector_sync.py
+++ b/ecosystem-automation/collector-watcher/src/collector_watcher/collector_sync.py
@@ -15,6 +15,7 @@
 """Collector metadata synchronization to registry."""
 
 import logging
+from pathlib import Path
 from typing import Any
 
 from semantic_version import Version
@@ -23,6 +24,7 @@ from watcher_common.version_detector import VersionDetector
 from .component_scanner import ComponentScanner
 from .deprecation_detector import DeprecationDetector
 from .inventory_manager import InventoryManager
+from .schema_copier import SCHEMA_RELATIVE_PATH, CollectorSchemaCopier
 from .type_defs import DistributionName
 
 logger = logging.getLogger(__name__)
@@ -163,19 +165,43 @@ class CollectorSync:
         """
         Save scanned components for a specific version.
 
+        Records the schema content hash in every component YAML file for drift
+        detection and parser routing. When saving the core distribution, also
+        copies metadata-schema.yaml into the shared meta/ directory — once per
+        version, since the schema belongs to the opentelemetry-collector core repo
+        and is identical across distributions.
+
+        Registry layout after this call (core example):
+            ecosystem-registry/collector/
+                core/v{version}/*.yaml          (component data, each with schema_hash)
+                meta/v{version}/metadata-schema.yaml  (shared schema copy)
+
         Args:
             distribution: Distribution name
             version: Version being saved
             components: Scanned components
         """
+        repo_path = Path(self.repos[distribution])
+        copier = CollectorSchemaCopier()
+        schema_src = repo_path / SCHEMA_RELATIVE_PATH
+        schema_hash = copier.compute_schema_hash(schema_src)
+
         repository = self.get_repository_name(distribution)
         self.inventory_manager.save_versioned_inventory(
             distribution=distribution,
             version=version,
             components=components,
             repository=repository,
+            schema_hash=schema_hash,
         )
-        logger.info("  Saved %s %s", distribution, version)
+
+        # Copy the schema file once, from core, into the shared meta/ directory.
+        # Contrib does not contain mdatagen; copying from core avoids duplication.
+        if distribution == "core":
+            meta_dir = self.inventory_manager.meta_schema_path(version).parent
+            copier.copy_schema(repo_path, meta_dir)
+
+        logger.info("  Saved %s %s (schema_hash=%s)", distribution, version, schema_hash)
 
     def initialize_previous_version(self, distribution: DistributionName) -> None:
         """
--- a/ecosystem-automation/collector-watcher/src/collector_watcher/collector_sync.py
+++ b/ecosystem-automation/collector-watcher/src/collector_watcher/collector_sync.py
@@ -24,7 +24,7 @@ from watcher_common.version_detector import VersionDetector
 from .component_scanner import ComponentScanner
 from .deprecation_detector import DeprecationDetector
 from .inventory_manager import InventoryManager
-from .schema_copier import SCHEMA_RELATIVE_PATH, CollectorSchemaCopier
+from .schema_copier import CollectorSchemaCopier
 from .type_defs import DistributionName
 
 logger = logging.getLogger(__name__)
@@ -165,16 +165,19 @@ class CollectorSync:
         """
         Save scanned components for a specific version.
 
-        Records the schema content hash in every component YAML file for drift
-        detection and parser routing. When saving the core distribution, also
-        copies metadata-schema.yaml into the shared meta/ directory — once per
-        version, since the schema belongs to the opentelemetry-collector core repo
-        and is identical across distributions.
+        Stores the upstream schema in content-addressable storage at
+        ``meta/schemas/{hash}.yaml`` (deduplicated across versions and
+        distributions) and records that hash in every component YAML for
+        drift detection and parser routing.
 
-        Registry layout after this call (core example):
+        Schema is always read from the core repo: ``mdatagen`` lives only in
+        ``opentelemetry-collector``, and the schema is identical across
+        distributions, so contrib carries the same ``schema_hash`` as core.
+
+        Registry layout after this call:
             ecosystem-registry/collector/
-                core/v{version}/*.yaml          (component data, each with schema_hash)
-                meta/v{version}/metadata-schema.yaml  (shared schema copy)
+                {distribution}/v{version}/*.yaml   (component data, each with schema_hash)
+                meta/schemas/{hash}.yaml           (one file per distinct schema)
 
         Args:
             distribution: Distribution name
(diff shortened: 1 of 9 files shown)


ecosystem-automation/collector-watcher/src/collector_watcher/metadata_parser.py
```diff
@@ -161,79 +141,187 @@ def _parse_status(self, status: dict[str, Any]) -> dict[str, Any]:
...
+        parser_class = cls._parsers.get(schema_version)
+        if parser_class is None:
+            supported = ", ".join(sorted(cls._parsers.keys()))
+            raise ValueError(f"Unsupported schema_version: {schema_version!r}. Supported: {supported}")
+        return parser_class()
+
+    @classmethod
+    def get_default_parser(cls) -> BaseMetadataParser:
+        """Return the parser for the latest registered schema version."""
+        latest = sorted(cls._parsers.keys())[-1]
```
jaydeluca: sorted(["v1", "v10", "v2"]) → ["v1", "v10", "v2"] (lexicographic), returning "v2" instead of "v10" as the latest.

```suggestion
        latest = sorted(cls._parsers.keys(), key=lambda v: int(v[1:]))[-1]
```

ecosystem-automation/collector-watcher/src/collector_watcher/schema_copier.py
```diff
@@ -0,0 +1,76 @@
...
+            logger.debug("Schema file not found in repo at %s, skipping copy", src)
+            return None
+
+        target_dir.mkdir(parents=True, exist_ok=True)
+        dst = target_dir / SCHEMA_FILENAME
+        shutil.copy2(src, dst)
+        logger.debug("Copied schema to %s", dst)
+        return SCHEMA_FILENAME
+
+    def compute_schema_hash(self, schema_path: Path) -> str:
```
jaydeluca, line 75: small nitpick since it's a small method anyway, but we could re-use some code from watcher_common/content_hashing.py 

```python
from watcher_common.content_hashing import compute_content_hash, UNKNOWN_HASH

def compute_schema_hash(self, schema_path: Path) -> str:
    if not schema_path.exists():
        return UNKNOWN_HASH
    return compute_content_hash(schema_path.read_bytes())
```

ecosystem-automation/collector-watcher/src/collector_watcher/collector_sync.py
```diff
@@ -163,19 +165,43 @@ def save_version(
...
             distribution=distribution,
             version=version,
             components=components,
             repository=repository,
+            schema_hash=schema_hash,
         )
-        logger.info("  Saved %s %s", distribution, version)
+
+        # Copy the schema file once, from core, into the shared meta/ directory.
+        # Contrib does not contain mdatagen; copying from core avoids duplication.
```
jaydeluca: when I run this as-is, im seeing `schema_hash: unknown` in the output for contrib modules

ecosystem-automation/collector-watcher/src/collector_watcher/collector_sync.py
```diff
@@ -163,19 +165,43 @@ def save_version(
...
+        Records the schema content hash in every component YAML file for drift
+        detection and parser routing. When saving the core distribution, also
+        copies metadata-schema.yaml into the shared meta/ directory — once per
+        version, since the schema belongs to the opentelemetry-collector core repo
+        and is identical across distributions.
+
+        Registry layout after this call (core example):
+            ecosystem-registry/collector/
+                core/v{version}/*.yaml          (component data, each with schema_hash)
+                meta/v{version}/metadata-schema.yaml  (shared schema copy)
```
jaydeluca: The current layout severs the connection between the hash and the file it refers to. Given `schema_hash: abc123def456` in a component YAML, there is no easy way to retrieve the schema, you'd need to know which version directory to look in, then compute the hash of that file to confirm it matches.

Instead, what if we wrote each distinct schema file once, named by its content hash:

```
ecosystem-registry/collector/meta/schemas/abc123def456.yaml
ecosystem-registry/collector/meta/schemas/ff9900aabbcc.yaml
```

then given `schema_hash: abc123def456` in any component YAML, the corresponding schema is always at `meta/schemas/abc123def456.yaml`. 

Then in the logic for detecing drift, instead of finding and sorting versioned paths, the workflows could then hash the fresh schema and check whether that file already exists:

```bash
FRESH_HASH=$(sha256sum "$FRESH" | cut -c1-12)
STORED="ecosystem-registry/collector/meta/schemas/${FRESH_HASH}.yaml"

if [ ! -f "$STORED" ]; then
  echo "schema_changed=true" >> $GITHUB_OUTPUT
  # diff against the most recently written schema file
else
  echo "schema_changed=false" >> $GITHUB_OUTPUT
fi
```

jaydeluca (commented):
awesome job! I left a few comments but this looks great!

ecosystem-automation/collector-watcher/src/collector_watcher/metadata_parser.py
ThyTran1402: Yup you're right.

ecosystem-automation/collector-watcher/src/collector_watcher/schema_copier.py
ThyTran1402: Hmmm good catch. I think there's another option we can only import `compute_content_hash` and leave UNKNOWN_HASH as it is.

``` 
from watcher_common.content_hashing import compute_content_hash

def compute_schema_hash(self, schema_path: Path) -> str:
    if not schema_path.exists():
        return UNKNOWN_HASH
    return compute_content_hash(schema_path.read_bytes())
```

ThyTran1402: Or as you suggested promoting `UNKNOWN_HASH` into `watcher_common.content_hashing` as a shared sentinel, then import it. We can actually switch to `compute_content_hash`. Kept `UNKNOWN_HASH` local for now since it's only used here; happy to promote it to `watcher_common` if/when a second caller needs it.

ThyTran1402: Let me know what you think? so I can adjust the code accordingly.

ecosystem-automation/collector-watcher/src/collector_watcher/collector_sync.py
ThyTran1402: Yes. I'll fix it.

ecosystem-automation/collector-watcher/src/collector_watcher/collector_sync.py
ThyTran1402: Yup that really makes sense to me. I'll convert meta/ to content-addressable storage (`meta/schemas/{hash}.yaml)`, drop `meta_schema_path(version)` from `InventoryManager`, make `copy_schema` idempotent and hash-named, and add an orphan sweep in `cleanup_snapshots` / `delete_version`.

ThyTran1402: I incorporated your feedback above into this PR.
Thank you for the suggestions 🙏

collector-schema-versioning-plan.md
```diff
@@ -0,0 +1,308 @@
+# Collector Schema Versioning — Implementation Plan
+
```
jaydeluca: i think this might've been committed by mistake

ecosystem-automation/collector-watcher/src/collector_watcher/schema_copier.py
jaydeluca: i think is good as-is, thanks!

registry-structure.md
```diff
@@ -0,0 +1,216 @@
...
+java/
+└── javaagent/
+    └── {version}/
+        └── instrumentation.yaml
+```
+
+**One aggregated file** per version containing all instrumentations.
+
+### File Format
+
```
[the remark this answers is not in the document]
jaydeluca: i think file might have been committed by mistake?

jaydeluca (commented):
thanks @ThyTran1402 ! there's two valid copilot comments about github action permissions and just adding UTF-8 to a parser for consistency. Then i think there are 2 files that might've been committed by mistake.

Everything else looks good, i've tested it locally and I see everything i expect, nice work!

registry-structure.md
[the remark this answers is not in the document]
ThyTran1402: Yup. You're right. I'll remove 2 redundant files.

collector-schema-versioning-plan.md
ThyTran1402: Yup. I'll remove it.

jaydeluca (approved):
thank you @ThyTran1402 , great job!

PR #363, opened 2026-05-06
fix: use numeric sort for coveredEntries version order

suhr25:
## SUMMARY

Fix incorrect version ordering in `groupByModule()` where versions like `12.0` were sorted before `9.2`. 

---

## FIX

Updated sorting in `normalize-instrumentation.ts` to use numeric-aware comparison:

```ts id="r2tw1j"
a.name.localeCompare(b.name, undefined, { numeric: true })
```

Also added a regression test to prevent future ordering issues.

---

## CHECKLIST

* [x] Fixed version sorting
* [x] Added regression test
* [x] Verified tests pass
* [x] No breaking changes

### DIFF
--- a/ecosystem-explorer/src/lib/normalize-instrumentation.test.ts
+++ b/ecosystem-explorer/src/lib/normalize-instrumentation.test.ts
@@ -110,6 +110,28 @@ describe("groupByModule", () => {
       "cassandra-4.4",
     ]);
   });
+
+  it("sorts entries by semantic version order, placing double-digit versions after single-digit ones", () => {
+    const entries = [
+      makeEntry({ name: "jetty-httpclient-12.0" }),
+      makeEntry({ name: "jetty-httpclient-9.2" }),
+      makeEntry({ name: "jetty-8.0" }),
+      makeEntry({ name: "jetty-11.0" }),
+      makeEntry({ name: "jetty-12.0" }),
+    ];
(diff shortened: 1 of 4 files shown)


ecosystem-explorer/src/lib/normalize-instrumentation.test.ts
```diff
@@ -110,6 +110,28 @@ describe("groupByModule", () => {
...
+    const httpclient = modules.find((m) => m.name === "jetty_httpclient")!;
+    expect(httpclient.coveredEntries.map((e) => e.name)).toEqual([
+      "jetty-httpclient-9.2",
+      "jetty-httpclient-12.0",
+    ]);
+    const jetty = modules.find((m) => m.name === "jetty")!;
+    expect(jetty.coveredEntries.map((e) => e.name)).toEqual([
+      "jetty-8.0",
+      "jetty-11.0",
+      "jetty-12.0",
```
jaydeluca, line 132: when i look at this on the site in the deploy preview, it seems the ordering is desc, could we ensure it works in ascending order instead?

image

suhr25: Thanks for reviewing!
I updated that path to use ascending sorting.

jaydeluca (approved):
thanks!

Issue #370, opened 2026-05-06
[UI/UX] Phase 1: Foundation


vitorvasc:
Part of #84.

This issue tracks the implementation of the shared blocks across the Explorer: theme system, navigation, status pills, card primitives, etc.

The components will be shipped behind the `V1_REDESIGN` feature flag, and breaking changes shouldn't affect the productive website until the cleanup PR finally removes the flag and the deprecated components.

## PRs

- [x] PR 0 - Add `V1_REDESIGN` flag - #365 
- [x] PR 1 - Theme system: `src/styles/` folder, light/dark/auto switching, no-flash init, `<ThemeToggle />` authored - #377 
- [x] PR 2 - NavBar v1: opentelemetry.io-style nav, logo, theme toggle - #453 
- [x] PR 2b - v1 scaffolding: introduce `src/v1/`, mount `<V1App />` via single boundary read, move dormant navbar into `src/v1/components/layout/` - #470 
- [x] PR 3 - SubNav: breadcrumb + optional right-side actions slot - #482
- [x] PR 4 - `<StatusPill>` covering all 6 OTel stability levels + `GlowBadge` `secondary`/`error` variants - #455 
- [x] PR 5 - `<TypeStripe>` primitive + `DetailCard` stripe slot - #483 
- [x] PR 6 - `<FooterV1>` + `<CncfCallout>` - #484 
- [x] PR 7 - Playwright visual regression baseline + axe-core a11y - #487

MeloveGupta:
Hey @vitorvasc I had a look at your entire idea and would love to help contribute to this...
Of course only if you are ok with it :)

bhuvi-nadar:
Hi, I’m a UX designer, would love to contribute

vitorvasc:
Hey, thanks to both of you for the interest!

@MeloveGupta,  I'm trying to keep it small to avoid headache of a bunch of conflicts, so phase 1 is the foundation and shared-components work. To be honest, I'm trying to avoid async work on this phase because a PR may depend on a previous one, but I believe PRs 4, 5 and 7 are good candidates to start as drafts in the meantime. Let me know if you'd like to pick any of them :) 

[...]

Issue #84, opened 2026-02-12
[Explorer] UI/UX Design


jaydeluca:
The current design is a placeholder. We are looking for contributors to help us come up with the "v1" design that we can use for the initial launch of the website. 

This might turn into a bit of a "meta" issue, as it is a bit broad. If you are interested in contributing to any piece of this effort, no matter how small, feel free to engage in this issue discussion and we can break off a smaller sub-issue for you to take ownership of.

For each ecosystem, there will be a series of pages:

- Explorer home page 
  - Landing page for the explorer that provides navigation to various areas of the ecosystem
- Individual Ecosystem Landing page
  - This should include a short overview/summary of the particular area of the OTel ecosystem (Collector, Java agent etc)
- "List" pages 
  - for exploring various components/instrumentations/configurations
  - searchable/filterable
  - we need to determine what information is valuable to display, what is useful to filer on
- Component Detail Page
  - Detailed overview
  - Embedded documentation
  - Tools for diffing across versions etc

vitorvasc:
Hey!

Last week I used this issue as input to brainstorm some ideas with my fellow Claude Opus 4.7, and asked it to generate a mockup. I'll leave the images with the results below. Alternatively, I've also pushed the HTML and markdown files to a separate branch in my fork, in case someone wants to take a look or keep iterating on them!

- https://github.com/vitorvasc/opentelemetry-ecosystem-explorer/blob/feat/84-layout-mockups/ecosystem-explorer-v1-mockups.html
- https://github.com/vitorvasc/opentelemetry-ecosystem-explorer/blob/feat/84-layout-mockups/ecosystem-explorer-v1-design-brief.md
- https://github.com/vitorvasc/opentelemetry-ecosystem-explorer/blob/feat/84-layout-mockups/ecosystem-explorer/DESIGN.md

As a starting point, I kinda liked the results, tbh. I'd still go deeper on some components though, like the "Pipeline anatomy" on the "Individual Ecosystem Landing Page". On the other hand, the "List" and "Detail" pages are looking very very interesting. 🤔 

Let me know what you think!

---

<details>

<summary>Explorer home page</summary>

<img width="2304" height="2174" alt="Image" src="https://github.com/user-attachments/assets/ebf76022-1baa-4f74-ae83-1dea856c2cca" />

</details>

---

<details>

<summary>Individual Ecosystem Landing page</summary>

[...]

PR #377, opened 2026-05-06
feat(theme): modular CSS, light/dark/auto theme system (Phase 1 - Foundation PR 1)

vitorvasc:
Part of the Phase 1 foundation work for the explorer redesign (#84).

Contributes to #370.

## What's in this PR

- Splits `src/index.css` into focused partials under `src/styles/`:
  - `tokens.css` — all CSS custom properties (dark + new light theme block)
  - `base.css` — global resets and body defaults
  - `syntax.css` — code block / syntax highlight rules
  - `index.css` — entry point that `@import`s the partials
- The monolithic file is removed; `src/main.tsx` now imports from `src/styles/index.css`.
- Replaces the old `{ themeId, setThemeId }` API with a cleaner `{ mode, resolved, setMode }` shape:
  - `mode` — the user's stored preference: `"light" | "dark" | "auto"` 
  - `resolved` — the value actually applied to the document (`"light" | "dark"`)
  - `setMode` — writes to `localStorage["td-color-theme"]` and updates `[data-theme]` on `<html>`
- Adds an inline `<script>` in `index.html` that runs before first paint, reads `localStorage["td-color-theme"]`, and sets `[data-theme]` on `<html>`. Prevents the white flash users see on dark-preferred systems before React hydrates.
- Adds a `[data-theme="light"]` block in `tokens.css` alongside the existing dark defaults. Reconciles the five remaining dark surface tokens (`card`, `muted`, `border`, and variants) to the navy palette.
- Authors the `<ThemeToggle />` component (cycles `light → dark → auto`). Not yet rendered.
- Migrates remaining literal-hue usages to brand primitives following the `primary = blue / secondary = orange` realignment already on `main`

⚠️ Production-visible changes (no `V1_REDESIGN`gate):

- Five dark-mode surface tokens (`card`, `card-secondary`, `muted`, `muted-foreground`, `border`) shift to the navy palette on merge.

## What's not in this PR

- The toggle is not visible yet. It mounts in PR 2 alongside the new `NavBar`.
- Light theme colors are plumbed but the page-level design (hero, cards, etc.) stays unchanged until later PRs activate `V1_REDESIGN` in `App.tsx`.
- No changes to the feature-flag system or routing.

## Verification

- `bun run test` — 781 tests pass
- `tsc -b` — clean
- Theme init: open DevTools → Application → Local Storage. Toggle the control and confirm `td-color-theme` updates. Reload; confirm the selected theme is applied before React loads (no flash).
- Auto mode: change OS color scheme with the toggle set to `auto`; confirm the page updates without a reload.

---

> Co-authored with Claude Opus 4.7. Reviewed with Gemini 3 (see the comment).


vitorvasc:
Code review using Gemini 3:

<details>
<summary>Code review</summary>

I've completed a thorough review of the **feat/84-pr1-theme-system** branch.
Overall, the architecture is sound, and the two-tier brand system is a
significant improvement for maintainability. However, I've identified two
correctness bugs regarding the "auto" mode and theme persistence, as well as a
discrepancy in the feature-flagging strategy.

### 1. Correctness: Stale State in "Auto" Mode

I identified a bug in `src/theme-context.tsx` where the `systemTheme` state can
become stale.

**Scenario:**

1. User starts in `auto` mode (e.g., system is `dark`). `resolved` is `dark`.
2. User switches to `light` mode. The `useEffect` cleanup removes the
   `matchMedia` listener.
3. User changes OS preference to `light` while the app is still in `light` mode.
   The `systemTheme` state is NOT updated because the listener is gone.
4. User switches back to `auto` mode. The `resolved` value uses the stale
   `systemTheme` (still `dark`) instead of the current OS value.
5. The `useEffect` for the listener runs, but it only reacts to _future_
   changes, it doesn't sync the state with the current value immediately.

**Recommended Fix:** Sync the state immediately when entering `auto` mode in the
effect:

```typescript
useEffect(() => {
  if (mode !== 'auto') return;
  const mql = window.matchMedia('(prefers-color-scheme: dark)');

  // Sync immediately upon entering auto mode
  setSystemTheme(mql.matches ? 'dark' : 'light');

  const handler = (e: MediaQueryListEvent) => {
    setSystemTheme(e.matches ? 'dark' : 'light');
  };
  mql.addEventListener('change', handler);
  return () => mql.removeEventListener('change', handler);
}, [mode]);
```

### 2. Correctness: Persistence Alias for `dark-blue`

The author mentioned keeping `dark-blue` as a backwards-compatible alias, but
it's currently broken for persistence.

In `src/theme-context.tsx`:

```typescript
function isValidMode(value: string | null): value is ThemeMode {
  return ['light', 'dark', 'auto'].includes(value as ThemeMode);
}

// ... in ThemeProvider initialization
const stored = localStorage.getItem(STORAGE_KEY);
return isValidMode(stored) ? stored : 'auto';
```

Because `isValidMode` excludes `dark-blue`, any user with that value in their
`localStorage` will be reset to `auto` instead of `dark`.

**Recommended Fix:** Explicitly handle the alias during initialization:

```typescript
const stored = localStorage.getItem(STORAGE_KEY);
if (stored === 'dark-blue') return 'dark';
return isValidMode(stored) ? stored : 'auto';
```

Additionally, the **no-flash script** in `index.html` should also handle this
alias to prevent a flash of the system theme for `dark-blue` users:

```javascript
var resolved = (stored === "light" || stored === "dark" || stored === "dark-blue")
  ? (stored === "dark-blue" ? "dark" : stored)
  : matchMedia(...)
```

### 3. Strategy: The "Global" Theme Disruption

The rationale states: _"Phase 1 is gated behind a `V1_REDESIGN` feature flag so
every PR lands on `main` without disrupting production."_

**Observation:** This is currently inaccurate. Because `ThemeProvider` is
globally applied in `main.tsx` and the new CSS tokens in `tokens.css` replace
the old `index.css` values immediately, **the app's primary color and background
will change for all users (cyan → dark blue / black → dark navy) the moment this
PR merges**, regardless of the `V1_REDESIGN` flag.

If the intention is strictly zero disruption for production, the `ThemeProvider`
or the HSL variable definitions would need to be conditional on the flag.
However, if a global color update is acceptable, the rationale should be updated
to reflect that "V1_REDESIGN" gates UI components, but not the base brand
colors.

### 4. Semantic Risk: `GlowBadge` Variant Naming

I agree with the author's assessment that `variant="primary"` now rendering
orange (`secondary` tokens) is misleading. While acceptable as a temporary state
for Phase 1, I recommend adding a `TODO` comment in `glow-badge.tsx` explicitly
pointing to the PR 4 cleanup to ensure it isn't forgotten if the roadmap shifts.

### Summary of focused areas

- **No-flash script:** Correct logic, but lacks the `dark-blue` alias.
- **Effect timing:** The `resolved` update is safe, but the `systemTheme` sync
  issue noted above is a real bug.
- **CSS chaining:** Lightning CSS and Tailwind v4 handle the variable
  indirection correctly at build/runtime. No concerns there.
- **Keyboard cycle:** `light → dark → auto` is a logical progression. No
  concerns.

**Verdict:** Fix the `auto` mode sync bug and the `dark-blue` persistence alias,
and clarify the global nature of the color change before merging.

</details>

vitorvasc:
@jaydeluca screenshots are working now :) :) 

I've [pushed a fix](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/377/commits/ff1f132a740e5e0ba5bbfdcbe9bd67979ae130a6) to this branch so we capture images for both dark and light themes, but I'll update the comment template to show both of them in a separate PR.

ecosystem-explorer/src/features/java-agent/instrumentation-detail-page.tsx
```diff
@@ -317,7 +317,7 @@ export function InstrumentationDetailPage() {
                                 return (
                                   <li key={feature} className="flex items-start gap-2 text-sm">
                                     <Check
-                                      className="text-primary mt-0.5 h-4 w-4 flex-shrink-0"
+                                      className="text-secondary mt-0.5 h-4 w-4 flex-shrink-0"
```
jaydeluca, line 320: im not sure if  it's this particular line, but the tabs on this page are rendering as:

image


but they should be using the primary color and look like:

image

vitorvasc: good catch - thanks!

that was actually a different file, but the same logic. just pushed the fixes:  https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/377/commits/3cb03b11819b56c53044b51ea44c0f7d9bd3cfe4

jaydeluca (approved):

Issue #97, opened 2026-02-19
[Explorer] Semantic convention integration


jaydeluca:
In the POC, we compared all attributes to the latest semconv in order to identify which ones were "compliant" with the latest version. We want to incorporate this same type of information, but perhaps in a different way.

We should have a way to evaluate instrumentations, and the java agent as a whole, against various versions of semconv. With the idea being the user can toggle between different target semconv versions, and then view the compliance of the emitted telemetry from the java agent against that particular version.

Requirements:

- User is able to select/toggle between semconv versions in some way
- When viewing an instrumentation detail page, the attributes are analyzed to see if they are part of semconv or not, and are labeled accordingly

Example of how it looks in the [POC](https://jaydeluca.github.io/instrumentation-explorer/library/2.25.0/camel-2.20), as an example:

<img width="912" height="558" alt="Image" src="https://github.com/user-attachments/assets/184412ed-245d-482d-a66c-9043aee5bdc5" />

We will need to evaluate whether it makes sense to "compute" this information ahead of time and persist it alongside the metadata, or if we should maintain the semconv data set and do a comparison on the client side "on demand".

SurbhiAgarwal1:
Hi @jaydeluca,

I’d like to work on this.

From the POC description, my understanding is that we need to evaluate emitted telemetry (attributes) against different versions of semantic conventions (semconv), and expose this comparison in the Explorer UI with version-aware compliance labeling.

Proposed approach:

**1. Semconv version handling**

* Maintain a structured dataset of semconv definitions versioned (e.g., JSON per version)
* Provide a UI toggle to switch the active semconv version context

**2. Attribute compliance evaluation**
For each instrumentation:

* Compare emitted attributes against the selected semconv version
* Classify attributes as:

  * compliant (exists in semconv)
  * non-compliant (missing/unknown)
  * potentially deprecated (exists in older versions but not in selected one)

**3. Computation strategy (key design choice)**
I see two viable approaches:

**Option A – Precompute (server-side)**

* Compute compliance per semconv version and persist alongside metadata
* Pros: faster UI, simpler client logic
* Cons: storage overhead, recomputation on semconv updates

**Option B – On-demand (client-side or API layer)**

* Load semconv definitions and compute compliance dynamically
* Pros: flexible, always up-to-date with semconv versions
* Cons: higher runtime cost, potential performance impact

👉 My initial preference is a hybrid approach:

* Preload semconv datasets
[...]

PR #382, opened 2026-05-07
feat(db-builder): integrate Weaver for semconv compliance checking (#97)

SurbhiAgarwal1:
# Technical Detail: Semantic Convention Integration (Issue #97)

This document provides a technical deep-dive into the implementation of the Semantic Convention compliance pipeline in the `explorer-db-builder`.

## 1. Architectural Overview
The integration follows a "sidecar" enrichment pattern. Instead of modifying the core data structures, we introduce a `SemconvEnricher` that evaluates telemetry metadata against standard OTel registries using the **OpenTelemetry Weaver** engine.

### Data Flow
1. **Extraction**: Retrieve metrics and spans from the normalized `InstrumentationData`.
2. **Translation**: Map OTel signals to a Weaver-compatible "Application Registry".
3. **Evaluation**: Execute `weaver registry check` against a specific semconv version.
4. **Annotation**: Persist compliance status back to the telemetry metadata.

## 2. Component: `SemconvEnricher`
**Location**: `explorer_db_builder/semconv_enricher.py`

This is the primary orchestrator for compliance checking.

### Transformation Logic
The enricher generates a temporary directory containing:
- **`manifest.yaml`**: Defines the instrumentation name and the dependency on the official OTel semantic convention registry (e.g., `github.com/open-telemetry/semantic-conventions@v1.37.0`).
- **`telemetry.yaml`**: Translates internal metadata into Weaver's definition format.
  - **Metrics**: Defined with `type: metric` and attributes using the `ref` keyword to ensure Weaver validates them against the registry's definitions.
  - **Spans**: Defined with `type: span`, using synthetic IDs based on the instrumentation name and span kind (e.g., `activej-http.SERVER`).

### Weaver Invocation
The enricher calls the `weaver` CLI via a subprocess. 
- **Success Condition**: If `weaver registry check` exits with code 0, all signals defined in the registry are considered compliant.
- **Error Handling**: If errors are reported (return code 1), the enricher parses the `stderr` output to identify specific signals that failed validation and marks them accordingly.

## 3. Pipeline Integration
**Location**: `explorer_db_builder/main.py`

The enrichment stage is integrated into `process_version` immediately after the `transform_instrumentation_format` call.

```python
transformed_inventory = transform_instrumentation_format(inventory)

# Enrich with semantic convention compliance
try:
    enricher = SemconvEnricher()
    enricher.enrich_inventory(transformed_inventory)
except Exception as e:
    logger.warning(f"Semantic convention enrichment failed: {e}")
```

This placement ensures that:
- Enrichment works on normalized, clean data.
- The pipeline remains resilient (a Weaver failure does not crash the build).

## 4. Frontend & Metadata Schema
**Location**: `ecosystem-explorer/src/types/javaagent.ts`

The compliance status is persisted as a `semconv_compliance` array on individual telemetry signals:

```json
{
  "name": "http.server.request.duration",
  "unit": "s",
  "semconv_compliance": ["1.37.0"]
}
```

This structure is extensible, allowing an instrumentation to be marked as compliant with multiple semantic convention versions over time.

## 5. Verification & Testing
**Location**: `tests/test_semconv_enricher.py`

A dedicated test suite validates the following:
- **YAML Generation**: Ensures the generated `manifest.yaml` and `telemetry.yaml` are valid and follow Weaver's specification.
- **Version Extraction**: Tests the regex-based extraction of versions from OTel schema URLs.
- **Mocked CLI Interactions**: Simulates various Weaver output scenarios (total success, partial failure, and system errors) to verify that the metadata is updated correctly.


lucacavenaghi97:
Hi @SurbhiAgarwal1, I just noticed this branch includes the commit from #380, which is still under review. You can see the effect on Copilot's review: several of its comments here are actually about the README code from #380, not the semconv work. I'd rather wait for #380 to be merged and this branch rebased on main before doing a full review, so the diff reflects only the Weaver integration. Thanks for the work!

SurbhiAgarwal1:
Hi @lucacavenaghi97,

I have rebased this branch onto `main` and completely removed all commits and code related to PR #380. The diff now strictly reflects only the Weaver semantic convention integration.

I have also fixed the test failures and formatting issues, so all CI checks are now green and passing. It is ready for your review. Thanks!

vitorvasc:
Hi @SurbhiAgarwal1, could you run one more rebase, please? There are new conflicts with this branch.

Thanks for your patience!

jaydeluca (commented):
hey @SurbhiAgarwal1 apologies on the delay on getting to review this. I think this is the right idea, but the PR diff has gotten a little out of sync i think. I see that you added the semconv_compliance field to the javaagent.ts but i dont see that data in any of the data files.

Would you be able to clean this PR up a bit? Exclude the plan file, and don't commit any of the changes in ecosystem-explorer/public/data/*, i will verify those files look good locally and then after merge we will run a job to regenerate the db itself

jaydeluca:
the public/data files are still there

SurbhiAgarwal1:
Hi @jaydeluca,

Apologies for the noise on that! My local Windows git environment triggered automatic CRLF line-ending conversions on the generated database files, which accidentally staged and committed all of them.

I have just force-pushed a clean commit (`2a73bf5`) that completely resets the `ecosystem-explorer/public/data/*` and `ecosystem-registry/` directories to match `main` exactly, and excludes the plan file. 

The PR diff is now completely clean and shows exactly 8 source files changed with zero database file modifications. All tests are passing and it's fully ready for your review!

.markdownlint-cli2.yaml
```diff
@@ -9,6 +9,7 @@ ignores:
   - "**/node_modules/**"
   - ".claude/**"
   - "ecosystem-explorer/public/data/**"
+  - "ecosystem-explorer/dist/**"
```
jaydeluca, line 13: is this really needed?

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/semconv_enricher.py
```diff
@@ -0,0 +1,217 @@
...
+        Args:
+            instrumentation: Instrumentation data dictionary.
+        """
+        telemetry_entries = instrumentation.get("telemetry", [])
+        if not telemetry_entries:
+            return
+
+        # POC: For now, we only support a single semconv version per instrumentation based on its schema_url
+        schema_url = instrumentation.get("scope", {}).get("schema_url", "")
+        version = self._extract_version(schema_url) or "1.37.0"
```
jaydeluca: let's make this "1.37.0" a constant somewhere more prominent in the file

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/semconv_enricher.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def _prepare_weaver_registry(self, registry_dir: str, instrumentation: Dict[str, Any], version: str) -> None:
+        """Prepares a Weaver-compatible registry directory.
+
+        Args:
+            registry_dir: Temporary directory to create the registry in.
+            instrumentation: Instrumentation data.
+            version: Semantic Convention version to check against.
+        """
+        # manifest.yaml
```
jaydeluca: there's a few other comments that seem unecessary in this file too
```suggestion
```

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/semconv_enricher.py
```diff
@@ -0,0 +1,217 @@
...
+        """
+        self.weaver_path = weaver_path
+
+    def enrich_inventory(self, inventory_data: Dict[str, Any]) -> None:
+        """Enriches an entire inventory (libraries and custom instrumentations).
+
+        Args:
+            inventory_data: Transformed inventory data.
+        """
+        for key in ["libraries", "custom"]:
```
jaydeluca, line 48: this takes a really long time to run, and there's no progress indication along the way, I've actually had it running for over 10 minutes and just got through 1 version.

In the [POC project](https://github.com/jaydeluca/agent-report-card), I was able to handle things in a single shot per agent version, instead of running weaver for every instrumentation library individually. We should do that here too

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/semconv_enricher.py
```diff
@@ -0,0 +1,217 @@
...
+            registry_dir: Path to the Weaver registry.
+
+        Returns:
+            Dict mapping signal IDs to a boolean (True if compliant).
+        """
+        # In a real environment, this would call 'weaver registry check -r <registry_dir>'
+        # For this POC, we'll implement the subprocess call and handle the output.
+        # If weaver is not found, it will raise an exception which is caught in enrich_instrumentation.
+
+        cmd = [self.weaver_path, "registry", "check", "-r", registry_dir]
```
jaydeluca, line 213: we will need to ensure weaver is installed in our [action](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/blob/main/.github/workflows/build-explorer-database.yml) runner

jaydeluca (changes requested):

.markdownlint-cli2.yaml
SurbhiAgarwal1: I have reverted all changes to .markdownlint-cli2.yaml to keep the linter configuration clean and unchanged.

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/semconv_enricher.py
SurbhiAgarwal1: Done! I've moved "1.37.0" to a class-level constant DEFAULT_SEMCONV_VERSION at the top of SemconvEnricher and stripped out all the redundant and overly verbose comments (like # manifest.yaml, # telemetry.yaml, # Metrics, etc.) to keep the implementation concise.

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/semconv_enricher.py
SurbhiAgarwal1: Agreed. We will ensure the weaver binary is installed in our GitHub Actions runner environment to make sure the compliance check runs successfully during CI

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/semconv_enricher.py
SurbhiAgarwal1: Excellent point. I have refactored enrich_inventory to process instrumentations in a single shot per Semantic Convention version. It now groups all libraries and custom instrumentations by their semconv version and executes weaver registry check exactly once per version group using a consolidated telemetry registry.

Spawning a single subprocess per version instead of one per individual library has dramatically optimized performance, bringing the execution time down from over 10 minutes to just a few seconds! I've also retained enrich_instrumentation with its original behavior so that our unit tests continue to pass seamlessly

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/semconv_enricher.py
SurbhiAgarwal1: I've removed # manifest.yaml and cleaned up all other redundant comments in this file to keep it clean and focused.

jaydeluca:
@SurbhiAgarwal1 you resolved all the comment threads with the requested changes, but I don't see any actual changes made. Did you perhaps forget to push?

jaydeluca:
@SurbhiAgarwal1 i ran this locally and the performance is good now, thanks for taking care of that. I'm not actually seeing the results of it working though when I run it. There are no attributes being annotated with the new semconv field.

jaydeluca:
@SurbhiAgarwal1 just wanted to check in to see if you are still interested working on this? It's a bit complicated so no worries if not.

there's been some developments in this space where we now also have some new tooling in https://github.com/open-telemetry/semantic-conventions-conformance that we might be able to leverage.

SurbhiAgarwal1:
Hi @jaydeluca, thanks for checking in! Yes, I'm definitely still interested in wrapping this up. I'll rebase on main, fix the ruff linting failure in CI, and investigate why attributes aren't getting annotated properly with the semconv_compliance field. I'll also take a look at semantic-conventions-conformance to see if we can leverage it here. Will push an update shortly!

PR #387, opened 2026-05-07
Move netlify edge function location

jaydeluca:
I noticed that the edge functions are not showing up in the netlify console. The netlify.toml sets base = "ecosystem-explorer", which means Netlify looks for edge functions at ecosystem-explorer/netlify/edge-functions/.

Trying to move it to see if that solves it

### DIFF
--- a/ecosystem-explorer/netlify/edge-functions/agent-negotiation.ts
+++ b/ecosystem-explorer/netlify/edge-functions/agent-negotiation.ts
@@ -79,7 +79,7 @@ export default async (request: Request, context: Context) => {
     "/agent/javaagent/versions/": "/agent/javaagent/versions.md",
   };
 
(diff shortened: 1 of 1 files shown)


ecosystem-explorer/netlify/edge-functions/agent-negotiation.ts
```diff
@@ -79,7 +79,7 @@ export default async (request: Request, context: Context) => {
     "/agent/javaagent/versions/": "/agent/javaagent/versions.md",
   };
 
-  let resolvedPath = agentPathMap[pathname] || pathname;
+  const resolvedPath = agentPathMap[pathname] || pathname;
```
jaydeluca, line 82: the file path is the real change, this let -> const was because the linter complained

maryliag (approved):

Issue #396, opened 2026-05-08
Dropdown items not visible due to low contrast (white background issue) in Telemetry section


hussainjamal760:
The dropdown menu currently has a visibility and contrast issue where the background color is white and the text color does not have sufficient contrast. Because of this, most dropdown items are not clearly readable in their default state, and only become visible when hovered.

From a user experience perspective, this can create confusion and friction, especially for first-time users or users who are quickly scanning options. Instead of instantly perceiving available choices, users are forced to interact (hover) to understand what options exist, which is not ideal for usability or accessibility.

This also impacts visual clarity and increases cognitive load, as users need to spend extra effort to identify and distinguish options within the dropdown.

Overall, the current design reduces readability and discoverability of menu items, which may negatively affect user interaction flow.

<img width="649" height="171" alt="Image" src="https://github.com/user-attachments/assets/bea228ce-7d19-4418-881d-639f1b489f4c" />

yugeshweb:
Hi, is this issue fixed ? if not let me know if i can help !

jaydeluca:
this was fixed by #397

PR #397, opened 2026-05-08
Fix dropdown visibility contrast issue

hussainjamal760:
Issue: #396 

Overview

This PR fixes a dropdown UI visibility issue where menu items were not clearly readable due to low contrast between the text and background.

The goal of this change is to improve readability, usability, and overall user experience by ensuring all dropdown options are clearly visible in their default state.

What was the issue?

The dropdown menu had a white background with text colors that blended in, making most items difficult to read unless hovered. This created a poor visual hierarchy and required unnecessary interaction just to view available options.

What has been improved?
Improved text and background contrast for better readability
Ensured all dropdown items are visible in default state (no hover required)
Enhanced visual clarity for faster scanning of options
Improved overall accessibility and user experience
User Impact

This change reduces confusion for users and improves interaction flow by making dropdown options immediately visible and easy to distinguish.

Testing
Verified dropdown visibility in default and hover states
Checked readability across multiple screen sizes
Ensured no layout or styling regressions introduced


jaydeluca (changes requested):
i am unable to reproduce the issue you are reporting. What browser are you using? I'm going to need to rely on you providing screenshots after each change to verify things are fixed since i cant reproduce

Also, it looks like this change makes the dropdown overflow the container:

image

hussainjamal760:
> i am unable to reproduce the issue you are reporting. What browser are you using? I'm going to need to rely on you providing screenshots after each change to verify things are fixed since i cant reproduce
> 
> Also, it looks like this change makes the dropdown overflow the container:

I’m using Chrome on Windows. I’ll provide screenshots with the next update so the behavior can be verified more easily.

Thanks for pointing out the dropdown overflow issue as well , I’ll investigate and fix that in the next revision.

hussainjamal760:
Here is the video link demonstrating both the issue before the fix and the behavior after the fix:

https://www.loom.com/share/0721d952f3284de6bf65e97b9672b604

jaydeluca (approved):

Issue #330, opened 2026-04-30
Collector page followups


jaydeluca:
These are followups for the new collector sections.

**IMPORTANT** This is a tracking issue but all of these should not be solved in a single PR. Each piece should be handled in a separate PR

## List Page 

- The stability sections currently just have a single badge. We need to provide a more detailed view into the stability information, for example what each signals stability level is
- Remove the "Browse technical details...." wording on the component. If a component has a description, use that instead, otherwise leave it blank

### "Component Info" section ✅ 

Completed via #355

- For each `type`, underneath the value we should add a short description, like on https://opentelemetry.io/docs/collector/components/receiver/ says "Receivers collect telemetry data from various sources and formats.". We should style it like we do with the "Features" on the javagagent page

<img width="1341" height="177" alt="Image" src="https://github.com/user-attachments/assets/d8a99339-740f-4c7a-a48f-6719c799a13c" />

So we will need to create a map of type -> description using the blurbs from opentelemetry.io, and then render them underneath.

- Fix the "component documentation" link, which currently links to https://github.com/opentelemetry-collector-contrib/tree/main/..... instead of https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main....  I think the link generation logic is not correct.

## Stability Tab

Fix the stability table to be more intuitive as to what each stability level actually means. Again it might be useful to take inspiration from https://opentelemetry.io/docs/collector/components/receiver/ and do a table of some sort

<img width="350" height="106" alt="Image" src="https://github.com/user-attachments/assets/b07e7f2c-2e37-45dd-a728-0ba617ce60a7" />

The distribution availability section should be completely revanmped, and we should provide information into what each distribution is and how to get it/use it.

Remove the "Ownership" tab entirely, it is not useful in its current state.

## Telemetry tab

Tracking as part of subtask #336

Add a new tab for telemetry if it exists, and render the metrics, example: https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/blob/main/ecosystem-registry/collector/contrib/v0.151.0/receiver.yaml#L77

STyle it similar to how we are displaying telemetry in the javaagent.

jaydeluca:
@DCchoudhury15 are you interested in any of the items on this issue?

kajal-jotwani:
can I also work on this issue?

DCchoudhury15:
[...]

PR #399, opened 2026-05-09
feat(collector): revamp stability tab and distribution availability

DCchoudhury15:
Closes part of #330

Revamped the Stability tab on the collector detail page:

- Replaced the old card-based stability view with a signal/level table
- Added a Stability Legend showing what each level means, filtered to only levels the component uses
- Replaced distribution badges with cards that include a description, install command, and documentation link
- Added specific info for contrib, core, and k8s distributions
- Removed the Ownership tab as it wasn't useful
- here is how it looks on my local machine -
image


DCchoudhury15:
Hi @jaydeluca, the PR replaces the old card-based stability view with a signal/level table driven purely from the YAML, revamps the distribution section with cards showing install commands for contrib, core and k8s, and removes the Ownership tab. One thing worth double checking is the `getSignalStability` function which returns on the first matching level it finds wanted to make sure that aligns with how the YAML data is structured. Thank you .

DCchoudhury15:
Hi @jaydeluca,  just pushed up the fixes for the copilot comments. I swapped the legend over to use the GlowBadge component so it perfectly matches the table, added the scope="col" attribute to the headers for screen readers, and cleaned up the column names so they read a bit better. I also added in a quick fallback state just in case a component has an empty distributions array. Ready for another look whenever you have time!

jaydeluca:
thanks @DCchoudhury15. I've reached out and asked for someone who is closer to the collector project to weigh in on how we might best visualize the stability elements. It might be a few days before he is able to get to it, but wanted to let you know we're going to just keep this on hold until we can get that feedback.

thanks for your patience!

vitorvasc:
FYI: I'm moving this PR to draft state so we can better track what's ready for review.

ecosystem-explorer/src/features/collector/collector-detail-page.tsx
```diff
@@ -37,6 +37,94 @@ const COMPONENT_TYPE_DESCRIPTIONS: Record<string, string> = {
...
 
+const STABILITY_DEFINITIONS: Record<string, { label: string; desc: string; color: string }> = {
+  development: {
+    label: "Development",
+    desc: "Not ready for general use. May have bugs or lack documentation.",
+    color: "text-blue-500",
+  },
+  alpha: {
+    label: "Alpha",
+    desc: "Ready for testing and early feedback. Not recommended for production.",
```
jaydeluca: could we try and get as close to the wording used in this doc as possible: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/component-stability.md for all of these?

```suggestion
    desc: "The component is ready to be used for limited non-critical workloads and the authors of this component would welcome your feedback.",
```

jaydeluca (commented):
i think we can merge this as is, and use #565 to discuss whether there's a different way we should display this, but we can use this as a starting point and iterate. it's all still gated so not too big of a deal. Just one last request about some of the wording for the different stability levels

DCchoudhury15:
Hi @jaydeluca , done the changes , ready for review when you are ready .

jaydeluca (approved):
thanks @DCchoudhury15 and thank you for your patience on the review 🙏

PR #401, opened 2026-05-09
fix: resolve permanent loading state in useConfigSchema and useConfigStarter when version is empty

### LINKED ISSUE #393: useConfigSchema and useConfigStarter get stuck in a permanent loading state when version is empty
I was reading through the hook files while working on the configuration builder feature and noticed an inconsistency that causes a real user-facing bug.

In `[ecosystem-explorer/src/hooks/use-configuration-data.ts]`(vscode-webview://0ht8dg5s7tnanudnj344htj01ahhe2of51tlhfk019dfp9b0h1fr/ecosystem-explorer/src/hooks/use-configuration-data.ts), both `useConfigSchema` (line 69) and `useConfigStarter` (line 109) guard against an empty version prop like this:

`if (!version) return;`
The problem is that when version is empty or falsy, the function exits without updating the loading state. The hook was initialized with `loading: true`, and it stays that way forever. Any component consuming either of these hooks will display a loading spinner indefinitely instead of showing a ready or idle state.

Compare this to `useCollectorComponents` in `[ecosystem-explorer/src/hooks/use-collector-data.ts]`(vscode-webview://0ht8dg5s7tnanudnj344htj01ahhe2of51tlhfk019dfp9b0h1fr/ecosystem-explorer/src/hooks/use-collector-data.ts), which handles the empty version case correctly:

`if (!version) {
  setState({ data: null, loading: false, error: null });
  return;
}`
The same correct pattern is also used in `useCollectorComponent` in the same file. So the fix is well established in the codebase already, it just was not applied consistently to the configuration data hooks.

To reproduce:

1. Mount any component that calls `useConfigSchema("")` or `useConfigStarter("")` with an empty string as the version (this happens naturally while the versions index is still loading and no version has been selected yet).

2. Observe that the returned state has `loading: true` and never updates to `loading: false`.

3. The component shows a loading spinner that never resolves.

### Expected behavior:

When version is empty, the hook should immediately return `{ data: null, loading: false, error: null },` the same way the collector hooks do. This lets the UI know there is nothing to load yet and it can render a ready/idle state rather than a spinner.

### Actual behavior:

The state stays stuck at { data: null, loading: true, error: null } permanently when version is empty.

### Affected files:

`ecosystem-explorer/src/hooks/use-configuration-data.ts`, lines 69 and 109

### Fix:

Replace the bare returns in both hooks with the same pattern used in 
`use-collector-data.ts:

// useConfigSchema
if (!version) {
  setState({ data: null, loading: false, error: null });
  return;
}

// useConfigStarter
if (!version) {
  setState({ data: null, loading: false, error: null });
  return;
}`

This is a small, focused, self-contained fix. No API surface changes, no type changes. The existing tests in `use-configuration-data.test.ts` can be extended with a simple case for each hook to cover the empty version path.

Rama542:
Fixes #393

Both useConfigSchema and useConfigStarter in use-configuration-data.ts had a bare return statement when the version argument was empty. Because the initial state has loading: true, this caused any component consuming these hooks to stay in a permanent loading state whenever version was not yet available (for example while the version index is still being fetched).

The fix follows the same pattern already used in use-collector-data.ts for useCollectorComponents and useCollectorComponent, which both call setState({ data: null, loading: false, error: null }) before returning early on an empty version.

Changes made:

In useConfigSchema, replaced the bare return with a proper setState call that sets loading to false before returning
In useConfigStarter, same change applied
Updated the two existing tests that were asserting the old broken behavior (loading: true on empty version) to now assert the correct behavior (loading: false on empty version)
All existing tests pass. No API changes, no type changes.

### DIFF
--- a/ecosystem-explorer/src/hooks/use-configuration-data.test.ts
+++ b/ecosystem-explorer/src/hooks/use-configuration-data.test.ts
@@ -129,11 +129,12 @@ describe("useConfigSchema", () => {
     expect(configData.loadConfigSchema).toHaveBeenCalledWith("1.0.0");
   });
 
-  it("should stay loading when version is empty", () => {
+  it("should set loading to false when version is empty", () => {
     const { result } = renderHook(() => useConfigSchema(""));
 
-    expect(result.current.loading).toBe(true);
+    expect(result.current.loading).toBe(false);
     expect(result.current.data).toBeNull();
+    expect(result.current.error).toBeNull();
     expect(configData.loadConfigSchema).not.toHaveBeenCalled();
   });
 
@@ -203,10 +204,11 @@ describe("useConfigStarter", () => {
     expect(result.current.error).toBeNull();
   });
 
-  it("stays loading when version is empty", () => {
+  it("sets loading to false when version is empty", () => {
     const { result } = renderHook(() => useConfigStarter(""));
-    expect(result.current.loading).toBe(true);
+    expect(result.current.loading).toBe(false);
     expect(result.current.data).toBeNull();
+    expect(result.current.error).toBeNull();
     expect(configData.loadConfigStarter).not.toHaveBeenCalled();
   });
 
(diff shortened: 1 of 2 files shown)


jaydeluca (approved):
nice find, thanks

PR #411, opened 2026-05-09
test: Collector error and invalid-version paths lack test coverage

### LINKED ISSUE #410: [test] Collector error and invalid-version paths lack test coverage
## Summary

The collector page test suite has a single smoke test that only checks the page title renders. Error states, loading states, and invalid version routes have no coverage. Without these tests, any future refactor to `CollectorPageInner` could silently break the error UI with no safety net.

---

## The Problem

`collector-page.test.tsx` currently contains one test:

```tsx
it("renders the page title", () => { ... });
```

The following behaviors are completely untested:

- `useCollectorVersions` returning an error -> page should show "Error loading data"
- `useCollectorComponents` returning an error -> page should show "Error loading data"
- An invalid version in the URL (e.g. `/collector/components/9.9.9`) causing a 404 on the manifest fetch -> the page correctly shows "Error loading data", but there is no test asserting this, so a future refactor could silently regress it to a misleading empty state
- Loading state while versions or components are being fetched

The correct error UI already exists in `CollectorPageInner` - `componentsError` and `versionsError` are both wired up. The problem is purely the absence of tests, meaning this behavior is one unguarded refactor away from breaking silently.

---

## Steps to Reproduce

**For the missing test coverage:**

```bash
cd ecosystem-explorer
npx vitest run src/features/collector/collector-page.test.tsx
```

Only 1 test runs. None of the error, loading, or invalid-version paths are exercised.

**To observe the current correct behavior that needs to be locked in with tests:**

1. Start the dev server: `bun run serve` from `ecosystem-explorer/`
2. Navigate to `/collector/components/9.9.9`
3. Observe: the page correctly shows "Error loading data / Please try refreshing the page."
4. There is no test asserting this, it is only verified manually

---

## Proposed Fix

Expand `collector-page.test.tsx` to cover:

1. **Versions error** - mock `useCollectorVersions` returning an error, assert "Error loading data" heading is shown
2. **Components error** - mock `useCollectorComponents` returning an error, assert "Error loading data" heading is shown
3. **Invalid version route regression** - render at `/collector/components/9.9.9` with `useCollectorComponents` returning a 404 error, assert "Error loading data" is shown and "Showing" text is absent
4. **Loading state** - mock both hooks as loading, assert "Loading components..." is shown
5. **Successful render** - mock both hooks with valid data, assert component cards and "Showing N components" are rendered

MeloveGupta:
## What

Expands `collector-page.test.tsx` from a single smoke test to a full suite covering error states, loading states, and invalid version routes.

---

## Changes

- Added 6 new tests to `src/features/collector/collector-page.test.tsx`:
  - **Loading state** - asserts "Loading components..." renders while hooks are loading
  - **Versions error** - asserts "Error loading data" renders when `useCollectorVersions` fails
  - **Components error** - asserts "Error loading data" renders when `useCollectorComponents` fails
  - **Invalid version regression** - renders at `/collector/components/9.9.9`, asserts error UI is shown, "Showing 0 components" is absent, and `useCollectorComponents` was called with "9.9.9"`
  - **Successful render** - asserts component cards and "Showing 2 components" render correctly
  - **Version selector** - asserts available versions are rendered in the dropdown

- Tightened the invalid-version regression assertion to use a `textContent` matcher to correctly handle the nested components

---

## Verified By

- `npx vitest run src/features/collector/collector-page.test.tsx` - 7 tests pass
- Full suite run before and after +6 passing tests
- Confirmed `/collector/components/9.9.9` shows "Error loading data" in the browser

---

## Type of Change

- [x] Test

Closes #410

### DIFF
--- a/ecosystem-explorer/src/features/collector/collector-page.test.tsx
+++ b/ecosystem-explorer/src/features/collector/collector-page.test.tsx
@@ -28,6 +28,7 @@ vi.mock("@/lib/feature-flags", () => ({
 }));
 
 import { useCollectorVersions, useCollectorComponents } from "@/hooks/use-collector-data";
+import type { CollectorComponent } from "@/types/collector";
 
 const mockVersionsData = {
   versions: [
@@ -36,7 +37,7 @@ const mockVersionsData = {
   ],
 };
 
-const mockComponents = [
+const mockComponents: CollectorComponent[] = [
   {
     id: "receiver-otlp",
     name: "otlpreceiver",
--- a/ecosystem-explorer/src/features/collector/collector-page.test.tsx
+++ b/ecosystem-explorer/src/features/collector/collector-page.test.tsx
@@ -46,7 +46,11 @@ const mockComponents: CollectorComponent[] = [
     ecosystem: "collector",
     type: "receiver",
     distribution: "core",
-    status: { stability: { stable: ["traces", "metrics", "logs"] } },
+    status: {
+      class: "receiver",
+      stability: { stable: ["traces", "metrics", "logs"] },
+      distributions: ["core"],
+    },
   },
   {
     id: "processor-batch",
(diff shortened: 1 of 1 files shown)


MeloveGupta:
Hey @jaydeluca ,
I'd love to have your review on this...

jaydeluca (approved):
Thanks for adding the tests. Just as feedback for the future, I think the regression test alone would have been enough in this case. We try to be careful about growing the test suite too much over time since it impacts execution time, so we usually prioritize tests that cover complex functionality or protect against known regressions.

MeloveGupta:
Sure, I'll keep that in mind :)

PR #413, opened 2026-05-09
fix: parse key=value pairs for map type in parseDefault

### LINKED ISSUE #408: parseDefault ignores the raw value for map type configurations and always returns an empty object
I was reading through the configuration code while exploring the Java agent feature and noticed that the `parseDefault `function in `ecosystem-explorer/src/lib/declarative-name.ts handles the "map"` type incorrectly.

Here is the relevant code at line 48:

`case "list":
  if (typeof raw !== "string" || raw === "") return [];
  return raw.split(",").map((item) => item.trim());
case "map":
  return {};`

For the `"list"` type, the function correctly parses a comma-separated string like `"GET,POST,DELETE"` into an array` ["GET", "POST", "DELETE"]`. But for the `"map" `type, it always returns an empty object `{}` regardless of what the raw value contains. The raw value is completely ignored.

This means if a map-type configuration has a non-empty default like `"service.name=my-service"` or `"timeout=5000,retries=3"`, calling `parseDefault("map", "service.name=my-service")` returns `{}` instead of `{ "service.name": "my-service" }`. The default values silently get dropped every time a user opens a map configuration field.

The existing test in `declarative-name.test.ts` only covers the empty string case:

`it("returns an empty object for map default", () => {
  expect(parseDefault("map", "")).toEqual({});
});`
There is no test for a non-empty map default, so the missing parsing logic was never caught.

The fix should parse the raw string for the map case using the same comma-separated approach as the list case, but splitting each entry on `=` to produce key-value pairs. Something like splitting `"key1=value1,key2=value2"` into `{ key1: "value1", key2: "value2" }`. A test for a non-empty map default should be added at the same time.

Rama542:
Fixes #408

The parseDefault function in declarative-name.ts was not parsing map type defaults correctly. When the type was "map", the function always returned an empty object regardless of what the raw value was. This means that any instrumentation configuration with a map type default like "host1=serviceA,host2=serviceB" would be silently ignored and the user would see no pre-filled values.

The fix adds proper CSV parsing for map types. Each comma-separated token is split on the first equals sign to produce a key and value pair. Tokens without an equals sign are skipped. Whitespace around keys and values is trimmed. Empty string input still returns an empty object as before.

New test cases cover:
- Parsing a standard key=value CSV string
- Trimming whitespace around keys and values
- Skipping invalid tokens that have no equals sign
- Empty string still returns empty object (existing behavior preserved)

The implementation mirrors the existing list type handling already present in the same switch statement.

https://github.com/user-attachments/assets/aab73319-d84b-4dac-92d5-fee098ce8b5b

### DIFF
--- a/ecosystem-explorer/src/lib/declarative-name.test.ts
+++ b/ecosystem-explorer/src/lib/declarative-name.test.ts
@@ -102,8 +102,29 @@ describe("parseDefault", () => {
     expect(parseDefault("list", "")).toEqual([]);
   });
 
-  it("returns an empty object for map default", () => {
+  it("returns an empty object for empty map default", () => {
     expect(parseDefault("map", "")).toEqual({});
+  });
+
+  it("parses key=value pairs for map default", () => {
+    expect(parseDefault("map", "host1=serviceA,host2=serviceB")).toEqual({
+      host1: "serviceA",
+      host2: "serviceB",
+    });
+  });
+
+  it("trims whitespace around keys and values in map default", () => {
+    expect(parseDefault("map", " host1 = serviceA , host2 = serviceB ")).toEqual({
+      host1: "serviceA",
+      host2: "serviceB",
+    });
+  });
+
+  it("skips pairs with no equals sign in map default", () => {
+    expect(parseDefault("map", "host1=serviceA,invalidentry,host2=serviceB")).toEqual({
+      host1: "serviceA",
+      host2: "serviceB",
+    });
   });
 
   it("trims whitespace inside list CSV", () => {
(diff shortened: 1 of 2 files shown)


lucacavenaghi97 (approved):
thanks for the fix! 🎉 Validated locally with a patched registry entry: stored {} before, parsed pairs after. LGTM!

Rama542:
Hi, all checks have passed and the changes are approved. Could you please merge this PR when you get a chance? 
Thank you!

jaydeluca (approved):

PR #417, opened 2026-05-09
Fix screenshot workflow: derive versions at runtime, fix collector URL, assert no error pages

### LINKED ISSUE #415: Screenshot workflow silently captures error pages due to stale hardcoded version and invalid collector URL
## Summary

`take-screenshots.mjs` hardcodes a Java instrumentation version (`2.25.0`) that no longer exists in the generated data, and uses an invalid collector detail URL. The workflow silently captures and uploads screenshots of error pages without failing.

---

## The Problem

Two hardcoded values in `ecosystem-explorer/scripts/take-screenshots.mjs` are stale:

1. `DETAIL_VERSION = "2.25.0"` - version `2.25.0` does not exist in `public/data/javaagent/versions/`. Available versions are `2.26.1` and `2.27.0`. Navigating to `/java-agent/instrumentation/2.25.0/spring-webmvc-6.0` renders "Error loading instrumentation - Unexpected token '<'..."

2. `/collector/components/latest/receiver-otlp` has two problems:
   - `latest` is not a valid version, only `0.150.0` exists in `public/data/collector/versions/`
   - `receiver-otlp` is not a valid component ID, the correct ID is `core-receiver-otlpreceiver`

The script has no content assertion after navigation, it only calls `settle()` (networkidle wait) and then `page.screenshot()`. So error pages are captured and uploaded as if they were valid screenshots, with no CI failure to signal the problem.

Hardcoding any specific version is also fragile and it will go stale again the next time a new Java agent or collector version is released. The fix should derive versions at runtime from the generated data files.

---

## Steps to Reproduce

Verify the stale data directly:

```bash
ls ecosystem-explorer/public/data/javaagent/versions/
# 2.26.1-index.json  2.27.0-index.json  (2.25.0 is absent)

ls ecosystem-explorer/public/data/collector/versions/
# 0.150.0-index.json  (latest is not a valid version token)

ls ecosystem-explorer/public/data/collector/components/ | grep otlp
# core-receiver-otlpreceiver  (receiver-otlp does not exist)
```

Then build and run the screenshot script:

```bash
cd ecosystem-explorer
bun run build
node scripts/take-screenshots.mjs
```

Observations:
- Script exits with "All screenshots completed successfully!" and no error
- Open `screenshots/desktop-detail-details.png` -> shows "Error loading instrumentation - Unexpected token '<'..."
- Open `screenshots/desktop-collector-detail.png` -> shows "404 - Page Not Found"

Both are error pages silently captured and uploaded by CI as if they were valid.

---

## Proposed Fix

In `take-screenshots.mjs`:

1. **Derive versions at runtime** - read `public/data/javaagent/versions-index.json` and `public/data/collector/versions-index.json` at script startup to resolve the latest version dynamically, instead of hardcoding `DETAIL_VERSION = "2.25.0"`. This prevents the script from going stale again on the next release.

2. **Fix the collector detail URL** - derive the latest collector version from the versions index and pick a known stable component ID (e.g. `core-receiver-otlpreceiver`) instead of using `latest/receiver-otlp`.

3. **Add content assertions** - after each navigation, assert that no error heading is present in the page before calling `page.screenshot()`. If an error is detected, throw so the script exits non-zero and CI fails visibly instead of uploading error pages silently.

---

## Steps to Verify

After the fix:

```bash
cd ecosystem-explorer
bun run build
node scripts/take-screenshots.mjs
```

- Script should exit with "All screenshots completed successfully!"
- `screenshots/desktop-detail-details.png` should show a valid instrumentation detail page
- `screenshots/desktop-collector-detail.png` should show a valid collector component page
- Reverting to the old hardcoded version and URL should cause the script to throw and exit non-zero

MeloveGupta:
## What

Fixes the screenshot workflow which was silently capturing and uploading error pages due to a stale hardcoded Java version and an invalid collector detail URL.

---

## Changes

**`ecosystem-explorer/scripts/take-screenshots.mjs`**
- Removed hardcoded `DETAIL_VERSION = "2.25.0"` - replaced with `resolveLatestVersion()` which reads `public/data/javaagent/versions-index.json` at runtime and picks the entry where `is_latest: true`. This prevents the script from going stale on every new release.
- Added `COLLECTOR_VERSION` resolved the same way from `public/data/collector/versions-index.json`
- Fixed collector detail URL from `/collector/components/latest/receiver-otlp` to `/collector/components/${COLLECTOR_VERSION}/core-receiver-otlpreceiver`
- Added `assertNoError()` helper that checks for error or 404 headings after each navigation and throws if found, so the script exits non-zero instead of silently uploading error pages

**`.github/workflows/screenshots-capture.yml`**
- Added `VITE_FEATURE_FLAG_COLLECTOR_PAGE: "true"` to the build step so the collector detail route is enabled in the production build used by the screenshot workflow

---

## Before / After

| Page | Before | After |
|------|--------|-------|
| Java instrumentation detail | ❌ "Error loading instrumentation - Unexpected token '<'" | ✅ Spring WebMVC detail, version 2.27.0 |
| Collector detail | ❌ "404 - Page Not Found" | ✅ OTLP Receiver detail, version 0.150.0 |

Before:
Screenshot from 2026-05-10 02-17-11
Screenshot from 2026-05-10 02-17-35

After:
Screenshot from 2026-05-10 02-37-52
Screenshot from 2026-05-10 02-37-28

---

## Verified By

- Built with `VITE_FEATURE_FLAG_COLLECTOR_PAGE=true bun run build`
- Ran `node scripts/take-screenshots.mjs` - with "All screenshots completed successfully!"
- Confirmed `screenshots/desktop-detail-details.png` shows Spring WebMVC 2.27.0 detail page
- Confirmed `screenshots/desktop-collector-detail.png` shows OTLP Receiver detail page
- Confirmed reverting to old values causes the script to throw and exit non-zero

---

## Type of Change

- [x] Bug fix
- [x] CI/tooling

Closes #415


MeloveGupta:
Hi @jaydeluca ,
I've raised this PR and I'd love to have your take on it...

ecosystem-explorer/scripts/take-screenshots.mjs
```diff
@@ -24,9 +24,23 @@ const SCREENSHOTS_DIR = path.resolve("screenshots");
 const PORT = 4173;
 const BASE_URL = `http://localhost:${PORT}`;
 
-// Pick an instrumentation known to have telemetry and configuration data
-const DETAIL_VERSION = "2.25.0";
+// Resolve latest versions at runtime from the generated data files.
+// This prevents the script from going stale when new versions are released.
+function resolveLatestVersion(indexPath) {
```
vitorvasc, line 29: nice catch :)

MeloveGupta:
Hey @vitorvasc ,
I've made the suggested changes from Copilot, could you review them once...

jaydeluca (approved):

PR #422, opened 2026-05-10
feat: implement global java agent config explorer with declarative toggle

### LINKED ISSUE #152: [Explorer] Java Configuration options explorer
As a user of the Java agent, I would like a way to look through all available config options.

There should be an option to select the version of the agent you are using and the config should use the options available for that specific version.

[(Minimal) Example in the POC](https://jaydeluca.github.io/instrumentation-explorer/configurations):

Image

DCchoudhury15:
Closes #152

Here is the implementation for the global Configuration Options Explorer for the Java Agent. This gives users a single place to search all configurations across all instrumentations and toggle between standard system properties and declarative YAML.

### Key Changes
* **UI & Search:** Built `JavaConfigurationListPage` with deep search (filters by name, description, declarative name, and instrumentation) and a view toggle for YAML blocks.
* **Data Aggregation:** Added `aggregate-configurations.cjs` to scrape the individual instrumentation JSONs, deduplicate the configs, and map which instrumentations use each property.
* **Build Pipeline:** Hooked the script into `prebuild` in `package.json` so it generates the global JSON automatically during Vercel/Netlify deployments.

### Review Updates
* **Removed Version Dropdown:** Dropped the "Agent Version" selector. Our current JSON pipeline doesn't actually group configs by agent release versions yet, so the dropdown was non-functional. We can add it back when the data supports it.
* **Local Dev Error State:** Fixed the silent fetch failure. Since `prebuild` doesn't run automatically during `bun run serve`, local development without a prior build would just show an empty search page. I added an explicit error state to tell devs: *"Data aggregation required... Try running the build script first."*
* Cleaned up dead variables, leftover comments, and trailing newlines in the aggregation script.

### Screenshots
WhatsApp Image 2026-05-10 at 20 10 08

.WhatsApp Image 2026-05-10 at 20 10 25


DCchoudhury15:
Hi @jaydeluca ,  a couple of things I'd love 
your thoughts on:

1. I held off on the version dropdown for now since the current data 
   structure doesn't map configs to agent versions  would you like me to 
   explore a different approach for that?
2. The aggregation script is hooked into `prebuild` but not `bun run serve` 
    I handled this with an explicit error state, but happy to do it 
   differently if you have a preference.

Looking forward to your feedback! Thanks.

ecosystem-explorer/package.json
```diff
@@ -9,6 +9,7 @@
   ],
   "scripts": {
     "serve": "vite",
+    "prebuild": "node aggregate-configurations.cjs",
     "build": "tsc -b && bun scripts/generate-schemas.mjs && vite build && bun scripts/generate-agent-docs.mjs",
```
jaydeluca: we aren't invoking `prebuild` anywhere, so the deploy preview isn't showing anything

image


how about we just add it to the `build` command? and we should use bun instead of node

ecosystem-explorer/src/features/java-agent/java-configuration-list-page.tsx
```diff
@@ -13,36 +13,164 @@
...
+import { ConfigurationCard, type ConfigurationFormat } from "./components/configuration-card";
+import type { Configuration } from "@/types/javaagent";
+
+interface GlobalConfiguration extends Configuration {
+  instrumentations?: string[];
+}
+
+const FORMAT_TABS = [
+  { value: "system-property", label: "Normal View (System Properties)" },
+  { value: "declarative", label: "Declarative Representation" },
```
jaydeluca:
```suggestion
  { value: "system-property", label: "System Properties" },
  { value: "declarative", label: "Declarative Configuration" },
```

let's make declarative the default view too

ecosystem-explorer/src/features/java-agent/java-configuration-list-page.tsx
```diff
@@ -13,36 +13,164 @@
...
+];
 
 export function JavaConfigurationListPage() {
+  const [format, setFormat] = useState<ConfigurationFormat>("system-property");
+  const [searchQuery, setSearchQuery] = useState("");
+  const [allConfigurations, setAllConfigurations] = useState<GlobalConfiguration[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [error, setError] = useState<string | null>(null);
+
+  // Fetch the newly aggregated configurations
```
jaydeluca:
```suggestion
```

jaydeluca (commented):
> I held off on the version dropdown for now since the current data
structure doesn't map configs to agent versions would you like me to
explore a different approach for that?

we should incorporate the ability to toggle between versions. If we need to alter the explorer-db-builder to provide a different representation of the data, we could also explore that

> The aggregation script is hooked into prebuild but not bun run serve
> I handled this with an explicit error state, but happy to do it
> differently if you have a preference.

i left a comment, i think we should just hook it into the standard build script. If we want to also incorporate some other kind of testing harness to catch situations where the process fails, let's handle that some other way

DCchoudhury15:
Hi @jaydeluca , thanks for the thorough review! I've addressed all the feedback and pushed the updates for the build scripts, error states, and UI. Before we wrap this up , would you like me to bundle the integration tests into this PR, or track them in a fast-follow?

ecosystem-explorer/scripts/aggregate-configurations.mjs
```diff
@@ -0,0 +1,74 @@
...
+              allConfigs.set(config.name, {
+                ...config,
+                instrumentations: [instrumentationName],
+              });
+            } else {
+              const existing = allConfigs.get(config.name);
+              if (!existing.instrumentations.includes(instrumentationName)) {
+                existing.instrumentations.push(instrumentationName);
+              }
+            }
```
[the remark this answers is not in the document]
jaydeluca: this can be ignored for now, soon all configs will have a `declarative_name`

ecosystem-explorer/src/features/java-agent/java-configuration-list-page.tsx
```diff
@@ -13,36 +13,168 @@
...
 
 export function JavaConfigurationListPage() {
+  const [format, setFormat] = useState<ConfigurationFormat>("declarative");
+  const [searchQuery, setSearchQuery] = useState("");
+  const [allConfigurations, setAllConfigurations] = useState<GlobalConfiguration[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+  const [error, setError] = useState<string | null>(null);
+
+  useEffect(() => {
+    fetch("/data/javaagent/global-configurations.json")
```
jaydeluca: we shouldn't use`fetch()` directly instead of the project-standard `fetchWithCache` + `STORES` from `idb-cache`. The other static data files in the codebase go through a cached fetch layer (`javaagent-data.ts`, `configuration-data.ts`, `collector-data.ts`). 

The fix is:
1. Add a `STORES.GLOBAL_CONFIGURATIONS` (or reuse an appropriate existing store) to `idb-cache.ts`
2. Add a `loadGlobalConfigurations()` function to `javaagent-data.ts` using `fetchWithCache`
3. Call that from the page (ideally via a custom hook to follow the `use-*` convention)

ecosystem-explorer/aggregate-configurations.cjs
```diff
@@ -0,0 +1,95 @@
+const fs = require("fs");
```
jaydeluca: All the other build scripts (`generate-schemas.mjs`, `generate-agent-docs.mjs`, `take-screenshots.mjs`) live in `ecosystem-explorer/scripts/` and use ES module syntax (`.mjs`). Can you move this to `scripts/aggregate-configurations.mjs` and convert it to ESM for consistency?

ecosystem-explorer/src/features/java-agent/java-configuration-list-page.tsx
```diff
@@ -13,36 +13,168 @@
...
+        return res.json();
+      })
+      .then((data) => {
+        setAllConfigurations(data);
+        setIsLoading(false);
+      })
+      .catch((err) => {
+        console.error("Failed to fetch configurations:", err);
+        setError(
+          "Data aggregation required: Global configuration data could not be loaded. Try running the build script first."
```
jaydeluca: End users won't know what a "build script" is. This message should be something like "Configuration data could not be loaded. Please try again later." The "run the build script" hint is useful for local dev but not for deployed environments.

ecosystem-explorer/src/features/java-agent/components/configuration-card.tsx
```diff
@@ -96,6 +97,15 @@ export function ConfigurationCard({ config, format }: ConfigurationCardProps) {
             ))}
           </div>
         )}
+
+        {instrumentations && instrumentations.length > 0 && (
+          <div className="border-border/50 text-muted-foreground mt-4 border-t pt-3 text-xs">
+            <span className="text-foreground font-semibold">
+              {instrumentations.length} Instrumentation{instrumentations.length === 1 ? "" : "s"}
+            </span>{" "}
+            using this configuration
```
jaydeluca: Since search can filter _by_ instrumentation name, users who search for an instrumentation name get matching configs but can't see _which_ of the listed instrumentations matched. A tooltip or expandable list of instrumentation names would be nice to get to see which ones

jaydeluca (changes requested):
> Before we wrap this up , would you like me to bundle the integration tests into this PR, or track them in a fast-follow?

If you'd like to focus on addressing some of my other comments and then look into that later, thats fine by me, but I'll leave the choice up to you.

ecosystem-explorer/src/features/java-agent/java-configuration-list-page.tsx
```diff
@@ -13,36 +13,169 @@
...
-      <div className="space-y-6">
-        <BackButton />
-        <div>
-          <h1 className="mb-2 text-3xl font-bold md:text-4xl">
-            <span className="from-otel-orange to-otel-blue bg-gradient-to-r bg-clip-text text-transparent">
-              OpenTelemetry Java Agent Configuration
-            </span>
-          </h1>
-        </div>
-        {isEnabled("JAVA_CONFIG_BUILDER") ? (
```
jaydeluca, line 34: cc @lucacavenaghi97 this is being removed. Can we move the link to the new config builder to the [java agent landing page?](https://explorer.opentelemetry.io/java-agent)

jaydeluca (approved):

PR #449, opened 2026-05-11
 feat: show page skeleton immediately while java agent list loads

### LINKED ISSUE #445: Improve Java Agent Loading display
When loading the collector list page, the heading and skeleton of the page are loaded immediately, and then there is a nice animated "loading" icon as the list loads.

We should incorporate the same into the java agent list page for consistency, which currently works a bit different

SzymonIwaniuk:
Closes: #445

Description:

The Java Agent instrumentation list page previously replaced the entire page with a plain loading message while data was fetching, hiding the heading and back button until data arrived.

This PR aligns it with the collector page behaviour: the heading, back button, and filter bar now render immediately, while an animated spinner is shown in the list area during loading. Error states are also handled inline rather than as full page replacements.

  Changes:
  • Removed early-return loading/error guards from JavaInstrumentationListPage
  • Always render the page shell (heading, back button, filter bar) on first paint
  • Replaced plain loading text with the animated Loader2 spinner matching the collector page
  • Updated tests to assert the heading is present during both loading and error states
  
  
 Before
 Screenshot 2026-05-11 at 20 11 12
 
 After
Screenshot 2026-05-11 at 20 05 16


lucacavenaghi97:
Nice, lines up with #445 ! Tried it locally with a throttled fetch and a couple of things stood out.

The subtitle reads "Explore 0 available instrumentations." for the whole loading window (and on error too), because the count is bound to `instrumentations?.length ?? 0` and the header now renders before data. Either drop the number from there or render the sentence only once `instrumentations` is non-null.

pr449-loading

The version selector is gated on `versionsData && versionsData.versions.length > 0` so it doesn't render at all during loading, then pops in when data arrives and shifts the layout. The collector keeps the `<select>` mounted and uses `disabled={versionsLoading}`, which is what avoids the shift. Worth mirroring it here, probably needs a `disabled` prop on `VersionSelector`.

collector-loading

Two smaller things: the new generic "Error loading data" no longer shows `error.message`, the previous code did, so a bit of a step back on debuggability. And the new tests assert only the h1, would be nice to also check the back button + filter bar are present in loading/error.

Copilot already flagged the `!resolvedVersion` infinite spinner case, agreed with that one.

SzymonIwaniuk:
Hey @lucacavenaghi97, appreciate your review, I've addressed all suggestions so:
 - subtitle now only renders once instrumentations is non-null
 -  VersionSelector is always mounted with `disabled={versionsLoading}`, (mirrored pattern in collector) 
 - fixed the `!resolvedVersion` infinite spinner
 - extend the loading and error tests + also assert the back button and filter bar are present, and the error test checks error.message is visible
 
Let me know if the changes look right to you. Thanks

jaydeluca (changes requested):
changes LGTM, also thank you @lucacavenaghi97 for the review.

CI failures need to be addressed and then this should be good to go

jaydeluca:
i think this pr is stuck due to a github incident affecting codeql, im going to close and reopen to see if i can un-stuck it

jaydeluca (approved):
thanks!

Issue #450, opened 2026-05-11
Fix Tab List on java instrumentation detail page


jaydeluca:
The tab section ends up with a horizontal scroller on mobile

<img width="527" height="864" alt="Image" src="https://github.com/user-attachments/assets/415efb72-fe73-4fd8-b312-ed2180aeaa49" />

aabhinavvvvvvv:
I can take this up

PR #451, opened 2026-05-11
Fix mobile scrolling in instrumentation detail tabs

vetrovk:
## Summary

Fixes the Java instrumentation detail page tabs so they no longer create a horizontal scroller on mobile.

## Changes

- Use the existing `SegmentedTabList` `fullWidth` layout for the instrumentation detail tabs
- Remove the horizontal overflow wrapper around the tab list
- Compact the tab trigger spacing on small screens so labels fit without scrolling

## Testing

- `git diff --check`

### DIFF
--- a/ecosystem-explorer/src/features/java-agent/instrumentation-detail-page.tsx
+++ b/ecosystem-explorer/src/features/java-agent/instrumentation-detail-page.tsx
@@ -276,9 +276,12 @@ export function InstrumentationDetailPage() {
           </div>
 
           <Tabs value={activeTab} onValueChange={setActiveTab} className="relative z-10">
-            <div className="overflow-x-auto px-4 pt-4 pb-0 sm:px-6">
+            <div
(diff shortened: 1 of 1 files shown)


maryliag:
Hi @vetrovk , thank you for your contribution!
Your description of testing being `git diff --check` doesn't really test your changes, it's only testing if your code have whitespaces errors.

Since this is a UI change, a test would be to share screenshots or video of before/after your solution, so we can properly review it was done correctly.

maryliag:
Hey @vetrovk I just now realized there was an issue created for this [here](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/issues/450) and someone else had already asked to work on this.
We want to be respectful of contributors that have already been assigned or made requests, so I will close your PR in favour of the other one created. When looking to a task to work in the future, please check if no one else is already working on them!
Thank you!

PR #453, opened 2026-05-12
feat(navbar): add v1 navbar (Phase 1 - Foundation PR 2)

vitorvasc:
Contributes to #84. Contributes to #370.

## What's in this PR

- Adds `<NavBar />` at `src/components/layout/nav-bar.tsx`, slotted into `App.tsx` via `isEnabled("V1_REDESIGN") ? <NavBar /> : <Header />`. Renders the logo lockup, a single external "Docs" link to opentelemetry.io, and the theme toggle.
  - `theme-toggle.css` — Bootstrap dropdown chrome with neutral gray surface (`#fff` light, `#343a40` dark), active row blue (always wins over the highlighted state), Bootstrap chevron caret via `::after`
- Rebuilds `<ThemeToggle />` as a 3-option Radix `DropdownMenu` (Light / Dark / Auto).
- Adds `OpenTelemetryWordmark` at `src/components/icons/`, inlining opentelemetry.io's navbar SVG (logo mark + wordmark in one piece).
- `main.tsx` sets `<html data-v1-redesign="true">` at bundle load when the flag is on, before first paint.
- `tokens.css` gates page surfaces on that attribute, switching `--background-hsl` and `--foreground-hsl` to opentelemetry.io's defaults (`#212529` / `#dee2e6` dark, `#fff` / `#212529` light)


jaydeluca:
all of our deploy previews for other changes are going to look off because it shows the new theme by default, which will make debugging /updating the existing theme very challenging as we make more and more changes to the new one.

I think we need to come up with another approach for how we can use deploy previews for this theme revamp work. It might make sense for us to just rely on screenshots / local validation and not enable it by default in deploy previews, or maybe we can have multiple deploy previews with different configs

vitorvasc:
@jaydeluca except for the last PR where we had some changes that weren't gated (specifically the primary/secondary being switched in the CSS file), the changes for this PR aren't supposed to reflect on the preview for other PRs (or the production website).

The deployment previews are gated in the [`.netlify.toml`](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/blob/main/netlify.toml#L5), and the feature flag should only be enabled for branches prefixed with `feat/84-*`. This same rule should also be applied to the screenshots workflow, following the changes introduced in this PR: https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/453/changes#diff-50fe99ab4c6fa87d9580c925f35a3c5a1d21b56dd0694610dad6f154ed12d7c6R52

jaydeluca:
> The deployment previews are gated in the [.netlify.toml](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/blob/main/netlify.toml#L5), and the feature flag should only be enabled for branches prefixed with feat/84-*. This same rule should also be applied to the screenshots workflow, following the changes introduced in this PR: [#453 (changes)](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/453/changes#diff-50fe99ab4c6fa87d9580c925f35a3c5a1d21b56dd0694610dad6f154ed12d7c6R52)


ahhhh i forgot about this piece, thank you!

ecosystem-explorer/src/App.tsx
```diff
@@ -80,7 +81,7 @@ export default function App() {
                   role="status"
                   aria-live="polite"
                 >
-                  <div className="text-muted-foreground text-sm font-medium">Loading...</div>
+                  <div className="text-muted-foreground text-sm font-medium">Loading…</div>
```
jaydeluca, line 83: this is interesting lol

jaydeluca (approved):

vitorvasc:
Just for the record: I've pivoted the Phase 1 mechanism based on what we discussed yesterday at our Comms SIG call (2026-05-12). Instead of feature-flagging individual components within shared files, we're now using the `V1_REDESIGN` feature flag to control a single boundary read in `App.tsx` that swaps out the entire `<V1App />` for `<LegacyApp />`. The v1 redesign code will now live under `src/v1/`. 

This gives us a much cleaner separation between the legacy and redesign code, and makes it easier to manage the transition. The rationale and details of this pivot are documented in the [`v1-routing-pivot.md`](https://github.com/vitorvasc/opentelemetry-ecosystem-explorer/blob/30a0016b7515fb45cce262cd00e8967cadde29e1/projects/84-ui-ux-design/v1-routing-pivot.md).

Issue #394, opened 2026-05-08
[UX/IA] Optimize Information Hierarchy and Reduce Cognitive Overload for Instrumentation Docs


hussainjamal760:
Context
The OpenTelemetry Ecosystem Explorer currently serves instrumentation documentation for over 250 components. While the data layer (JSON) is rich and complete, the current presentation lacks a clear entry point for users, leading to high cognitive load. Technical details (metrics, spans, configs) are presented immediately, often overwhelming first-time users who need to understand the "why" before the "how."

 The Problem
Cognitive Overload: Users are hit with deep technical tables immediately.
Missing Value Proposition: The "What is this?" and "Why does it matter?" context is buried under technical specs.
Flat Hierarchy: No distinction between high-level overview and deep technical reference.
Scalability Strain: The interface needs to handle 250+ variations without per-file manual adjustments.

 Objectives
Implement a Progressive Disclosure strategy.
Establish a clear Information Hierarchy: Overview → Capabilities → Technical Specs.
Improve the Entry Point experience without modifying the underlying 250+ JSON files.
Maintain a completely Data-Agnostic presentation layer.

PR #460, opened 2026-05-13
feat:Expand and collaps button added

hussainjamal760:
issue #394 
# Description: Added Expand All/Collapse All functionality to Telemetry Section

This PR introduces "Expand All" and "Collapse All" controls for the Metrics and Spans sections in the Java Agent instrumentation details page. The goal is to improve information hierarchy and reduce cognitive load by allowing users to toggle the visibility of technical details globally.

###  Features
- **Global Expansion Controls**: Added segmented pill-shaped buttons to expand or collapse all Metrics and Spans simultaneously.
- **Improved Information Hierarchy**: Details are now collapsible, keeping the initial view clean while allowing deep-dives on demand.

###  UI/UX Enhancements
- **Consistent Styling**: Matched the new controls with the application's design system (using `SegmentedTabList` aesthetic: `bg-muted` container and `secondary` active states).
- **Refined Layout**: 
    - Repositioned global buttons below section headers for better flow.
    - Simplified card headers by removing redundant individual chevrons.
    - Restored badge alignment (e.g., `histogram`, `CLIENT Span`) to the right side for better readability.
- **Responsive Spacing**: Adjusted margins and padding for a balanced look on both mobile and desktop.

###  Technical Improvements
- **Performance Optimization**: Replaced `useEffect` state synchronization with inline state adjustment during render to prevent cascading renders.
- **Component Flexibility**: Updated `SectionDivider` to support custom `className` props.
- **Clean Code**: Removed unused functions and applied consistent formatting via Prettier.

## Verification Results
- **Automated Tests**: All 18 unit tests in `telemetry-section.test.tsx` passed successfully.
- **Manual Check**: Verified that the "Expand/Collapse All" interaction is fluid and visual states are consistent with the main navigation tabs.

image


vitorvasc (commented):
Thank you @hussainjamal760! 

We could also consider adding controls to collapse or expand a single element when clicking on this section, as well as the visual indicators for that:

Image

What do you think?

hussainjamal760:
> Thank you @hussainjamal760!
> 
> We could also consider adding controls to collapse or expand a single element when clicking on this section, as well as the visual indicators for that:
> 
> Image
> What do you think?

Yeah, I understand your point. But I think having `http.client.request.duration`, the badge, and the expand/collapse button all in the same line might make the UI feel a bit messy and crowded. Maybe we could keep the interaction simpler or separate the controls slightly to maintain readability.

what do you think?

vitorvasc:
> Yeah, I understand your point. But I think having `http.client.request.duration`, the badge, and the expand/collapse button all in the same line might make the UI feel a bit messy and crowded. Maybe we could keep the interaction simpler or separate the controls slightly to maintain readability.
> 
> what do you think?

That makes sense. If the visual indicator feels like too much information, we can skip it and use `cursor: pointer` when hovering that section instead. I believe this still indicates to users that they can interact with it. 

My rationale for allowing elements to be collapsed independently is that some libraries have a lot of metrics, for example: https://deploy-preview-460--otel-ecosystem-explorer.netlify.app/java-agent/instrumentation/2.27.0/grpc-1.6. 

While collapsing all elements indeed reduces the cognitive load while navigating the page, the usability still isn't quite there yet, since the user currently has to either collapse everything or expand everything. If they want to see the details for a single metric at the bottom of the page, for example, they still have to expand everything and scroll through a bunch of information they don't actually need.

hussainjamal760:
> > Yeah, I understand your point. But I think having `http.client.request.duration`, the badge, and the expand/collapse button all in the same line might make the UI feel a bit messy and crowded. Maybe we could keep the interaction simpler or separate the controls slightly to maintain readability.
> > what do you think?
> 
> That makes sense. If the visual indicator feels like too much information, we can skip it and use `cursor: pointer` when hovering that section instead. I believe this still indicates to users that they can interact with it.
> 
> My rationale for allowing elements to be collapsed independently is that some libraries have a lot of metrics, for example: https://deploy-preview-460--otel-ecosystem-explorer.netlify.app/java-agent/instrumentation/2.27.0/grpc-1.6.
> 
> While collapsing all elements indeed reduces the cognitive load while navigating the page, the usability still isn't quite there yet, since the user currently has to either collapse everything or expand everything. If they want to see the details for a single metric at the bottom of the page, for example, they still have to expand everything and scroll through a bunch of information they don't actually need.

That’s a fair point, especially for instrumentations with a large number of metrics. I agree that allowing individual sections to be expanded/collapsed would improve navigation and make it easier for users to focus only on the information they need.

I still think we should be careful with the visual density of the header row, but using `cursor: pointer` instead of adding another explicit indicator sounds like a clean compromise. That could keep the UI lightweight while still making the interaction discoverable.

I’ll explore an approach where individual items can be toggled independently without overcrowding the layout.

jaydeluca (changes requested):
the feature itself looks good, can you address the copilot comments that I gave a " 👍 " to? then this should be good to go

hussainjamal760:
> the feature itself looks good, can you address the copilot comments that I gave a " 👍 " to? then this should be good to go

Sure ,

vitorvasc (approved):

jaydeluca (approved):

PR #467, opened 2026-05-13
feat: add v1-registry-sync dry-run tool for syncing V2 data into V1 entries

### LINKED ISSUE #465: feat: sync stability and description from V2 registry into existing V1 entries
## Background

Follow-up from the research in #119. V1 registry entries have no stability field, 
so users visiting opentelemetry.io cannot tell whether a component is alpha, beta, 
or stable. V2 tracks per-signal stability for every component it covers.

## What to build

A new script or workflow that reads stability, display_name, and description from 
V2 registry data and writes those values into the matching V1 entries under 
data/registry/ in the opentelemetry.io repo. Changes would be submitted as a PR 
for review.

This is a new pipeline separate from collector-sync (which writes to data/collector/ 
for Hugo shortcodes, not to data/registry/) and separate from the otelbot nightly 
job (which handles package.version via the Go module index).

## Related

Follow-up from #119. Pairs with the auto-generate issue above.

Rama542:
## Summary

This PR adds a new ecosystem-automation/v1-registry-sync Python package as a 
first step toward issue #465.

The tool reads the latest V2 registry snapshot from ecosystem-registry/collector/ 
and generates a report showing exactly which stability, display_name, and 
description values would be written into matching V1 entries under 
opentelemetry.io/data/registry/. For stability, it selects the most stable signal 
level across all signals per component (stable > beta > alpha > development > 
deprecated > unmaintained).

This version runs in dry-run mode only. The actual write step to opentelemetry.io 
would require coordination with the opentelemetry.io maintainers on the V1 schema 
and PR workflow, and is left for a follow-up.

## What was added

- models.py: ComponentSyncData and V1SyncReport dataclasses
- reader.py: reads V2 YAML files and extracts fields for V1 sync
- reporter.py: serializes the report to JSON or YAML
- main.py: CLI entry point (uv run v1-registry-sync)
- 18 unit tests covering the reader and reporter modules

## How to run

From the repo root:
uv run v1-registry-sync --distribution contrib --format json

Closes #465


lucacavenaghi97:
Hi @Rama542, a few things worth checking before this lands.

**1. `stability` is not in the V1 schema**

[`opentelemetry.io/data/registry-schema.json`](https://github.com/open-telemetry/opentelemetry.io/blob/main/data/registry-schema.json) has no `stability` property and declares `additionalProperties: false` at the top level. The validator in `gulp-src/validate-registry.js` would reject any V1 entry containing `stability`, so the `"stability": "<level>"` value the report currently emits in `proposed_v1_changes` would not be applicable today as-is.

**2. The report does not say which V1 entry would be updated**

Each entry reads as `{"name": "kafkareceiver", "proposed_v1_changes": {...}}` but does not include the target V1 file path or whether a matching V1 entry exists. That is the piece a follow-up writer would need to act on, and it is also where most of the design difficulty of the sync lives (matching across naming conventions, handling renames and deprecations, skipping third-party distribution entries).

It could be worth adding a `target_v1_file` plus `v1_entry_exists` flag to each entry, so the dry-run becomes more directly actionable.

**3. Could reuse `InventoryManager` from `watcher-common` / `collector-watcher`**

[`watcher-common`](../ecosystem-automation/watcher-common/) exposes `BaseInventoryManager` with `list_versions`, `list_release_versions`, `list_snapshot_versions`. [`collector-watcher.InventoryManager`](../ecosystem-automation/collector-watcher/src/collector_watcher/inventory_manager.py) extends that with the `distribution` axis and `load_versioned_inventory(distribution, version)`, which already does what `_find_latest_version` + the per-type yaml loop in `reader.py` does. The same pattern is used by `configuration-watcher`, `java-instrumentation-watcher`, and (as a non-watcher consumer) `explorer-db-builder`.

Picking up `list_release_versions(distribution)` instead of the local helper would also fix a latent issue: `_find_latest_version` currently returns the highest semver-sorted directory name including `v0.152.1-SNAPSHOT`, which means the tool can pick up unreleased data. `list_release_versions` excludes snapshots.

**Smaller notes**

- `display_name` would overwrite V1 `title`. From a 30-component sample most divergences are cosmetic (V1 adds a "Collector" infix), but a handful of entries lose real information (e.g. `otelarrowexporter`: V1 says "OpenTelemetry Protocol with Apache Arrow Exporter", V2 says "OpenTelemetry Arrow Exporter"). Limiting the initial sync to `description` might be safer.
- No `README.md` under `v1-registry-sync/` yet; the sibling watcher packages have one.

lucacavenaghi97:
Six changes look good as applied!

One follow-up on `target_v1_file`. Current `f"collector-{name}.yml"` builds paths like `collector-kafkareceiver.yml`, but actual V1 files are `collector-{component_type}-{slug}.yml`. So `v1_entry_exists` is returning `false` for essentially every component even though most do exist in V1.

In my opinion the cleanest fix would be to join on the Go module path, which both registries already carry:
- V1 stores it explicitly as `package.name` (e.g. `github.com/open-telemetry/opentelemetry-collector-contrib/receiver/kafkareceiver`)
- V2 carries the same value implicitly: `github.com/open-telemetry/opentelemetry-collector-contrib/{component_type}/{name}`

Building a `dict[go_module_path -> v1_file_path]` once at startup by reading the `package.name` field of each V1 file, then looking up the expected path per V2 component would side-step the naming inconsistencies entirely. Across the 249 contrib components in v0.151.0, 244 match this way; 5 are genuinely missing in V1 (`azurefunctionsreceiver`, `googlesecopsexporter`, `drainprocessor`, `spanpruningprocessor`, `datadogconnector`, all recent additions). Those misses would be real signals rather than matcher bugs.

Including `expected_go_module_path` on every row in the report (matched or not) might also help triage when a match fails. Happy to discuss other approaches if you see a simpler path.

One related note: the existing `test_target_v1_file_follows_naming_convention` test asserts `collector-fooreceiver.yml` for `fooreceiver`, so it currently passes against the wrong convention. Worth replacing with a fixture that mirrors a realistic V1 naming if you go this route.

Rama542:
Good catch, thank you. The naming convention approach was broken because actual V1 files follow collector-{component_type}-{slug}.yml, not collector-{name}.yml, so v1_entry_exists was returning false for almost everything.

Switched to matching on Go module path instead. At startup the tool now reads every .yml file in --v1-registry-dir and builds a {package.name -> filename} index. Each V2 component's expected module path is constructed as github.com/open-telemetry/opentelemetry-collector-contrib/{component_type}/{name} and looked up in the index to find the actual V1 file. expected_go_module_path is included on every row so misses are easy to triage.

I also updated the test fixture to use a realistic V1 filename (collector-receiver-fooreceiver.yml) so the test covers the actual matching logic rather than the old wrong convention.

lucacavenaghi97 (approved):
The Go module path matching reads as I'd expect, and all the follow-up points from both rounds are now addressed. Approving.

Worth merging `main` into the branch before this lands. It's a few commits behind.

PR #469, opened 2026-05-13
fix: prevent partial registry push when a watcher fails

### LINKED ISSUE #468: bug: nightly workflow can commit and push partial registry data when a watcher fails
## What's wrong

In `nightly-registry-update.yml`, all three watchers run with `continue-on-error: true`:

```yaml
- name: Run collector-watcher
  continue-on-error: true

- name: Run java-instrumentation-watcher
  continue-on-error: true

- name: Run configuration-watcher
  continue-on-error: true
```

The commit and push step that follows has no check on whether the watchers actually succeeded:

```yaml
- name: Commit and push changes
  if: steps.check_changes.outputs.has_changes == 'true'
  run: |
    git push -f origin HEAD:${BRANCH}
```

The only condition is `has_changes == 'true'` — it does not check watcher outcomes at all.

## The problem

If one watcher fails (e.g. collector-watcher hits a GitHub API rate limit or timeout),
the other two watchers still succeed and write their data. The `check_changes` step
sees changes and sets `has_changes=true`. The push step then force-pushes a partial
registry — missing all data from the failed watcher — to the remote branch. A PR is
auto-created and can be merged into main.

The result is that the frontend ends up serving incomplete or missing data to users
for the ecosystem that failed, with no indication that anything went wrong.

The `notify-*` jobs do open GitHub issues when a watcher fails, but the push already
happens before those notifications fire.

## Suggested fix

Add watcher outcome checks to the push step condition:

```yaml
- name: Commit and push changes
  if: |
    steps.check_changes.outputs.has_changes == 'true' &&
    steps.collector_watcher.outcome == 'success' &&
    steps.java_instrumentation_watcher.outcome == 'success' &&
    steps.configuration_watcher.outcome == 'success'
```

This ensures the registry is only updated when all watchers complete successfully.

## File affected

`.github/workflows/nightly-registry-update.yml`

SurbhiAgarwal1:
## What's wrong

In `nightly-registry-update.yml`, all three watchers run with `continue-on-error: true`.
The commit and push step that follows has no check on whether the watchers actually succeeded:

Fixes #468

```yaml
- name: Commit and push changes
  if: steps.check_changes.outputs.has_changes == 'true'

### DIFF
--- a/.github/workflows/nightly-registry-update.yml
+++ b/.github/workflows/nightly-registry-update.yml
@@ -223,7 +223,11 @@ jobs:
           fi
 
       - name: Commit and push changes
-        if: steps.check_changes.outputs.has_changes == 'true'
+        if: |
+          steps.check_changes.outputs.has_changes == 'true' &&
+          steps.collector_watcher.outcome == 'success' &&
+          steps.java_instrumentation_watcher.outcome == 'success' &&
+          steps.configuration_watcher.outcome == 'success'
         env:
           GH_TOKEN:
             ${{ steps.repo_check.outputs.is_primary == 'true' && steps.otelbot-token.outputs.token


jaydeluca:
this is intentional behavior, see: https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/issues/142

We want partial PRs to be created, and an issue is automatically raised for an individual watcher that fails, so we are alerted and can followup

PR #471, opened 2026-05-13
docs: add research findings for license and tags auto-derivability in V1 registry

### LINKED ISSUE #461: research: investigate whether license and tags fields can be auto-derived for V1 registry entries
## Background

As part of the research in #119, Proposal B suggests auto-generating V1 registry 
entries for new collector components using V2 data. Before that can be built, two 
fields need investigation since they exist in V1 but not in V2:

## What needs investigating

**license field**
All contrib components in opentelemetry-collector-contrib are Apache 2.0. If that 
holds true for every component V2 tracks, the license field can be hardcoded or 
inferred during generation and does not need a manual sidecar.

**tags field**
V1 uses tags like go, collector, and receiver for search and filtering on 
opentelemetry.io. It needs to be checked whether these tags are predictable 
from component type and language alone, or whether they require manual input 
per component.

## Why it matters

If both fields can be derived automatically, the only fields still needing human 
input when auto generating a V1 entry would be authors and createdAt. That makes 
the review step much lighter and brings Proposal B closer to being practical.

## Related

Follow up from #119 and PR #446.

Rama542:
This PR adds the research findings for issue #461.

The investigation checked existing V1 registry entries at 
opentelemetry.io/data/registry across all five component types (receiver, 
exporter, processor, extension, connector) to answer whether the license 
and tags fields can be generated automatically when creating new V1 entries 
from V2 data.

The short answer is yes for both.

Every single collector entry in V1 uses license: Apache 2.0. All 
opentelemetry-collector and opentelemetry-collector-contrib components are 
Apache 2.0 by project policy, so this can just be hardcoded during generation.

For tags, every entry follows the same three-tag formula: go, the component 
type, and collector. Since the component type is already in V2 metadata and 
the other two are fixed for all collector components, the entire tags list 
is predictable without any manual input.

With these two fields confirmed as auto-derivable, the only field that truly 
requires a human reviewer when generating a new V1 entry is authors. The 
createdAt field can be approximated from git history. This makes Proposal B 
from issue #119 much lighter to implement than originally estimated.

Closes #461

### DIFF
--- a/projects/461-license-tags-research/00-research.md
+++ b/projects/461-license-tags-research/00-research.md
@@ -1,0 +1,117 @@
+---
+title: "Research — License and Tags Auto-Derivability for V1 Registry Entries"
+issue: 461
+type: audit
+phase: 1
+status: complete
+last_updated: "2026-05-13"
+---
+
+## Overview
+
+This document captures the research findings for
+[issue #461](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/issues/461). The
+question is whether the `license` and `tags` fields in V1 registry entries can be derived
+automatically from available data, or whether they require manual input per component.
+
+This matters because it directly affects how much human review is needed when auto-generating new V1
+entries from V2 data (Proposal B from
+[issue #119](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/issues/119)).
+
+---
+
+## Method
+
+A sample of existing V1 registry entries was inspected from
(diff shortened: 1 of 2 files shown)


jaydeluca (approved):
looks good. perhaps as part of the next step, we could add these fields to the components metadata in the V2 registry as part of the watcher?

Rama542:
Hi @jaydeluca Thank you for the feedback and for merging this. That is a great suggestion, I will look into adding these fields to the V2 registry component metadata as part of the watcher in a follow up.

Issue #459, opened 2026-05-12
Declarative config builder final phase


jaydeluca:
Final round of thoughts for the declarative configuration builder before we make it live - feel free to push back on things if you have other ideas or feel strongly about not doing them:

- [x] Let's put a disclaimer somewhere at the top of the page that this is a new feature and there could be bugs, and we encourage people to open issues if they encounter any problems.
- [x] Add a link to the [opentelemetry.io declarative configuration documentation](https://opentelemetry.io/docs/zero-code/java/agent/declarative-configuration/) somewhere
- [x] Could we also provide a quick helper with how to pass the file `-Dotel.config.file=/path/to/otel-config.yaml`, maybe even in the header comment in the output file
- [x] Add comment in the output file that indicates which version of the java agent instrumentation agent configs it was based on
- [x] Update the default starting point to [mirror the example](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/otel-getting-started.yaml), so we need to add: 
  - [x] [resource block with detectors](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/otel-getting-started.yaml#L13-L22), including the attributes referencing the env var
  - [x] [sampler](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/otel-getting-started.yaml#L33-L36)
  - [x] Populate the endpoint values by default with the [examples too](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/otel-getting-started.yaml#L41)
- [x] I think we are still missing the individual instrumentation configs? we have the enable/disable but need the rest
  - [x] The declarative_names will be backfilled by next week - the new java agent release has all fields backported
- [x] Remove feature flag 🚀  gating

Wishlist / Stretch (can be tracked as separate followup issues):
- [x] At some point we'll want a java agent version selector, but im happy with keeping this pointed at latest and creating an issue to add that as a followup
- [x] "expand all" / "collapse all buttons for all the SDK/instrumenation components
- [x] Ability to expand/collapse sections so if you want to expand the output YAML to take up more of the screen
- [x] state tracking such that whatever section you last interacted with (whether its typing in a field, clicking a button, expanding / contracting) was highlighted in the YAML, its clear what section you're working on.

djvu2k6:
[...]

PR #474, opened 2026-05-13
feat(config-builder): final-phase polish (#459)

lucacavenaghi97:
Final polish for #459: Beta badge + docs/issue links in the header, `-Dotel.config.file=...` helper and Java agent version in the YAML header, starter template mirroring upstream `otel-getting-started.yaml` (including sampler), drop the `file_format:` schema-description comment, bump `STORAGE_KEY`, remove the `JAVA_CONFIG_BUILDER` feature flag. Also a small `findNodeByPath` fix to handle `circular_ref` nodes and a `!version` gate on the tab content to avoid a transient error flash.

**Per-instrumentation forms** (from #341 and #413) already render for libraries whose configs have `declarative_name` populated. v2.27.0 is at ~27%, upstream main at ~99%. Once v2.27.1 ships and the inventory updates, the remaining libraries will populate with no further changes here.


ecosystem-explorer/src/hooks/use-configuration-builder.ts
```diff
@@ -53,7 +53,7 @@ import {
   validateAll as validateAllNodes,
 } from "@/lib/config-validation";
 
-const STORAGE_KEY = "otel-config-builder-state-v2";
+const STORAGE_KEY = "otel-config-builder-state-v3";
```
jaydeluca, line 56: no need to change this but as a note, we can also bump the overall [db schema version](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/blob/main/ecosystem-explorer/src/lib/api/idb-cache.ts#L19) which would have a similar effect

jaydeluca (approved):
I'm sure we'll have some iterations and tweaks but I think this is good to ship. amazing work on this!

PR #484, opened 2026-05-14
feat(v1): Footer (Phase 1 - Foundation PR 6)

vitorvasc:
Swaps the legacy `<Footer />` placeholder in `<V1App />` for the v1 chrome. Lands the two-cluster Docsy-style footer plus the OTel-purple CNCF callout above it.

Contributes to #84.

## What's in this PR

- `<FooterV1 />` at `src/v1/components/layout/footer.tsx` mirrors opentelemetry.io's `.td-footer`. Dark surface in both themes per upstream's `.td-box--dark` rule, with OTel-orange `#f5a800` hover instead of Docsy's auto-mixed blue.
- `<CncfCallout />` at `src/v1/components/layout/cncf-callout.tsx`. OTel-purple `td-box--secondary` band above the footer with the upstream CNCF wordmark inlined as `<CncfLogo />` (uses `role="img"` plus `aria-label` since it's not inside a labeled link) and a Linux Foundation attribution line
- Scoped CSS partials at `src/v1/styles/footer.css` and `src/v1/styles/cncf-callout.css`. Both use `.td-footer__*` / `.td-cncf-callout__*` BEM hierarchies that only match elements rendered by `<V1App />`, so they stay inert in the legacy bundle. `.td-footer__container` and `.td-cncf-callout__container` both cap at `max-width: 1320px` (upstream Bootstrap `.container-xxl`)
- `<V1App />` drops the legacy `<Footer />` import + JSX and mounts `<CncfCallout />` + `<FooterV1 />` instead. JSDoc updated to drop the PR 6 placeholder note
- `src/v1/styles/index.css` picks up the two new `@import` lines plus a comment block listing every v1 chrome partial

## What's not in this PR

- Phase 2-5 route swaps inside `<V1App />` (HomeV1, list page, detail page, etc.) belong to their respective phase PRs. Only the footer swap is here

## Verification

- `bun run typecheck` clean
- `bun run lint` clean
- `bun run test src/v1/components/layout/footer.test.tsx src/v1/components/layout/cncf-callout.test.tsx` passes 11/11
- `bun run test` passes 848/848 across 87 files
- `bun run format -- src/v1 projects` no-op on in-scope paths

---

> Co-authored with Claude Opus 4.7.


lucacavenaghi97 (approved):
LGTM

PR #485, opened 2026-05-14
fix(explorer): run aggregate-configurations in dev server

lucacavenaghi97:
## Summary

- `bun run serve` now runs `aggregate-configurations.mjs` before Vite, mirroring `build` / `netlify-build:*`. Without this, `/java-agent/configuration` fails in local dev because `global-configurations.json` is missing and Vite serves the SPA fallback, which the page tries to parse as JSON.
- Commits the generated `public/data/javaagent/global-configurations.json` so a fresh clone of `main` works without running the aggregator manually. Follows the project convention that everything under `public/` is tracked.

### DIFF
--- a/ecosystem-explorer/package.json
+++ b/ecosystem-explorer/package.json
@@ -8,7 +8,7 @@
     "esbuild"
   ],
   "scripts": {
-    "serve": "vite",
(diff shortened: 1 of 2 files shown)


jaydeluca (approved):
thanks!

PR #487, opened 2026-05-14
feat(v1): visual regression + a11y (Phase 1 - Foundation PR 7)

vitorvasc:
Adds an isolated component showcase, integrates axe-core into the capture rotation, ships a standalone pixel-diff script with pixelmatch, and wires a main-branch baseline that every PR comment now diffs against.

Contributes to #84 and #370.

## What's in this PR

- New dev-only `/_dev/components` route in `src/v1/features/_dev/components-page.tsx`. Mounts the design-system primitives in their canonical states. Gated by a new `DEV_SHOWCASE` feature flag in `src/lib/feature-flags.ts`. Off in production, on in `.env.development` and in the CI screenshot builds. URL-only, no nav entry
- New `scripts/diff-screenshots.mjs` standalone tool using `pixelmatch` + `pngjs`. Walks a baseline and a current directory of PNGs, writes per-file diff images, and reports matched / changed / new / missing as JSON. Tunable via `PIXELMATCH_THRESHOLD` (default 0.1 per pixel) and `DIFF_RATIO_THRESHOLD` (default 0.001 per file). Exits non-zero when any file is new, missing, or over budget so CI can gate on it. `readPng` tolerates corrupt or zero-byte PNGs by classifying them as `"unreadable"` rather than aborting
- `scripts/take-screenshots.mjs` now runs `@axe-core/playwright` once per page-theme combination (viewports skipped since they don't change a11y semantics). Writes a per-page violations report to `a11y/{name}-{theme}.json` and prints a summary at the end. The capture rotation gains an 8th step for the dev showcase page
- New `.github/workflows/screenshots-baseline.yml` triggers on push to `main` (paths-filtered to the explorer surface), runs the same capture machinery, and commits the result to `otelbot/screenshots/main/`. This is the baseline every PR's diff compares against
- `.github/workflows/screenshots-capture.yml` extended to fetch the `main` baseline from `otelbot/screenshots` (when present), run `diff-screenshots.mjs` against the PR's captures, and upload the diff PNGs + report as a second artifact. No-op when the baseline doesn't exist yet
- `.github/workflows/screenshots-commit.yml` extended to download the diff artifact (continue-on-error if absent), commit any diff PNGs to `otelbot/screenshots/{PR}/diffs/main/`, and extend the existing PR comment with a conditional "Visual diffs vs main" section. The section is suppressed entirely when no files exceed the per-file ratio threshold, so PRs without visual changes get the same comment they get today. Single comment per PR, upserted via the existing `<!-- screenshots-bot -->` marker
- Python comment-builder extracted from the inline workflow heredoc into `.github/scripts/build-screenshots-comment.py`. Testable locally with env vars and lintable in isolation
- Package scripts `screenshots:capture` and `screenshots:diff` for consistent invocation
- devDeps added: `@axe-core/playwright`, `pixelmatch`, `pngjs`, `@types/pngjs`

## What's not in this PR

- Pixel-diff threshold tuning. The defaults (0.1 per pixel, 0.001 changed-pixel ratio) are conservative starting points. Once the baseline lands and a few PRs run against it, the thresholds can move based on real signal
- Reusable workflow extraction for the shared capture preamble (checkout, setup-bun, install deps, install Playwright, build). The capture and baseline workflows share ~20 lines of preamble; factoring into a `workflow_call` reusable workflow earns ~70 LOC dedup. Tracked as a follow-up so this PR stays scoped to behaviour
- Parallelizing the diff loop with `Promise.all` to drop ~1-2 min to ~10-20 sec. Within CI budget today, deferred as a free win
- Adding SubNav, TypeStripe, FooterV1, and CncfCallout to the showcase. They live on in-review PRs and get added as those PRs merge

## Verification

- `bun run typecheck` clean
- `bun run lint` clean
- `bun run test` passes 837/837 across 85 files
- `bun run build` succeeds; `/_dev/components` is a lazy-loaded chunk that ships at a few KB and only registers as a route when `DEV_SHOWCASE` is enabled
- YAML parse OK on all three workflow files
- `build-screenshots-comment.py` smoke-tested locally with three scenarios:
  - No diff report present: renders 141 lines, no diff section
  - Diff report with one sub-threshold change + one over-budget + one new file: only the over-budget and new rows surface, the sub-threshold row is filtered out, summary line reads "1 over budget, 1 new, 0 missing"
  - Malformed diff report: degrades gracefully to "no diff section"

---

> Co-authored with Claude Opus 4.7.


lucacavenaghi97:
Quick question on the `DEV_SHOWCASE` gating. It's on in `.env.development` and in the CI screenshot workflows, but `netlify.toml` doesn't extend the existing `feat/84-*` block to flip it on, so `/_dev/components` 404s on the deploy preview. As a reviewer I can't click through to inspect axe live or check focus indicators without running it locally.

Would it make sense to mirror the `V1_REDESIGN` pattern here? Something like:

```toml
if [[ "$HEAD" == feat/84-* || "$BRANCH" == feat/84-* ]]; then
  export VITE_FEATURE_FLAG_V1_REDESIGN=true
  export VITE_FEATURE_FLAG_DEV_SHOWCASE=true
fi
```

Or is the intent to keep the showcase strictly local + CI? Either way works, just wanted to surface it.

vitorvasc:
@lucacavenaghi97 good catch! Rule added to the `netlify.toml` as well: https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/487/commits/6d3c7ed9013fe5ab5e28eb96746e4c7e403cd824

lucacavenaghi97 (approved):

PR #490, opened 2026-05-15
fix(test): replace lazy-load heading assertion in App.test

suhr25:
## SUMMARY

Fix flaky legacy app rendering test in `App.test.tsx` that failed during full test suite execution.

---

## FIX

The test was waiting for a heading rendered inside a lazy-loaded component, which sometimes timed out when all tests ran in parallel. This update replaces that async check with simple synchronous assertions by verifying the absence of `.v1-app` and checking for the `"OTel Explorer"` text. The approach matches the sibling V1 test and makes the test stable and deterministic.

---

## CHECKLIST

* [x] Reproduced issue locally
* [x] Verified fix with full test suite
* [x] Minimal test-only change
* [x] No production code changes

### DIFF
--- a/ecosystem-explorer/src/App.test.tsx
+++ b/ecosystem-explorer/src/App.test.tsx
@@ -23,18 +23,19 @@ describe("App", () => {
     vi.unstubAllEnvs();
   });
 
-  it("renders the legacy app when V1_REDESIGN is disabled", async () => {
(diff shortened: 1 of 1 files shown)


suhr25:
Hi @vitorvasc 
Kindly Review.
Thanks!

vitorvasc (approved):
Thanks!

PR #491, opened 2026-05-15
fix: stop spin button at exclusive bound

suhr25:
## SUMMARY

Fixes an issue where `NumberInputControl` treated `exclusiveMinimum` / `exclusiveMaximum` as HTML `min` / `max` values. This caused the spin button to stop at values that were actually invalid and immediately failed validation.

## FIX

Updated `NumberInputControl` to only use inclusive `minimum` and `maximum` values for the HTML input attributes. Since HTML number inputs treat `min` and `max` as inclusive bounds, using exclusive schema constraints was misleading the UI and allowing users to land on invalid values like `0` for fields that must be greater than `0`. Added a regression test to ensure exclusive-only constraints no longer set HTML `min`/`max` attributes.

## CHECKLIST

* [x] Reproduced issue locally
* [x] Applied fix in `NumberInputControl`
* [x] Added regression test
* [x] Verified tests pass locally


suhr25:
Hi @vitorvasc @jaydeluca,
Kindly review.
Thanks!

jaydeluca:
@suhr25 can you provide steps on how to reproduce this issue? For example, which input do you see this problem occurring on?

jaydeluca:
in order to help keep track of what is ready for review, i'm going to put this into draft state. Feel free to open it back up when you have added reproduction steps.

If you are not able to actually reproduce this issue, then please close the PR.

suhr25:
Thanks, I was able to reproduce this locally.

Steps:
1. Open `/java-agent/configuration/builder`
2. Enable Logger Provider → Processors → Add item → Batch
3. Find "Max Queue Size", enter `1`
4. Click the spin-down arrow once

Before this PR, the spinner stops at `0`, but tabbing away shows "Must be greater than 0".

Root cause:

// before
const min = constraints?.minimum ?? constraints?.exclusiveMinimum;

// after
const min = constraints?.minimum;

HTML `min` is inclusive, while `exclusiveMinimum` is not. A regression test is included in this PR.

jaydeluca:
Thanks for providing reproducer steps and investigating this.

After testing your branch against `main`, I don’t think this change improves the overall UX. Removing the constraint entirely means the spinner no longer has a lower bound, so users can decrement to `-1`, `-100`, etc. until validation runs on blur. In practice, that behavior seems worse than stopping at `0`.

More generally, I want to reiterate feedback I’ve given on a few previous PRs: please try to focus issues and contributions on problems you actually encountered while using the site, or on cases that are likely to impact real users. Many edge cases are technically valid, but every issue and PR still takes maintainer time to investigate, validate, discuss, and maintain.

Right now, a number of the recent PRs feel very speculative or extremely low-impact, which makes it difficult for us to prioritize review time effectively.

Thanks again for taking the time to contribute, but I don’t think we’ll move forward with this change.

PR #498, opened 2026-05-16
fix: use (name, subtype) tuple as component key in DeprecationDetector

### LINKED ISSUE #493: DeprecationDetector uses component name alone as unique key, ignoring subtype — can produce false deprecations for nested extension components
## Bug Description

In `deprecation_detector.py`, `_build_component_set` identifies components solely by `name`:

```python
return {component["name"] for component in components}
```

However, collector extension components have a `subtype` field (`"encoding"`, `"observer"`, `"storage"`), introduced by the nested component scanning in `component_scanner.py`. Two components in the same component-type list can share a `name` but differ by `subtype` (e.g., an `encoding` extension and a `storage` extension both named `"zipkin"`). They are logically distinct components.

## Example Scenario

Two extension components named `"zipkin"` exist:
- `{ "name": "zipkin", "subtype": "encoding" }`
- `{ "name": "zipkin", "subtype": "storage" }`

If the `encoding` one is removed in a new version, `_build_component_set` won't detect the removal because both share the same `name` key. The set sees `"zipkin"` in both old and new — so no deprecation is triggered.

## Impact

- A component that moves from one subtype group to another would be **missed** by the deprecation detector — the removal would be invisible.
- Conversely, a subtype group restructure could trigger **false deprecations**.
- Current tests only cover `"subtype": None`, so this gap is not caught.

## Proposed Fix

Change the set key in `_build_component_set` to be `(name, subtype)`:

```python
return {(component["name"], component.get("subtype")) for component in components}
```

And update `removed_names` to `removed_keys`, comparing the full tuple.

Also add a test case with subtypes to the existing `test_deprecation_detector.py`.

## Scope

`deprecation_detector.py` + `test_deprecation_detector.py` — a focused, self-contained fix.

Rama542:
Fixes #493

## What was wrong

The `_build_component_set` method in `DeprecationDetector` was building a plain set of component names to figure out which components had been removed between versions. That works fine when every component in a list has a unique name, but collector extension components can have a `subtype` field (encoding, observer, storage) because they live in nested directories. Two extension components with the same name but different subtypes are completely separate things, yet the old code treated them as the same.

A concrete example: if you had a `zipkin` encoding extension and a `zipkin` storage extension in the previous version, and then the encoding one got removed in the next version, the detector would see `zipkin` in both the old and new sets and conclude nothing was removed. The deprecation would go untracked.

## What I changed

Changed the set to hold `(name, subtype)` tuples instead of plain names. Updated the check inside `detect_deprecated` to match against those tuples. The variable was also renamed from `removed_names` to `removed_keys` to reflect that it now holds tuples rather than plain strings.

## Tests

Added two new test cases in `test_deprecation_detector.py`:

- `test_detect_deprecated_subtype_removal`: verifies that removing one subtype entry while keeping another with the same name is correctly detected as a deprecation.
- `test_detect_deprecated_same_name_different_subtype_no_false_positive`: verifies that when a component with a specific subtype survives unchanged, no false deprecation is reported.

All 8 tests pass.

### DIFF
--- a/ecosystem-automation/collector-watcher/src/collector_watcher/deprecation_detector.py
+++ b/ecosystem-automation/collector-watcher/src/collector_watcher/deprecation_detector.py
@@ -52,11 +52,11 @@ class DeprecationDetector:
             previous_set = self._build_component_set(previous_components.get(component_type, []))
             current_set = self._build_component_set(current_components.get(component_type, []))
 
-            removed_names = previous_set - current_set
+            removed_keys = previous_set - current_set
 
             deprecated_list = []
             for component in previous_components.get(component_type, []):
-                if component["name"] in removed_names:
+                if (component["name"], component.get("subtype")) in removed_keys:
                     deprecated_component = self._create_deprecated_component(
                         component, previous_version, current_version
                     )
@@ -67,9 +67,9 @@ class DeprecationDetector:
         return deprecated_components
 
     @staticmethod
-    def _build_component_set(components: list[dict[str, Any]]) -> set[str]:
-        """Build a set of component names."""
-        return {component["name"] for component in components}
+    def _build_component_set(components: list[dict[str, Any]]) -> set[tuple[str, str | None]]:
+        """Build a set of (name, subtype) tuples to uniquely identify components."""
+        return {(component["name"], component.get("subtype")) for component in components}
 
     @staticmethod
     def _create_deprecated_component(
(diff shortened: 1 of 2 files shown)


ecosystem-automation/v1-registry-sync/README.md
```diff
@@ -0,0 +1,47 @@
+# V1 Registry Sync
```
jaydeluca, line 1: i think this file is unrelated to this PR?

jaydeluca (changes requested):
most of these changes seem unrelated to the PR

ecosystem-automation/v1-registry-sync/README.md
Rama542: Hi @jaydeluca, you are right, those extra commits got pulled in from an earlier branch I was working on and they have nothing to do with this fix. I have rebased the branch so it now only contains the two files that were actually changed for this issue: `deprecation_detector.py` and `test_deprecation_detector.py`. All 8 tests pass locally. Sorry for the noise, please take another look when you get a chance.

jaydeluca (approved):

PR #508, opened 2026-05-18
feat: Add Expand all / Collapse all controls to Configuration Builder

### LINKED ISSUE #476: Add `Expand all` / `Collapse all` controls to the Configuration Builder
The SDK and Instrumentation tabs in `src/features/java-agent/configuration/configuration-builder-page.tsx` render many collapsible cards. Today the user has to click each one individually to expand or collapse it.

Follow-up to #459.

### What to change

In `src/features/java-agent/configuration/configuration-builder-page.tsx` (or a small new toolbar component):

- Add two buttons (`Expand all`, `Collapse all`) visible in both tab bodies.
- Wire them to flip the expansion state of every section card.
- The expansion state lives in the individual `group-renderer` / `general-section-card` components today; expose a way to drive them externally (context, props, or a shared store).

### Why

Faster navigation on dense pages with many configurable sections, and easier "fully open" inspection when scanning the output.

### Tests

Integration test: click `Expand all`, assert all relevant cards have `aria-expanded="true"`; click `Collapse all`, assert the opposite.

MeloveGupta:
## What

Added "Expand all" and "Collapse all" buttons to both the SDK and Instrumentation tabs of the Configuration Builder. Previously users had to click each section individually to expand or collapse it.

---

## Changes

**`src/features/java-agent/configuration/components/section-expansion-context.tsx`** (new file)
- New `SectionExpansionProvider` context that exposes `expandAll()` and `collapseAll()` actions and broadcasts a signal with a nonce so subscriber components re-render on every call, even repeated ones of the same action

**`src/features/java-agent/configuration/components/group-renderer.tsx`**
- Top-level group sections now subscribe to the expansion signal via `useSectionExpansion()`
- On expand: sections that are enabled open; disabled sections are left collapsed
- On collapse: all sections collapse regardless of enabled state

**`src/features/java-agent/configuration/components/general-section-card.tsx`**
- Lifted expansion state from `defaultExpanded` prop to controlled `useState`
- Subscribes to the expansion signal to respond to Expand all / Collapse all

**`src/features/java-agent/configuration/configuration-builder-page.tsx`**
- Wrapped both `SdkTabContent` and `InstrumentationTabBody` with `SectionExpansionProvider`
- Added `ExpandCollapseToolbar` component rendering "Expand all | Collapse all" above the section list in both tabs

---

## Before / After

| | Before | After |
|--|--------|-------|
| SDK tab | Each section must be expanded individually | "Expand all" opens all enabled sections at once; "Collapse all" collapses all |
| Instrumentation tab | Same manual per-section interaction | Same Expand all / Collapse all controls |

Expand all
Screenshot from 2026-05-18 17-21-53

Collapse all
Screenshot from 2026-05-18 17-22-03

---

## Verified By

- Clicked "Expand all" on SDK tab - all enabled sections expanded
- Clicked "Collapse all" on SDK tab - all sections collapsed to headers
- Clicked "Expand all" on Instrumentation tab - General settings and all instrumentation cards expanded
- Clicked "Collapse all" on Instrumentation tab - all sections collapsed
- `npx tsc --noEmit` - no TypeScript errors

---

## Type of Change

- [x] Feature

Closes #476


MeloveGupta:
Hey @lucacavenaghi97 , just pushed a PR!

The initial implementation used `useEffect` to sync expansion state from the signal, which triggered the `react-hooks/set-state-in-effect` lint error in CI. I've reworked it to derive the expanded state during render using `useMemo` instead, the signal's action are used to compute `resolvedExpanded` directly, which gets passed as the controlled `open` prop to `FieldSection`. This avoids the cascading render concern entirely.

Both "Expand all" and "Collapse all" still work correctly on the SDK and Instrumentation tabs, Let me know if you'd like any changes!

MeloveGupta:
Hey @lucacavenaghi97 , I have pushed another update addressing the Copilot feedback!

The previous `useMemo` approach had a real bug, after clicking "Expand all" or "Collapse all", `resolvedExpanded` would permanently override the local `expanded` state, so individual section toggles would stop working. I've fixed this by storing the last seen signal tick in state (`lastNonce`). During render, if the signal has a new tick, we sync the local `expanded` state to match the bulk action and record the tick, subsequent renders see the same tick and leave `expanded` alone, so individual toggles work normally again.

Regarding the integration test suggestion, happy to add that as a follow-up, or I can include it in this PR if you'd prefer it here. Just let me know!

Also, I'd love to have your review on this...

MeloveGupta:
Hey @lucacavenaghi97 , I have pushed another update!

The previous approach of deriving `resolvedExpanded` from the signal during render had two problems that Copilot correctly flagged, calling setState during render is a React anti-pattern that can cause issues in Strict Mode, and it also permanently overrode local expansion state so individual section toggles stopped working after a bulk action.

I've rearchitected the context to own an `overrides` map. When "Expand all" or "Collapse all" is clicked, the map is cleared and `bulkAction` is set. Each section derives its open state from the override map first, then falls back to `bulkAction`, then to its own local state. When a section is individually toggled, it writes its new state into the override map, so the bulk action stays in effect for all other sections while that one section is independently controlled.

I'd love you you had look on it...

lucacavenaghi97:
Honestly testing this locally I don't find it very useful in the current shape. On the SDK tab "Expand all" still leaves all the sub-sections (Schedule Delay, Export Timeout, Sampler...) closed, so you still end up clicking through them one by one. On the Instrumentation tab it only opens the General settings card and leaves the 179 module rows untouched, which is probably where users would expect the biggest payoff.

I think we should iterate on this and treat the two tabs separately, since the right behavior is pretty different:

- SDK tab: expanding only the top-level cards is half the job. Either we open recursively down to the leaf fields, or we keep the bulk control scoped per-card so you can fully open one section at a time without flooding the page.
- Instrumentation tab: the toolbar should drive the module list as well, otherwise the buttons feel broken on the tab where they would matter most.

@jaydeluca what do you think?

@MeloveGupta the original issue was too narrow on my side, let's wait for Jay's input and then realign.

MeloveGupta:
Thanks for the feedback @lucacavenaghi97 ! 
That makes sense, I can see how the current behavior feels incomplete, especially on the Instrumentation tab where the 179 module rows are untouched.

Happy to wait for @jaydeluca for the  input before iterating. Once we align on the right direction for each tab I'll pick it back up...

jaydeluca:
i agree with @lucacavenaghi97's idea of recursively opening the leafs, and that for the instrumentation tab we should be opening all the instrumentation modules as well

MeloveGupta:
Hey @lucacavenaghi97 and @jaydeluca , I just pushed an update addressing both your feedback!

Here's what I changed:

**SDK tab** - "Expand all" now recursively opens all nested sections down to the leaf fields (Schedule Delay, Export Timeout, Sampler, Readers, Detectors, etc.), not just the top-level cards. "Collapse all" collapses everything recursively too. Individual section toggles still work independently after a bulk action.

**Instrumentation tab** - "Expand all" now also drives the 178 module rows in the instrumentation list, opening each row to show its configurable options. "Collapse all" closes them all. Individual row toggles still work independently.

Also resolved a few merge conflicts with upstream changes (`activePreviewKey`, `handleInteraction`, `PruneInstrumentationsForAgentVersion`) along the way.

Let me know if this is closer to what you had in mind, or if you'd like any further adjustments!

lucacavenaghi97:
Tested on the deploy preview (SDK tab): "Expand all" still doesn't open the leaf fields. It only expands the top couple of levels, the actual leaves (Schedule Delay, Export Timeout, Exporter, Limits, etc.) stay collapsed. "Collapse all" works fine.

MeloveGupta:
@lucacavenaghi97 , I'll have a look into it, and update the PR shortly...

lucacavenaghi97:
Hi @MeloveGupta ! I noticed this PR has been inactive for about a week. Do you plan to continue working on it in the next few days, or would you prefer that I take over and finish it? I'd be happy to help move it forward and get it merged. Let me know what works best for you. Thanks!

MeloveGupta:
Hey @lucacavenaghi97 , Sorry for the delay, actually for the last few days I have been caught up with my End Sems and some travelling, I do wish to continue on this issue but I am still travelling so I would atleast require a week or so to continue the work and update the PR. But, if the issue is important and is getting held back then I'd be more than happy for you to take it up...

lucacavenaghi97:
No problem at all! Thanks. There's no rush on this one, so take your time and continue whenever you're back and have some bandwidth. Good luck with your exams, and enjoy your travels! 🙂

MeloveGupta:
> Tested on the deploy preview (SDK tab): "Expand all" still doesn't open the leaf fields. It only expands the top couple of levels, the actual leaves (Schedule Delay, Export Timeout, Exporter, Limits, etc.) stay collapsed. "Collapse all" works fine.

Hi @lucacavenaghi97 , what I found out was that the root cause was in two places:

1. `schema-renderer.tsx` - the `wrap()` function for leaf fields (text inputs, toggles, number inputs, etc.) was using `defaultExpanded` only and never reading `bulkAction` from `SectionExpansionContext`. Added a `WrappedLeaf` component that mirrors the same bulk action logic used by `GroupRenderer`.

2. `group-renderer.tsx` - non-top-level groups were using `open={expanded}` instead of `open={resolvedExpanded}`, so nested groups like Limits were also ignoring the bulk action.

Also made `useSectionExpansion` return a safe no-op fallback when used outside a provider, which fixed two unit tests that rendered `SchemaRenderer` in isolation without a `SectionExpansionProvider` wrapper.

Both Expand all and Collapse all are now fully working across all levels - groups, nested groups, and leaf fields. Tested locally across all SDK sections.

Could you review all the changes and tell me if any further changes are required from my side.

lucacavenaghi97:
Thanks a lot @MeloveGupta for the work here! I pushed a small update on top: merged main to clear the conflicts and made Expand/Collapse all reach every nested section on both tabs, plus an integration test. Hope that's ok with you!

lucacavenaghi97 (approved):

jaydeluca (approved):

Issue #477, opened 2026-05-13
Allow collapsing the form panel so the YAML preview can take more screen space


lucacavenaghi97:
The generated YAML in the Configuration Builder is rendered in a 420px-wide preview column. Reading or comparing a longer output in that space can feel cramped.

Follow-up to #459.

### What to change

Add an "Expand preview" button in the preview header that opens the YAML in a near-fullscreen dialog. The dialog should include the same Copy / Download actions and close on ESC or click-outside.

### Why

Reviewing or comparing the generated YAML benefits from more screen space on both axes, without rearranging the working layout.

LoveChauhan-18:
Hi, I would like to work on this issue. Can you please assign it to me!

LoveChauhan-18:
Hi,@lucacavenaghi97, 
I have made an Implementation Plan: Issue #477 — Collapse Form Panel

Goal: Allow users to collapse the middle "Form" panel in the Configuration Builder to give the YAML preview more horizontal screen space.

Technical Approach:
1. State Management: Add a uiState object to the ConfigurationBuilderState (in src/types/configuration-builder.ts) with an isFormCollapsed boolean. This ensures the preference is persisted in localStorage alongside the configuration data.
2. Logic: Implement a TOGGLE_FORM_COLLAPSE action in the configurationBuilderReducer to handle the switching logic.
3. Dynamic Layout: In configuration-builder-page.tsx, update the 3-column grid template.
- Standard: lg:grid-cols-[256px_minmax(0,1fr)_420px]
- Preview Mode: lg:grid-cols-[256px_1fr] (with the middle column hidden).
4. UI Integration: Add a toggle button in the PreviewCard header using Lucide icons (Maximize2 / Minimize2) to allow users to switch modes easily.

Acceptance Criteria:

- The collapse state is preserved across page refreshes.
- The YAML preview expands smoothly to fill the freed space.
- The feature works consistently across both "SDK" and "Instrumentation" tabs.


Please let me know if like the plan!

LoveChauhan-18:
Hi, @lucacavenaghi97 , Please reply if  you like the plan. Should I proceed by making the PR ?

lucacavenaghi97:
Hi @LoveChauhan-18,

[...]

PR #513, opened 2026-05-18
feat(explorer): add Radix Dialog overlay to expand YAML preview (#477)

LoveChauhan-18:
## Description
This pull request implements the "Expand Preview" Dialog overlay (Approach B) in the Java Agent Configuration Builder, as specified in **Issue #477**. 

To maximize legibility and ease of inspection for large auto-generated configurations, users can now open a beautiful, near-fullscreen modal overlay displaying the complete YAML code block. The dialog incorporates full-parity Copy and Download actions, matches the dark glassmorphic themes of the explorer, and adheres to strict CNCF/WAI-ARIA accessibility standards.

### Reusable UI Dialog Primitive
* **Wrapped Radix Dialog** ([dialog.tsx](file:///c:/Users/lovek/OneDrive/Desktop/Ecosystem/opentelemetry-ecosystem-explorer/ecosystem-explorer/src/components/ui/dialog.tsx)): Implemented a reusable, styled wrapper utilizing `@radix-ui/react-dialog`. It features standard backdrop blur, smooth fade/zoom entry and exit transitions, focus trapping, and accessible title/description bindings.
* **Primitive Unit Tests** ([dialog.test.tsx](file:///c:/Users/lovek/OneDrive/Desktop/Ecosystem/opentelemetry-ecosystem-explorer/ecosystem-explorer/src/components/ui/dialog.test.tsx)): Added a robust test suite covering the dialog's lifecycle, trigger events, and backdrop/close dismissals.

### Configuration Builder Integration
* **Dialog Trigger & Layout** ([preview-card.tsx](file:///c:/Users/lovek/OneDrive/Desktop/Ecosystem/opentelemetry-ecosystem-explorer/ecosystem-explorer/src/features/java-agent/configuration/components/preview-card.tsx)): Integrated the Maximize button (`Maximize2` icon) next to the "Reset" action inside the output preview card header.
* **Full-parity Dialog Controls**: Implemented a responsive full-screen dialog layout (`max-w-4xl w-[90vw] max-h-[85vh]`) holding identical Copy and Download controls that cleanly hook into the configuration validation framework (`validateAll()`).

### Enhanced Test Coverage
* **Scoped Integration Tests** ([preview-card.test.tsx](file:///c:/Users/lovek/OneDrive/Desktop/Ecosystem/opentelemetry-ecosystem-explorer/ecosystem-explorer/src/features/java-agent/configuration/components/preview-card.test.tsx)): Extended the preview card unit tests to cover modal rendering, opening trigger actions, and correct clipboard/download callback invocations using stable `@testing-library/react` `within(dialog)` scoping.

### Automated Checks Passed
* **Typechecking**: `bun run typecheck` passed successfully with **0 errors**.
* **Linting & Code Style**: `bun run lint` and `bunx prettier --check` passed with **0 warnings** on all changed files.
* **Copyright Headers**: `python scripts/check_copyright.py` verified that all files have valid copyright headers.
* **Unit & Integration Tests**: Full test suite inside `vitest run` passed successfully: **91/91 test files passed, 881/881 tests passed**.

### Manual Verification
Visual styling, transition animations, and focus containment were thoroughly validated on a local Vite server dev setup at `http://localhost:5173/java-agent/configuration/builder`:
1. Clicked the Maximize button: Dialog launched instantly with a beautiful scale-in animation and backdrop blur.
2. Verified focus trapping: Keyboard navigation remained locked inside the active modal overlay.
3. Tested Copy and Download buttons inside the modal: Verified perfect behavior parity with the inline output preview.
4. Dismissed the dialog: Confirmed focus returned cleanly back to the Maximize trigger button, conforming to WAI-ARIA standards.


ecosystem-explorer/src/components/ui/dialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+  ComponentPropsWithoutRef<typeof RadixDialog.Content> & {
+    showCloseButton?: boolean;
+  }
+>(({ className = "", children, showCloseButton = true, ...props }, ref) => (
+  <DialogPortal>
+    <DialogOverlay />
+    <RadixDialog.Content
+      ref={ref}
+      className={`border-border/50 bg-card/95 text-foreground data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=open]:fade-in-0 data-[state=closed]:fade-out-0 data-[state=open]:zoom-in-95 data-[state=closed]:zoom-out-95 fixed top-[50%] left-[50%] z-50 grid w-full max-w-lg translate-x-[-50%] translate-y-[-50%] gap-4 border p-6 shadow-xl backdrop-blur-md duration-200 sm:rounded-lg ${className}`}
+      {...props}
```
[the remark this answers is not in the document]
lucacavenaghi97: Confirming this and adding measurements after applying the fix locally. With `grid w-full max-w-lg gap-4` removed from the wrapper's defaults (keeping the positioning, background, padding and animations), `preview-card.tsx:104`'s `flex flex-col max-h-[85vh] w-[90vw] max-w-4xl` finally takes effect: on a 1440x900 viewport the dialog goes from 448px to 784px wide, the YAML stops clipping horizontally, and the inner container stops overflowing the dialog bottom by ~58px (which is what produced the visual glitch where the YAML escaped past the rounded corner on long configs). Same fix solves both the layout-precedence problem you flagged and the sizing.

ecosystem-explorer/src/components/ui/dialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+    >
+      {children}
+      {showCloseButton && (
+        <RadixDialog.Close className="ring-offset-background focus:ring-primary data-[state=open]:bg-accent data-[state=open]:text-muted-foreground absolute top-4 right-4 cursor-pointer rounded-sm opacity-70 transition-opacity hover:opacity-100 focus:ring-2 focus:ring-offset-2 focus:outline-none disabled:pointer-events-none">
```
lucacavenaghi97: `data-[state=open]:bg-accent data-[state=open]:text-muted-foreground` doesn't do anything here:

1. The `accent` token isn't defined in `src/themes.ts` (`grep -rn "accent" src/` returns only this line), so `bg-accent` resolves to an unset CSS variable and the button gets no background.
2. The selector `data-[state=open]` never has a counterpart to switch on, because the close button is only rendered while the dialog is open. It's a shadcn pattern that makes sense on toggle triggers (dropdown, popover) but not on a close button.

Drop both classes.

ecosystem-explorer/public/data/javaagent/global-configurations.json
```diff
@@ -41,12 +41,12 @@
...
-    "description": "Enables experimental span attributes `aws.agent`, `aws.lambda.function.arn` and `aws.lambda.function.name` for AWS SDK instrumentation.",
+    "description": "Enables the experimental span attributes `aws.agent`, `aws.queue.name`, `aws.lambda.function.arn` and `aws.lambda.function.name` for AWS SDK instrumentation.",
     "name": "otel.instrumentation.aws-sdk.experimental-span-attributes",
     "type": "boolean",
     "instrumentations": [
-      "aws-sdk-2.2",
-      "aws-sdk-1.11"
+      "aws-sdk-1.11",
+      "aws-sdk-2.2"
     ]
```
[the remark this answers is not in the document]
lucacavenaghi97: Adding context: `bun run serve` auto-runs `scripts/aggregate-configurations.mjs`, which regenerates this file every time. So this diff is most likely just the output of running the dev server locally and got picked up by the commit. Reverting it is the cleanest path.

ecosystem-explorer/src/components/ui/dialog.tsx
```diff
@@ -0,0 +1,97 @@
...
+          <X className="h-4 w-4" />
```
lucacavenaghi97: Touch target: this close button has no padding, so the click area is just the 16x16 icon (measured 14x14 in the rendered DOM at desktop). That's well below WCAG 2.5.8 AA (24x24 minimum).

Fix I tested locally: switch `top-4 right-4` to `top-2 right-2` and add `p-2 inline-flex items-center justify-center` to the existing classes. The icon stays visually anchored (15px from top/right of the dialog, was 16px) and the click target bumps to 28x28.

ecosystem-explorer/src/features/java-agent/configuration/components/preview-card.tsx
```diff
@@ -70,19 +77,64 @@ export function PreviewCard({ schema, javaAgentVersion }: PreviewCardProps): JSX
...
+                <div className="space-y-1">
+                  <DialogTitle className="text-xl font-semibold">
+                    YAML Configuration Preview
+                  </DialogTitle>
+                  <DialogDescription className="text-muted-foreground text-xs">
+                    Complete generated YAML configuration for your OpenTelemetry Java Agent.
+                  </DialogDescription>
+                </div>
+                <div className="flex flex-wrap items-center gap-2">
+                  <CopyButton
```
lucacavenaghi97: This Copy/Download block duplicates lines 60-75 verbatim: same `validateAll` call, same filename template (`otel-config-${state.version}.yaml`), same MIME, same handler. Same goes for the chip-button class string, which now appears a lot in this file.

Suggest extracting two small primitives inside this file:

```tsx
function HeaderActionButton({ icon: Icon, label, ...props }) {
  return (
    <button
      type="button"
      className="border-border/60 bg-card text-foreground hover:bg-card/80 inline-flex cursor-pointer items-center gap-1 rounded-md border px-3 py-1.5 text-xs"
      {...props}
    >
      <Icon className="h-3 w-3" aria-hidden="true" />
      {label}
    </button>
  );
}

function PreviewActions({ yaml, filename, onValidate }) {
  return (
    <>
      <CopyButton text={yaml} onClick={onValidate} className="<same chip class>" />
      <HeaderActionButton
        icon={Download}
        label="Download"
        onClick={() => { onValidate(); downloadText(filename, yaml, "text/yaml"); }}
      />
    </>
  );
}
```

Then both headers render `<PreviewActions yaml={yaml} filename={filename} onValidate={validateAll} />`, and `Add all` / `Reset` become `<HeaderActionButton icon={...} label="..." onClick={...} />`. Bonus: bumping the padding from `py-1` to `py-1.5` in the shared button takes the Copy/Download height from 23px to ~26px and gets the chip-buttons over the WCAG 2.5.8 AA threshold (24x24).

lucacavenaghi97: While at it, I'd add `focus-visible:ring-primary focus-visible:ring-2 focus-visible:ring-offset-2 focus-visible:outline-none` to the `HeaderActionButton` classes. The icon-only Expand trigger currently has no visible focus ring on tab (the browser default is `1px auto rgb(16,16,16)`, barely visible on the dark card). All the other chip buttons share the same gap, so adding it in the primitive covers them at once. Pattern used elsewhere in the codebase, e.g. `tabs.tsx:28`.

ecosystem-explorer/src/features/java-agent/configuration/components/preview-card.test.tsx
```diff
@@ -123,12 +123,46 @@ describe("PreviewCard", () => {
...
+    expect(validateAll).toHaveBeenCalledTimes(1);
+
+    downloadSpy.mockClear();
+    validateAll.mockClear();
+    const modalDownloadBtn = within(dialog).getByRole("button", { name: /download/i });
+    expect(modalDownloadBtn).toBeInTheDocument();
+    fireEvent.click(modalDownloadBtn);
+    expect(validateAll).toHaveBeenCalledTimes(1);
+    expect(downloadSpy).toHaveBeenCalledTimes(1);
+  });
```
lucacavenaghi97, line 176: One test missing: content parity between the inline preview and the dialog. The PR's main claim is that the dialog shows the same YAML, but nothing asserts the strings actually match. A `expect(within(dialog).getByText(...).textContent).toBe(inlinePre.textContent)` (or similar query) would pin this down. Without it, a future change that breaks parity (different className causing different rendering, accidentally passing different yaml, etc.) wouldn't be caught.

ecosystem-explorer/src/features/java-agent/configuration/components/preview-card.test.tsx
```diff
@@ -123,12 +123,46 @@ describe("PreviewCard", () => {
 
   it("downloads the YAML with the schema-versioned filename and agent-stamped content", () => {
     render(<PreviewCard schema={schema} javaAgentVersion="2.27.0" />);
-    fireEvent.click(screen.getByRole("button", { name: /download/i }));
+    fireEvent.click(screen.getAllByRole("button", { name: /download/i })[0]);
```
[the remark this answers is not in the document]
lucacavenaghi97: Agreed. Scoping with `within(screen.getByLabelText("Output Preview"))` is the cleaner path, especially once both headers render the same primitive (other comment).

lucacavenaghi97:
That's my review for now. Thanks for the work, take a look at the comments above whenever you have a moment

LoveChauhan-18:
Hi @lucacavenaghi97, thanks for the detailed review! I've addressed all the feedback in the latest commit. Here's a summary of what was fixed:

dialog.tsx
- Removed the default grid w-full max-w-lg gap-4 classes from DialogContent so consumer-supplied layout classes (flex flex-col max-h-[85vh] w-[90vw] max-w-4xl) now take effect without fighting a hardcoded grid. This resolves both the layout-precedence issue and the YAML overflow past the rounded corner on long configs.
- Dropped the dead data-[state=open]:bg-accent data-[state=open]:text-muted-foreground classes from the close button (the accent token isn't defined in our theme).
- Fixed the close button touch target: moved to top-2 right-2 and added p-2 inline-flex items-center justify-center, bringing the click area to 28×28px (WCAG 2.5.8 AA compliant).

preview-card.tsx
- Extracted HeaderActionButton and PreviewActions primitives — both the inline card header and the dialog header now render <PreviewActions />, eliminating the verbatim Copy/Download duplication.
- "Add all" and "Reset" also use <HeaderActionButton /> now.
- Bumped padding to py-1.5 across all chip-buttons (>24px height, WCAG 2.5.8 AA).
- Added focus-visible:ring-primary focus-visible:ring-2 focus-visible:ring-offset-2 focus-visible:outline-none to all interactive buttons, including the expand trigger.

preview-card.test.tsx
- Scoped the Copy and Download button queries to within(screen.getByLabelText("Output Preview")) — no more fragile getAllByRole(...)[0] index selectors.
- Added a content parity assertion: expect(dialogPre?.textContent).toBe(inlinePre?.textContent) to pin down that the dialog shows the same YAML as the inline preview.

global-configurations.json
- Reverted the auto-generated dev-server churn back to the main baseline.

All tests (881/881), lint, and typecheck passes  locally. Let me know if there's anything else to address!

lucacavenaghi97:
Thanks for the work on all the comments, this is really clean now. Agreed with Copilot's two latest comments. Three small things if you're in there:
- `showCloseButton` on `DialogContent` is declared but no caller uses it
- `max-h-[85vh]` → `max-h-[85dvh]` to avoid the address-bar jump on iOS Safari
- `describe("Dialog Wrapper")` → `describe("Dialog")` to match the other wrappers (`tabs.test.tsx`, `segmented-tabs.test.tsx`)

LoveChauhan-18:
Hi @lucacavenaghi97, addressed Copilot's two latest comments in the latest commit:

preview-card.tsx — HeaderActionButton props spread order Fixed: {...props} now spreads before type="button" and className, so type="button" is always enforced regardless of what the caller passes, and any caller-supplied className is merged into the base class string via a template literal rather than silently overriding it.

dialog.test.tsx — flaky close assertion Fixed: the close test is now async and the post-close assertion is wrapped in await waitFor(...) to safely handle Radix Presence's async unmount, preventing potential flakiness across timing/animation changes.

Also addressed your three additional points:

- Removed the unused showCloseButton prop from DialogContent
- max-h-[85vh] → max-h-[85dvh] for iOS Safari
- describe("Dialog Wrapper") → describe("Dialog") to match the other wrappers

All 12 targeted tests pass. Let me know if there's anything else to address!

lucacavenaghi97 (approved):

jaydeluca (approved):

PR #514, opened 2026-05-18
fix: show accurate error message when versions list fails to load in ConfigurationBuilderPage

### LINKED ISSUE #512: ConfigurationBuilderPage shows a misleading schema error when the versions list fails to load
## Summary

When the versions index fetch fails in ConfigurationBuilderPage, the page shows a generic schema error message instead of accurately telling the user that loading the available versions failed. This makes it very hard for users to understand what actually went wrong.

## The Problem

The page fetches three things on load: the list of available versions, the schema for the selected version, and the starter template. The hook for versions is:

```ts
const versions = useConfigVersions();
```

It returns a standard DataState shape with loading, data, and error fields. The error field is properly set when the network call fails. However, the page never checks versions.error at all.

When the versions fetch fails, versions.data is null, so the version string falls back to an empty string. The schema and starter hooks both short-circuit immediately when version is empty and return loading: false with no error and no data. This means root ends up as null.

The page then hits the condition:

```tsx
schema.error || !root
```

Since root is null, this evaluates to true and the user sees:

```
Failed to load schema.
```

That message is wrong. The schema never failed. The versions list did. The user has no way to know that, and the VersionSelector area is also silently empty, which adds to the confusion.

## Steps to Reproduce

1. Open DevTools and throttle the network or block the versions index URL
2. Navigate to the Configuration Builder page
3. Observe that the page shows "Failed to load schema." with no further context

## Expected Behavior

The page should check versions.loading and versions.error separately from the schema and starter states. If the versions fetch is still in flight, show a loading indicator. If it has failed, show a message that specifically says the versions list could not be loaded so the user knows exactly what went wrong.

## Proposed Fix

Add a versions.loading check at the top of the render so the page does not try to proceed with an empty version string. Add a versions.error check that shows a versions-specific error message. This keeps the responsibility of each error message tied to the piece of data it actually represents.

The loading and error states for versions, schema, and starter are all independent and each deserves its own handling rather than collapsing into a single generic message.

## Scope

The change is isolated to ConfigurationBuilderPage and the affected lines are the loading/error block and the VersionSelector guard.

Rama542:
Fixes #512

## What was happening

Every time someone opened the Configuration Builder page, the very first thing the app does is fetch the list of available schema versions. The hook that does this returns a proper loading state and an error state, but the page was never reading either of them.

Because of that, two things went wrong at the same time. First, while versions were still loading, the version string was empty, which caused the schema and starter hooks to return immediately with no data and no error. The page then hit the check for root being null and displayed "Failed to load schema" before any schema had even been requested. Users were seeing an error message on a completely normal page load for a brief moment, or permanently if the network was slow.

Second, if the versions fetch actually failed due to a network error or a bad response, the exact same misleading message appeared. The user had no way of knowing the real problem was with the versions list, not the schema. They would likely try reloading thinking the schema was broken when actually the versions endpoint was unreachable.

## What changed

Added a check for versions.loading and versions.error at the top of the render chain, before any schema or starter state is evaluated. The order of checks is now versions loading, then versions error, then schema and starter loading, then schema and starter errors. Each state gets its own specific message so users always know exactly what the page is waiting on or what failed.

While versions are still in flight, "Loading versions" appears in both the header area where the version selector would show up and in the main content area. This way there is always visible feedback and no gap where an incorrect error message sneaks in.

If versions fail to load, the page now says "Failed to load available versions" which is accurate and gives the user the right signal about what to investigate. The tabs block is only rendered after versions have loaded successfully, which also prevents the schema and starter hooks from being called with an empty version string.

## Testing

TypeScript reports no type errors. All 33 integration tests pass.

### DIFF
--- a/ecosystem-explorer/src/features/java-agent/configuration/configuration-builder-page.tsx
+++ b/ecosystem-explorer/src/features/java-agent/configuration/configuration-builder-page.tsx
@@ -352,42 +352,48 @@ export function ConfigurationBuilderPage() {
             ) : null}
           </div>
         </div>
-        <Tabs value={activeTab} onValueChange={setActiveTab}>
-          <TabsContent value="sdk">
-            {!schemaVersion || schema.loading || starter.loading ? (
-              <p className="text-muted-foreground mt-4 text-sm">Loading schema…</p>
-            ) : schema.error ? (
-              <p className="mt-4 text-sm text-red-400">Failed to load schema.</p>
-            ) : starter.error ? (
-              <p className="mt-4 text-sm text-red-400">Failed to load starter template.</p>
-            ) : root ? (
-              <SdkTabContent
-                schema={root}
-                starter={starter.data}
-                schemaVersion={schemaVersion}
-                javaAgentVersion={javaAgentVersion}
-                activeTab={activeTab}
-              />
-            ) : null}
-          </TabsContent>
-          <TabsContent value="instrumentation">
-            {!schemaVersion || schema.loading || starter.loading ? (
-              <p className="text-muted-foreground mt-4 text-sm">Loading schema…</p>
-            ) : schema.error ? (
-              <p className="mt-4 text-sm text-red-400">Failed to load schema.</p>
-            ) : starter.error ? (
-              <p className="mt-4 text-sm text-red-400">Failed to load starter template.</p>
-            ) : root ? (
-              <InstrumentationTabContent
-                schema={root}
-                starter={starter.data}
-                schemaVersion={schemaVersion}
-                javaAgentVersion={javaAgentVersion}
-                activeTab={activeTab}
-              />
-            ) : null}
-          </TabsContent>
-        </Tabs>
+        {schemaVersionsState.loading ? (
+          <p className="text-muted-foreground mt-4 text-sm">Loading versions…</p>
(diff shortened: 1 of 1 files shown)


lucacavenaghi97:
Hi @Rama542, thanks for looking into this. The symptom from #512 doesn't reproduce on current main: #492 already reworked this block, so blocking the versions endpoint leaves the page on "Loading schema…" without ever showing "Failed to load schema." If you don't have objections I'd close this and update #512 to match.

Rama542:
Hi @lucacavenaghi97, thanks for the context on #492. You are right that the old "Failed to load schema." symptom no longer shows up, but when versions fail to load the page now sits on "Loading schema..." indefinitely, which still leaves users without any useful feedback.

I have rebased the branch cleanly on top of current main and narrowed the change down to just wrapping the Tabs block in a versions loading and error guard. So if the versions endpoint fails, the user now sees "Failed to load available versions." instead of a spinner that never resolves. The conflict is gone and all existing tests still pass.

Happy to adjust the wording or approach if you have a preference, but I think the behavior is still worth fixing even after #492. Let me know what you think.

lucacavenaghi97 (approved):
Thanks for the rebase. Fix looks good

jaydeluca (approved):

PR #522, opened 2026-05-20
fix: use (name, subtype) tuple as deduplication key in add_deprecated_components

### LINKED ISSUE #523: bug: add_deprecated_components silently drops deprecations when two components share a name but have different subtypes
## Summary

`InventoryManager.add_deprecated_components` deduplicates entries by component name alone. When two components share the same name but live under different subtypes, the second removal is silently dropped and never written to the deprecations index.

## Background

The OpenTelemetry Collector has nested component directories for extensions. For example `extension/encoding` and `extension/storage` can both contain a component called `zipkin`. These are separate components that just happen to share a name. The correct way to uniquely identify them is the `(name, subtype)` pair.

## Steps to reproduce

1. A component named `zipkin` under subtype `encoding` gets removed in version 0.140.0. `add_deprecated_components` writes it to the index correctly.
2. In a later release, `zipkin` under subtype `storage` is also removed.
3. `add_deprecated_components` is called again. It builds `existing_names = {"zipkin"}` from the index.
4. When it checks the new `zipkin` (storage) entry, it finds `"zipkin" in existing_names` and skips it.
5. The storage deprecation is never recorded. It is gone from the registry permanently.

## Root cause

In `inventory_manager.py`, the deduplication set is built from names only:

```python
existing_names = {comp["name"] for comp in deprecations[distribution][component_type]}

if component["name"] not in existing_names:
    deprecations[distribution][component_type].append(component)
```

## Expected behavior

Both `zipkin (encoding)` and `zipkin (storage)` should appear as separate entries in the deprecations index because they are separate components.

## Relation to issue 493

Issue 493 identified the same root cause in `DeprecationDetector._build_component_set`, which was fixed in PR 498 by switching from name-only sets to `(name, subtype)` tuples. The same fix needs to be applied to the persistence layer in `add_deprecated_components`.

## Affected file

`ecosystem-automation/collector-watcher/src/collector_watcher/inventory_manager.py`

Rama542:
Closes #523

## What this fixes

The `add_deprecated_components` method in `InventoryManager` was checking for duplicates using only the component name. That works fine when every component name is unique, but it breaks down as soon as two components share the same name under different subtypes.

Here is a concrete example. The Collector has extensions under both `extension/encoding` and `extension/storage`. If a component called `zipkin` existed under both, and the encoding variant was removed first and recorded as deprecated, then when the storage variant was removed later, the code would find `zipkin` already in `existing_names` and skip the second entry entirely. The storage deprecation would be permanently lost from the registry with no warning.

## The fix

Changed the deduplication set from a flat set of names to a set of `(name, subtype)` tuples. This is exactly the same approach used to fix `DeprecationDetector._build_component_set` in PR 498 and issue 493. The detection side was already corrected there, and this PR corrects the persistence side so correctly detected removals are not dropped before they reach the registry.

Before:
```python
existing_names = {comp["name"] for comp in deprecations[distribution][component_type]}

if component["name"] not in existing_names:
```

After:
```python
existing_keys = {
    (comp["name"], comp.get("subtype"))
    for comp in deprecations[distribution][component_type]
}

if (component["name"], component.get("subtype")) not in existing_keys:
```

## Tests

Added two new test cases in `test_inventory.py`:

`test_add_deprecated_components_same_name_different_subtype`
Verifies that when `zipkin (encoding)` is already in the index and `zipkin (storage)` is newly deprecated, both entries end up in the index.

`test_add_deprecated_components_no_duplicate_same_name_and_subtype`
Verifies that an entry with the exact same name and the exact same subtype is still correctly skipped, so the existing deduplication behavior is preserved.

All 145 tests in the collector-watcher suite pass.

### DIFF
--- a/ecosystem-automation/collector-watcher/src/collector_watcher/inventory_manager.py
+++ b/ecosystem-automation/collector-watcher/src/collector_watcher/inventory_manager.py
@@ -353,10 +353,14 @@ class InventoryManager:
             new_deprecated: New deprecated components by type
         """
         for component_type, components in new_deprecated.items():
-            existing_names = {comp["name"] for comp in deprecations[distribution][component_type]}
+            existing_keys = {
+                (comp["name"], comp.get("subtype"))
+                for comp in deprecations[distribution][component_type]
+            }
 
             for component in components:
-                if component["name"] not in existing_names:
+                key = (component["name"], component.get("subtype"))
+                if key not in existing_keys:
                     deprecations[distribution][component_type].append(component)
                     logger.info(
                         f"Added deprecated {component_type}: {component['name']} "
--- a/ecosystem-automation/collector-watcher/tests/test_inventory.py
+++ b/ecosystem-automation/collector-watcher/tests/test_inventory.py
@@ -607,3 +607,110 @@ def test_add_deprecated_components_multiple_distributions(temp_inventory_dir):
     assert deprecations["core"]["receiver"][0]["name"] == "corereceiver"
     assert len(deprecations["contrib"]["exporter"]) == 1
     assert deprecations["contrib"]["exporter"][0]["name"] == "contribexporter"
+
+
+def test_add_deprecated_components_same_name_different_subtype(temp_inventory_dir):
+    """A component removed under a second subtype should be added even when the
+    same name is already present in the index under a different subtype."""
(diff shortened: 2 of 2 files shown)


jaydeluca:
is this a duplicate of #498 ?

Issue #500, opened 2026-05-16
config-builder: extend YAML section highlight to Instrumentation tab


lucacavenaghi97:
Follow-up to #478, which scopes the YAML section highlight to the SDK tab.

On the Instrumentation tab the form cards use synthetic `data-section-key` values (`"general"` and `"instrumentations"`) that don't map to a top-level YAML key. Both correspond to nested paths under `instrumentation/development`. Per-instrumentation rows have no `data-section-key` at all.

Extending the highlight there needs a separate design call. A few directions to evaluate:

- Map both synthetic keys to `instrumentation/development` and accept that both UI cards highlight the same YAML block.
- Add per-row `data-section-key` on `InstrumentationRow` and let the highlight target the individual entry inside `instrumentation/development` (more precise but requires the renderer to detect nested section boundaries).
- Expose `instrumentation/development` as a visible top-level section in the TOC, with General and Instrumentations as nested cards.

djvu2k6:
Hey! I’d like to take a look at this issue and explore the highlighting behavior for the Instrumentation tab if nobody else is already working on it.

Pittu-Sharma:
Hi, @lucacavenaghi97  since I am working on the  #478  issue. I understand this issue too. I would also like to work on this issue also. Can you please aassign it to me !
Soon I will share you the inplementation plan for this issue if you allow me to work on this issue.

lucacavenaghi97:
Hi @djvu2k6, sorry for the late reply. I was holding off because this issue is a follow-up to #478, and I wanted #536 to land on `main` first so we'd have a clean base to build on.

Are you still interested in picking this up? If yes, before jumping into implementation I'd like to align on the approach. Could you share how you'd plan to tackle it so we can pick the direction together?

Will assign you once you confirm.

lucacavenaghi97:
Hi @Pittu-Sharma, thanks for the offer. Since @djvu2k6 commented first I'd like to give them the chance to pick it up.

Pittu-Sharma:
> Hi [@Pittu-Sharma](https://github.com/Pittu-Sharma), thanks for the offer. Since [@djvu2k6](https://github.com/djvu2k6) commented first I'd like to give them the chance to pick it up.

Sure, I will wait for @djvu2k6 reply.

djvu2k6:
Hey @lucacavenaghi97, yes — I’d still be interested in working on this.

[...]

PR #536, opened 2026-05-20
feat: highlight active YAML preview section in Configuration Builder (#478)

### LINKED ISSUE #478: Highlight in the YAML preview the section the user is currently editing
The Configuration Builder splits the form on the left and the YAML preview on the right. When the user edits a deeply-nested field, it's not immediately obvious which lines in the YAML correspond to that change.

Follow-up to #459.

### What to change

- Track the last-interacted section in the builder state (typing in a field, clicking a button, expanding or collapsing a card). The section key is already available through the `data-section-key` attributes on every card.
- In `src/features/java-agent/configuration/components/preview-card.tsx` (or the underlying `YamlCodeBlock`), wrap or annotate the YAML lines per top-level section.
- On state change, highlight the corresponding YAML block (for example a background tint that fades after a couple of seconds, or a persistent left border on the currently-active section).

### Why

Makes it obvious which output the current edit affects, particularly for users new to declarative configuration.

### Tests

Integration test: click a control inside one section, assert the preview applies the active marker (class or data attribute) to the matching YAML block.

Pittu-Sharma:
Closes #478

### What was happening

The YAML preview card in the Configuration Builder rendered all sections as a single static code block with no visual link to the form on the left. When a user interacted with a section card (e.g., "Tracer Provider", "Resource", "Instrumentation General settings"), there was no feedback in the preview showing which YAML block corresponded to what they were editing.

### What changed

- **Event delegation**: Added `onFocusCapture` / `onPointerDown` handlers on the section container in `ConfigurationBuilderPage`. They walk up the DOM with `closest("[data-section-key]")` — piggybacking on the attribute already set by `SectionCardShell` for the scroll-spy — and update `activePreviewKey` state without prop-drilling into every field component.
- **Structured YAML rendering**: Extracted `generateYamlSections` and `structuredToString` from `yaml-generator.ts`. `generateYamlSections` returns a `StructuredYaml` array (one entry per top-level key); `structuredToString` serialises it back to a string. `generateYaml` now delegates to both, keeping a single source of truth.
- **Section highlighting in `YamlCodeBlock`**: When `sections` + `activeKey` props are provided, each section is rendered as its own `<div>`. The active section receives a left border and subtle background tint (`border-l-2 border-primary bg-primary/5`). The legacy `code` string prop is retained for `ConfigurationCard` single-snippet use.
- **Tests**: Unit tests added for `generateYamlSections` and `structuredToString` in `yaml-generator.test.ts`; `YamlCodeBlock` and `PreviewCard` component tests updated for the new prop signatures.

### Verification

- Local UI testing confirms that clicking/focusing any SDK or Instrumentation field smoothly highlights the corresponding YAML block in the preview pane.
- `bun run typecheck`, `bun run lint`, and tests (`bun run test` / `bun run test:integration`) all pass successfully locally.

### DIFF
--- a/ecosystem-explorer/src/features/java-agent/configuration/components/yaml-code-block.tsx
+++ b/ecosystem-explorer/src/features/java-agent/configuration/components/yaml-code-block.tsx
@@ -56,14 +56,7 @@ export function YamlCodeBlock({
     }));
   }, [finalStructured.sections]);
 
-  const isSectionActive = (secKey: string) => {
-    if (activePreviewKey === secKey) return true;
-    if (secKey === "instrumentation/development") {
-      return activePreviewKey === "general" || activePreviewKey === "instrumentations";
-    }
-    return false;
-  };
-
+  const isSectionActive = (secKey: string) => activePreviewKey === secKey;
   const renderLines = (lines: Token[][]) => {
     return lines.map((tokens, i) => (
       <Fragment key={i}>
@@ -84,33 +77,37 @@ export function YamlCodeBlock({
   return (
     <pre className={className}>
       {headerLines.length > 0 && (
-        <div className="opacity-80">
+        <span className="block">
           {renderLines(headerLines)}
           {"\n"}
-        </div>
+        </span>
       )}
 
       {fileFormatLines.length > 0 && (
-        <div>
+        <span className="block">
           {renderLines(fileFormatLines)}
           {"\n"}
-        </div>
+        </span>
       )}
 
       {sectionsWithTokens.map((sec) => {
+        if (sec.key === "legacy") {
+          return <Fragment key={sec.key}>{renderLines(sec.lines)}</Fragment>;
+        }
+
         const isActive = isSectionActive(sec.key);
         return (
-          <div
+          <span
             key={sec.key}
-            data-section-key={sec.key}
-            className={`my-1 rounded-sm border-l-2 py-1 pl-3 transition-all duration-300 ${
+            data-yaml-section={sec.key}
+            className={`my-1 block rounded-sm border-l-2 py-1 pl-3 transition-all duration-300 ${
               isActive
                 ? "bg-primary/5 border-primary shadow-[0_0_12px_rgba(var(--primary-hsl),0.05)]"
                 : "border-transparent bg-transparent"
             }`}
           >
             {renderLines(sec.lines)}
-          </div>
+          </span>
         );
       })}
     </pre>
--- a/ecosystem-explorer/src/features/java-agent/configuration/components/yaml-code-block.tsx
+++ b/ecosystem-explorer/src/features/java-agent/configuration/components/yaml-code-block.tsx
@@ -14,31 +14,105 @@
  * limitations under the License.
  */
 import { Fragment, useMemo, type JSX } from "react";
-import { tokenize } from "@/lib/yaml-highlight";
+import { tokenize, type Token } from "@/lib/yaml-highlight";
+import type { StructuredYaml } from "@/lib/yaml-generator";
 
 interface YamlCodeBlockProps {
-  code: string;
+  structured?: StructuredYaml;
+  // Plain-string fallback kept for ConfigurationCard, which renders individual config snippets.
+  code?: string;
+  activePreviewKey?: string | null;
   className?: string;
 }
 
-export function YamlCodeBlock({ code, className }: YamlCodeBlockProps): JSX.Element {
-  const lines = useMemo(() => tokenize(code), [code]);
+export function YamlCodeBlock({
+  structured,
+  code,
+  activePreviewKey = null,
+  className,
+}: YamlCodeBlockProps): JSX.Element {
+  const finalStructured = useMemo<StructuredYaml>(() => {
+    if (structured) return structured;
+    if (code !== undefined) {
+      return {
+        header: "",
+        fileFormat: "",
+        sections: [{ key: "legacy", content: code }],
+      };
+    }
+    return { header: "", fileFormat: "", sections: [] };
+  }, [structured, code]);
+
+  const headerLines = useMemo(() => tokenize(finalStructured.header), [finalStructured.header]);
+  const fileFormatLines = useMemo(
+    () => tokenize(finalStructured.fileFormat),
+    [finalStructured.fileFormat]
+  );
+
+  const sectionsWithTokens = useMemo(() => {
+    return finalStructured.sections.map((sec) => ({
+      key: sec.key,
+      lines: tokenize(sec.content),
+    }));
+  }, [finalStructured.sections]);
+
+  const isSectionActive = (secKey: string) => {
+    if (activePreviewKey === secKey) return true;
+    if (secKey === "instrumentation/development") {
+      return activePreviewKey === "general" || activePreviewKey === "instrumentations";
+    }
+    return false;
+  };
+
+  const renderLines = (lines: Token[][]) => {
+    return lines.map((tokens, i) => (
+      <Fragment key={i}>
+        {tokens.map((t, j) =>
+          t.kind === "ws" ? (
+            t.text
+          ) : (
+            <span key={j} className={`y-${t.kind}`}>
+              {t.text}
+            </span>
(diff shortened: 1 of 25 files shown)


lucacavenaghi97 (commented):
All five of @Copilot's inline comments here look correct. I've left additional inline comments below on things that came up in live testing.

ecosystem-explorer/src/features/java-agent/configuration/components/yaml-code-block.tsx
```diff
@@ -14,31 +14,105 @@
...
+        const isActive = isSectionActive(sec.key);
+        return (
+          <div
+            key={sec.key}
+            data-section-key={sec.key}
+            className={`my-1 rounded-sm border-l-2 py-1 pl-3 transition-all duration-300 ${
+              isActive
+                ? "bg-primary/5 border-primary shadow-[0_0_12px_rgba(var(--primary-hsl),0.05)]"
+                : "border-transparent bg-transparent"
+            }`}
```
lucacavenaghi97: A few things I'd change on this wrapper:

1. The highlight is too subtle to notice on first glance. I'd align with the brand accent already on this page (the title gradient uses `otel-orange`):
   ```diff
   -border-l-2 border-primary bg-primary/5 shadow-[0_0_12px_rgba(var(--primary-hsl),0.05)]
   +border-l-4 border-otel-orange bg-otel-orange/10
   ```
   I'd also drop the `shadow-[...]` and the `transition-all duration-300` (animates border width too, looks odd on section switch). `transition-colors duration-200` is enough.

2. The highlight clips at the end of long lines. The wrapper is `display: block` and only spans the parent's width, but long YAML lines (e.g. `endpoint: ${OTEL_EXPORTER_OTLP_ENDPOINT...}/v1/logs`) extend past the right edge under `white-space: pre`, so the orange background cuts off mid-line. I'd add `min-w-max` to the wrapper. The `<pre>` already has `overflow-auto`, so horizontal scroll continues to work.

3. I'd rename `data-section-key` to `data-yaml-section` here. The form's `SectionCardShell` already uses `data-section-key`. `closest('[data-section-key]')` works only because the handler is scoped to `sectionsContainerRef`, but different concerns deserve different attribute names.

ecosystem-explorer/src/features/java-agent/configuration/components/yaml-code-block.tsx
```diff
@@ -14,31 +14,105 @@
...
+    }));
+  }, [finalStructured.sections]);
+
+  const isSectionActive = (secKey: string) => {
+    if (activePreviewKey === secKey) return true;
+    if (secKey === "instrumentation/development") {
+      return activePreviewKey === "general" || activePreviewKey === "instrumentations";
+    }
+    return false;
+  };
```
lucacavenaghi97: I'd remove this mapping branch. We agreed in the earlier review to scope this issue to the SDK tab, with #500 tracking the Instrumentation-tab extension. Live testing confirmed two problems with the current mapping: silent no-op when `instrumentation/development` isn't in the YAML yet (default state, the section is auto-enabled only after a meaningful edit), and ambiguous UX once it is (General and a row inside Instrumentations highlight the same block). `isSectionActive` would collapse to `activePreviewKey === secKey`.

ecosystem-explorer/src/features/java-agent/configuration/components/yaml-code-block.tsx
```diff
@@ -14,31 +14,105 @@
...
+    if (structured) return structured;
+    if (code !== undefined) {
+      return {
+        header: "",
+        fileFormat: "",
+        sections: [{ key: "legacy", content: code }],
+      };
+    }
+    return { header: "", fileFormat: "", sections: [] };
+  }, [structured, code]);
```
lucacavenaghi97, line 44: When the structured input only contains the synthetic `legacy` section (the `code`-prop fallback used by `ConfigurationCard` and the expanded preview dialog), I'd render the tokens directly without a section wrapper. Right now the dialog opens `<YamlCodeBlock code={yaml}>`, the fallback wraps the whole YAML in a pseudo-section, and that pseudo-section inherits `pl-3 border-l-2 transparent` from the section styling, producing a stray left indent that looks like a vertical line next to the YAML.

ecosystem-explorer/src/features/java-agent/configuration/components/yaml-code-block.tsx
```diff
@@ -14,31 +14,105 @@
...
-              <span key={j} className={`y-${t.kind}`}>
-                {t.text}
-              </span>
-            )
-          )}
-          {i < lines.length - 1 ? "\n" : ""}
-        </Fragment>
-      ))}
+      {headerLines.length > 0 && (
+        <div className="opacity-80">
```
lucacavenaghi97: I'd drop `opacity-80`. The dimmed header reads worse than what's on main

ecosystem-explorer/src/features/java-agent/configuration/configuration-builder-page.tsx
```diff
@@ -199,6 +220,17 @@ function InstrumentationTabBody({
...
+  const [activePreviewKey, setActivePreviewKey] = useState<string | null>(null);
+
+  const handleInteraction = (e: React.BaseSyntheticEvent) => {
+    const key = (e.target as HTMLElement)
+      .closest("[data-section-key]")
+      ?.getAttribute("data-section-key");
+    if (key) {
+      setActivePreviewKey(key);
+    }
+  };
```
lucacavenaghi97: I'd remove the `activePreviewKey` state and `handleInteraction` for this tab, drop `onFocusCapture`/`onPointerDown` from the container below, and pass `activePreviewKey={null}` to `PreviewCard`. A short comment explaining that highlighting is SDK-only and linking #500 would make the intent explicit.

lucacavenaghi97:
Thanks for picking this up. Happy to do another pass once the points above are in.

Minor: `setActivePreviewKey(key)` in the SDK tab handler runs on every pointer/focus event. A `if (key && key !== activePreviewKey)` guard would avoid the wasted closure work.

Pittu-Sharma:
@lucacavenaghi97 Thanks for the thorough review! I've just pushed a new commit addressing all the feedback from you and Copilot:

**Addressed your feedback:**
- **Wasted closure work**: Added the `if (key && key !== activePreviewKey)` guard in the SDK tab's pointer/focus handler.
- **Section styling**: Renamed `data-section-key` to `data-yaml-section` in `YamlCodeBlock` to properly separate the concerns between the configuration builder form and the YAML preview.
- **Header dimming**: Dropped `opacity-80` from the YAML header rendering.
- **Highlight mapping**: Scoped the `isSectionActive` highlighting strictly to the SDK tab and removed the mapping branch logic. The `PreviewCard` in the Instrumentation tab now explicitly receives `activePreviewKey={null}` until #500 is tackled.
- **Fallback synthetic legacy branch**: The fallback plain-string `code` prop (used in the `ConfigurationCard` and the expanded preview dialog) no longer wraps the YAML in a pseudo-section, which fixes the stray left-indent styling issue.

**Addressed Copilot's findings:**
- **Valid HTML**: `YamlCodeBlock` now renders the active sections using `<span className="block">` instead of `<div>` to remain valid phrasing content inside the `<pre>` tag.
- **Schema**: Removed the unrelated `markdown_hash` property from `javaagent-instrumentation.schema.json`.
- **Test coverage**: Renamed the `YamlCodeBlock` test to better reflect the `toContain()` assertions, added a new unit test for the `data-yaml-section` tailwind styling, and added a full end-to-end integration test in `configuration-builder-card-clicks.integration.test.tsx` that asserts interacting with a section highlights the correct YAML block.

Ready for another look whenever you have the time!

lucacavenaghi97 (approved):
Pushed a polish commit on top: otel-orange highlight, General-card leaf fields highlight by their real YAML key, aria-live announcement, prefers-reduced-motion gate, plus an integration test for the leaf-click flow.

LGTM.

lucacavenaghi97 (approved):

jaydeluca (approved):

PR #548, opened 2026-05-21
feat(config-builder): add Import YAML button to load existing configurations

### LINKED ISSUE #540: Configuration Builder: Add ability to import an existing YAML config file
## Summary

Right now the Configuration Builder is a great tool for creating a new OpenTelemetry Java Agent configuration from scratch. You can fill in values, see the live YAML preview on the right, and download the finished file. However, there is currently no way to go in the other direction: if you already have a YAML configuration file on your machine, you cannot load it into the builder to inspect or edit it visually.

## The Problem

A common real-world workflow looks like this:

A team already has an existing `otel-config.yaml` that was written by hand or copied from a colleague. They want to use the Configuration Builder to review what they have, understand each field, or make changes without touching the raw YAML directly. Right now they have no way to do this. They have to start from scratch in the builder and manually re-enter every value they already have in their file.

Another common case: a user builds a configuration in the builder, downloads the YAML, makes a quick manual edit outside the tool, and later wants to bring that file back into the builder to continue editing. That round-trip is completely broken today.

## What Makes This Interesting

The infrastructure for this feature already exists inside the codebase. The `useConfigurationBuilder` hook already exposes a `loadFromYaml(yaml: string)` function that parses a YAML string, hydrates the builder state, and enables the corresponding sections. It even has unit tests covering the happy path and invalid input cases. The missing piece is purely on the UI side: there is no button, file input, or paste area in the Configuration Builder page that calls this function.

You can see the implementation in `ecosystem-explorer/src/hooks/use-configuration-builder.ts` around line 246, and the type signature at line 75.

## What the Feature Should Do

The builder page should have an import option, something like an "Import YAML" button near the existing Download button in the Output Preview card, or in the page header near the version selectors. When clicked it could either:

Open a file picker so the user can select a `.yaml` file from their machine, or show a text area where the user can paste raw YAML content.

After the user provides the YAML, the builder parses it using the existing `loadFromYaml` function and populates all the fields. Errors in the YAML (bad syntax, unknown fields) should surface as a clear message rather than silently failing.

## Why This Matters

This feature would close a significant gap in the builder experience. As the Configuration Builder grows into the primary way teams manage their OTel Java agent setup, the ability to bring an existing configuration into the visual editor will become increasingly important. It also improves trust: users can validate that the builder correctly represents what their hand-written YAML actually does.

The core logic is already written and tested. This is an opportunity to wire up a genuinely useful UI feature on top of existing foundation work.

Rama542:
Closes #540

## What this does

Adds an Import button to the Output Preview card in the Configuration Builder. Clicking it opens a modal where you can either pick a `.yaml` or `.yml` file from your machine using a file picker, or paste raw YAML directly into a text area. Once you click Load Configuration, the builder parses the YAML and populates all the fields so you can continue editing visually.

Before this change, users who already had an existing OpenTelemetry Java Agent config file had no way to bring it into the builder. They had to start from scratch and re-enter every value by hand. This was especially painful for teams that manage their configs in version control and just wanted to use the builder to inspect or make a small change.

## How it works

The `loadFromYaml` function was already implemented and tested inside `useConfigurationBuilder`. This PR just wires it up to a UI. The new `ImportYamlButton` component renders the trigger button and a native `<dialog>` element that handles the file reading and paste workflow. When the YAML fails to parse, the error message is shown inside the dialog so the user knows what went wrong before any state changes happen.

The file input reads the selected file using `FileReader` and puts the content into the same textarea, so the Load logic only ever needs to deal with one code path.

## Files changed

- `src/features/java-agent/configuration/components/import-yaml-button.tsx` — new component
- `src/features/java-agent/configuration/components/import-yaml-button.test.tsx` — 8 new tests covering open, close, validation, success, and error paths
- `src/features/java-agent/configuration/components/preview-card.tsx` — import and render the new button
- `src/features/java-agent/configuration/components/preview-card.test.tsx` — adds the Import button to the existing render assertion, adds dialog polyfill for jsdom

## Testing

All tests pass locally. Typecheck and lint are clean.

### DIFF
--- a/ecosystem-explorer/src/features/java-agent/configuration/components/import-yaml-button.test.tsx
+++ b/ecosystem-explorer/src/features/java-agent/configuration/components/import-yaml-button.test.tsx
@@ -1,0 +1,140 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { render, screen, fireEvent, waitFor } from "@testing-library/react";
+import { ImportYamlButton } from "./import-yaml-button";
+
+const loadFromYaml = vi.fn();
+
+vi.mock("@/hooks/use-configuration-builder", () => ({
+  useConfigurationBuilder: () => ({
+    loadFromYaml: (...a: unknown[]) => loadFromYaml(...a),
+  }),
+}));
+
+// jsdom does not implement HTMLDialogElement.showModal/close
+HTMLDialogElement.prototype.showModal = vi.fn(function (this: HTMLDialogElement) {
+  this.setAttribute("open", "");
+});
+HTMLDialogElement.prototype.close = vi.fn(function (this: HTMLDialogElement) {
+  this.removeAttribute("open");
+});
+
+describe("ImportYamlButton", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("renders the Import trigger button", () => {
+    render(<ImportYamlButton />);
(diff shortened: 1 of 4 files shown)


jaydeluca:
Thanks for the contribution here. I’m not sure we’re ready to move forward with implementation yet.

A few things stood out while testing this:

We would probably want some kind of validation before import. Right now it’s possible to import arbitrary content, which either silently does nothing or can introduce invalid/unexpected YAML.

I think we would also want the imported fields to automatically update the existing UI inputs so the state stays synchronized.

The UX still feels a bit unclear to me at the moment, especially with the modal placement and overall workflow.

This also feels large enough from a product/UX perspective that we should first get agreement on the issue itself before implementing a solution.

Thanks again for taking the time to work on this and open the PR. For now, I’m going to close it until we reach consensus on the design and expected behavior in the issue discussion.

PR #588, opened 2026-05-25
instrumentation YAML preview highlighting

djvu2k6:
- Added active section highlighting for YAML preview
- Fixed instrumentation section key mapping
- Improved real-time preview interaction feedback
- Ensured proper visual highlight using activePreviewKey


lucacavenaghi97:
Thanks for taking this on! I tested it on the Instrumentation tab though, and it doesn't work yet: the highlight never triggers when you interact with the form.

On #500 we'd landed on nested per-module sub-blocks under `instrumentation/development`, but digging into the actual YAML, that turns out not to be possible, there aren't per-module blocks to highlight.

Here's what's going on. The actions land in two different top-level YAML blocks:
- enabling/disabling a module writes to `distribution` as a plain list of names, so a module is just a line like `- activej_http`, not its own block
- the General settings card and a module's expanded config both write to `instrumentation/development`, where `java` is a free-form key-value map with no per-module structure

So per-module block highlighting isn't really possible: enabling N modules adds N entries to one shared list in one block. The rows are currently keyed to `instrumentation/development.java.<module>`, but the toggle changes the `distribution` block, so the keys never match and nothing highlights, which is why it doesn't work right now.

For now I'd keep it block-level like the SDK tab:
- enable/disable toggle, `distribution`
- expanded module config and General card, `instrumentation/development`

Heads up that with this, configuring a module and configuring General light up the same block, since they share it in the YAML. That's fine for now. True per-line highlighting (just the `- activej_http` line) is doable but a bigger, separate change.

There are also a few `console.log`s left in `yaml-code-block.tsx` to remove.

djvu2k6:
@lucacavenaghi97 Thanks for the detailed review!

You're right — the original implementation tried to map highlights to per-module YAML paths, but the generated YAML structure only exposes shared top-level blocks (`distribution` and `instrumentation/development`), so the keys never aligned correctly.

I've updated the behavior to match the SDK-tab approach:

* enable/disable actions now highlight the `distribution` block
* expanded module configuration and General settings now highlight `instrumentation/development`

I also removed the remaining debug `console.log`s from `yaml-code-block.tsx`.

Everything should now behave consistently with the actual YAML structure.

lucacavenaghi97:
Thanks for the work on this! It wasn't highlighting at all on this tab, so I pushed a couple of fixes on top (Copilot's included) plus tests.

lucacavenaghi97 (approved):

djvu2k6:
Thanks for the help and for pushing the fixes and tests, @lucacavenaghi97 
I appreciate the review and the guidance. I learned quite a bit while working through the highlighting behavior and YAML structure.
Thanks again for taking the time to help get this across the finish line.

jaydeluca (approved):
nice work! thanks @djvu2k6 and thanks @lucacavenaghi97 for the review!

Issue #128, opened 2026-03-07
[Explorer] Paginate or lazy load the Java agent instrumentation list


jaydeluca:
Currently we load and display all instrumentations when loading `java-agent/instrumentation`. We should instead paginate this.

Requirements:
* Lazy load ~20 at a time that updates on scrolling (or make a recommendation on another batch size)
* Filtering and searching should still apply to the full list instrumentations

IlyasFardaouix:
I'm not seeing any Java agent instrumentation code in the provided source files. However, based on the issue description, I'm assuming this is related to the `collector_sync.py` module. 

Could you clarify if this is the correct file, or if the Java agent instrumentation code is located elsewhere?

ThyTran1402:
Hi,

I can take a look and work on this issue.

Thanks a lot.

ThyTran1402:
Hi @jaydeluca,

I was able to find the root cause of this issue. The list page renders the full result set in one shot. And there is no slicing, virtualization, or sentinel-based loading anywhere between `useInstrumentations()` and the JSX. For a recent Java agent version that's hundreds of cards mounted on first paint; that's what's making the page heavy.

The fix is:
Render `visibleCount` groups per section, with an `IntersectionObserver` sentinel at the bottom of each section that bumps `visibleCount` by the page size. Reset `visibleCount` whenever filters or `resolvedVersion` change, so search/filter results always start from the top.

Batch size: 24, not 20: so, 24 = 12 full rows = clean rows at any breakpoint and avoids a half-row at the fold; 20 leaves a ragged row on 2-col layouts. If you'd rather, 20 is fine functionally which just biases toward a half row.

I'll open PR for this. Please let me know if that works.

jaydeluca:
fixed by #593

PR #593, opened 2026-05-26
(feat) paginate and lazy load for java agent instrumentation

ThyTran1402:
**Motivation:**  The `/java-agent/instrumentation/<version>` page rendered every instrumentation card on first paint, producing hundreds of mounted components for recent agent versions. This PR introduces scroll-triggered pagination so only the first 24 groups per section render initially, with subsequent batches loaded as the user scrolls.

**Root cause:** `JavaInstrumentationListPage` mapped the full `libraryGroups` and `customGroups` arrays into the JSX with no slicing, virtualization, or sentinel-based loading. 

Each card is non-trivial, so it mounts a `<Link>`, two gradient/grid background `<div>` s, two badge components, and a markdown-rendered description `(renderWithInlineCode)`. For a current 2.x Java agent the result was hundreds of cards mounting on first paint. The data fetch itself `(useInstrumentations)` was fine; the cost was render, not network.

**Fix:** Render only `visibleCount` groups per section. When an `IntersectionObserver` sentinel at the bottom of each grid enters the viewport, bump `visibleCount` by the page size. Filtering and search continue to operate over the full instrumentation list which pagination only slices what gets rendered.

**Issue resolved:** 
- #128


jaydeluca:
as far as i can tell through the deploy demo, this doesn't work. It only shows me the first batch of isntrumentations, and when i scroll down, nothing happens

ThyTran1402:
Hi @jaydeluca,

image

image

Good catch. I just fixed the issue. Now, I think when you scroll down it'll display all of the instrumentations 😃.

Thank you!

jaydeluca (commented):
can confirm all looks to be working nicely now 👍 

can you address the two copilot comments and then this should be good to go

ThyTran1402:
I addressed 2 copilot comments and fixed some of the format checks

jaydeluca (approved):
LGTM, thanks!

Issue #532, opened 2026-05-20
Synchronize Automation


jaydeluca:
We have a handful of automation scripts that are generating information. Some of the information is checked into git, and some are not (LLM files).

We need to ensure that the various moving parts that generate these files are in sync. For example, we sometimes see things like:

```
error: Your local changes to the following files would be overwritten by checkout:
	ecosystem-explorer/public/data/javaagent/global-configurations.json
	ecosystem-explorer/public/schemas/collector-component.schema.json
	ecosystem-explorer/public/schemas/javaagent-instrumentation.schema.json
```

What we need to do as part of this issue is identify the different automation components that contribute to what is shown on the website. When we generate new databases using the db-builder automation, we need to ensure all relevant files are updated and in sync. Right now we are seeing `ecosystem-explorer/public/data/javaagent/global-configurations.json` be out of sync when trying to update the java-instrumentation data.

ecosystem-explorer/public/data/javaagent/global-configurations.json is generated by ecosystem-explorer/scripts/aggregate-configurations.mjs, but it probably makes sense to rewrite it in python to live within the explorer-db-builder automation. The schema files are javascript specific so they are probably better suited to remain written in javascript, but we need to ensure there is proper automation to keep them up to date.

Pittu-Sharma:
Hi, I would like to work on this iisue. Can you please assign it to me!

jaydeluca:
@Pittu-Sharma this is already assigned to someone else

Pittu-Sharma:
> [@Pittu-Sharma](https://github.com/Pittu-Sharma) this is already assigned to someone else

Sure, I will look for another issue.

PR #602, opened 2026-05-27
fix: add .prettierignore path in ecosystem-explorer/package.json

### LINKED ISSUE #586: bug: .prettierignore rule not applied when running prettier from subdirectory
## Problem

Currently, the `.prettierignore` at the repo root has:

```
# Data (avoid formatting large generated files)
ecosystem-explorer/public/data
```

This works correctly when running `prettier --check .` from the repo root. However, `ecosystem-explorer/package.json` has its own `format:check` script that also runs `prettier --check .`, but from within the `ecosystem-explorer/` subdirectory. In that context, the ignore pattern `ecosystem-explorer/public/data` doesn't match `public/data/...`, so those files get picked up.

This is what caused the accidental formatting of `public/data` files in #496 (and many other PRs).

## Possible fixes

**1. Remove `format`/`format:check` from `ecosystem-explorer/package.json`**
The root already covers everything with `prettier --check .`. Having a duplicate script in the sub-package running from a different CWD is the root cause.

**2. Use `--ignore-path` in the sub-package script**
Change `ecosystem-explorer/package.json` from: `"format:check": "prettier --check ."` to: `"format:check": "prettier --check . --ignore-path ../.prettierignore"`. Forces prettier to always use the root ignore file regardless of CWD.

**3. Add a local `.prettierignore` inside `ecosystem-explorer/`**
Add `public/data` to a new `ecosystem-explorer/.prettierignore` so the rule works regardless of where prettier is invoked from.

/cc @open-telemetry/ecosystem-explorer-approvers thoughts? Personally, I like the 2nd option.

omthorat2004:
fixes #586

# Summary

This PR updates the Prettier scripts in `ecosystem-explorer/package.json` to explicitly use the root `.prettierignore` file when commands are executed from within the `ecosystem-explorer/` directory.

# Changes

## package.json

Updated:

```json
"format:check": "prettier --check ."
```

to:

```json
"format:check": "prettier --check . --ignore-path ../.prettierignore"
```

and:

```json
"format": "prettier --write ."
```

to:

```json
"format": "prettier --write . --ignore-path ../.prettierignore"
```

This ensures that Prettier correctly respects the root `.prettierignore` configuration, including ignoring the `public/data` directory, even when commands are run from inside the `ecosystem-explorer/` subdirectory.


omthorat2004:
@jaydeluca I also noticed that running bun run build regenerates files inside public/schemas/ (for example collector-component.schema.json and javaagent-instrumentation.schema.json).

Someone can run bun run build when taking screenshots. I have attached the image where we can see public/schemas added in unstated changes.
Do contributors need to add these files every time contribution ? If not we can add them in .gitignore.


1. public/schemas folder added in unstated changed
Screenshot 2026-05-27 at 7 28 53 AM

2. git diff output
Screenshot 2026-05-27 at 7 29 37 AM


Note: This is not the part of this issue but I noticed while solving current issue.

omthorat2004:
@vitorvasc I have solved the bug rated to .prettierignore.

omthorat2004:
I just rebased the code.

@jaydeluca Could you please merge it whenever you get time so we can avoid repeated rebases/merge conflict resolutions as new changes come in?

jaydeluca:
> @jaydeluca I also noticed that running bun run build regenerates files inside public/schemas/ (for example collector-component.schema.json and javaagent-instrumentation.schema.json).

Should be resolved by #532

jaydeluca (approved):

omthorat2004:
> > @jaydeluca I also noticed that running bun run build regenerates files inside public/schemas/ (for example collector-component.schema.json and javaagent-instrumentation.schema.json).
> 
> Should be resolved by #532

That's good then.

PR #606, opened 2026-05-28
feat(config-builder): add Import YAML dialog for loading existing configurations

### LINKED ISSUE #540: Configuration Builder: Add ability to import an existing YAML config file

Rama542:
## What this PR does

Adds an **Import** button to the configuration builder's preview panel header. Clicking it opens a dialog where users can paste an existing `otel-config.yaml` or upload a file, and the form fills in automatically with all recognised fields.

Closes #540.

## How it works

- A new `ImportYamlDialog` component handles everything: trigger button, dialog open/close state, textarea, file upload via `FileReader`, and error reporting.
- Before handing off to the existing `loadFromYaml` hook the dialog runs its own validation pass using `js-yaml`:
  - Empty input shows a clear prompt to paste or upload.
  - Syntax errors surface the parser message so the user knows exactly what went wrong.
  - YAML that parses to a list or scalar (not a mapping) is rejected with a plain-language explanation.
- On a successful import the dialog closes and the builder reflects the loaded values immediately.
- Hook rejections (e.g. version mismatches surfaced by the hook in future) are caught and displayed as an inline error without closing the dialog.

## UX

The Import button sits alongside the existing Add all, Reset, and expand controls so it feels like a natural part of the panel rather than a bolt-on.

## Testing

13 new unit tests cover: trigger rendering, dialog open, all validation error cases, successful import, hook rejection, error clearing on edit, state reset on reopen, file upload via `FileReader`, and the file input aria-label. Two additional tests in `preview-card.test.tsx` verify the button renders and the dialog opens from the card.

All existing tests continue to pass (967 unit tests, 45 integration tests).

### DIFF
--- a/ecosystem-explorer/src/features/java-agent/configuration/components/import-yaml-dialog.test.tsx
+++ b/ecosystem-explorer/src/features/java-agent/configuration/components/import-yaml-dialog.test.tsx
@@ -1,0 +1,208 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { render, screen, fireEvent, waitFor } from "@testing-library/react";
+import { ImportYamlDialog } from "./import-yaml-dialog";
+
+const loadFromYaml = vi.fn();
+
+vi.mock("@/hooks/use-configuration-builder", () => ({
+  useConfigurationBuilder: () => ({
+    state: {
+      values: {},
+      enabledSections: {},
+      validationErrors: {},
+      version: "1.0.0",
+      isDirty: false,
+    },
+    loadFromYaml: (...args: unknown[]) => loadFromYaml(...args),
+    setValue: vi.fn(),
+  }),
+}));
+
+function openDialog() {
+  fireEvent.click(screen.getByRole("button", { name: /import/i }));
+}
+
+describe("ImportYamlDialog", () => {
+  beforeEach(() => {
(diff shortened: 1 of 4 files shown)


jaydeluca:
@Rama542 I closed the other PR for a reason. Please wait until we have consensus on the issue before opening again.

Rama542:
sorry about that, i misunderstood the process. i will wait here until there is consensus on the design before opening another PR. happy to help with the discussion or answer any questions about my implementation if that helps move things forward.

PR #587, opened 2026-05-25
feat(v1): GlobalSearch (Phase 2 - Home page PR 2)

vitorvasc:
Adds the cross-ecosystem search input to the v1 home, replacing the placeholder skeleton inside `<CoverBlock>` with a real keyboard-navigable typeahead.

Closes #371. Contributes to #84.

## What's in this PR

- Adds `<GlobalSearch>` under `src/v1/components/home/`: controlled input with ⌘K / Ctrl+K focus shortcut (either modifier on either platform), debounced typeahead (200ms), keyboard-navigable dropdown (↓/↑ cycling, Enter, Escape, `aria-activedescendant`), client-side `<Link>` navigation that honors cmd/ctrl/shift/middle-click for opening in a new tab, and `sessionStorage` last-query persistence
- Three-tier result row: `[ecosystem pill] title [stability pill]` / 2-line-clamped description / `ecosystem · facets · version` meta. Cap of 10 visible results with a `Showing N of M matches` footer. Stale results dim while a new query is in flight instead of blanking to "Searching…"
- Reuses `<StatusPill>` (stability) and `<GlowBadge>` (ecosystem, via a local variant map). Ecosystem mapping: `collector` → `info` (blue), `java-agent` → `accent` (OTel-orange), `page` → `muted`. Zero bespoke pill CSS in this component
- Adds the engine under `src/lib/search/`: a generic matching engine (ranked substring matcher, title > description > keywords) plus an orchestrator that composes per-ecosystem `SearchSource` modules (pages, Java Agent, Collector) from a single `SEARCH_SOURCES` registry.
- Each result carries a generic `facets` descriptor rendered in the meta line: Collector emits the component type (e.g. `receiver`), Java Agent emits `standalone library` for instrumentations that also ship without the agent. Agent-only rows stay unmarked so the meta line doesn't repeat the ecosystem pill
- Adds `src/v1/hooks/use-search.ts` returning `DataState<SearchResult[]>` per the v1 typescript-frontend rules; short-circuits empty queries and preserves prior results across loading transitions for the stale-while-loading visual
- Gates `useSearch` on `isOpen` so a persisted query in sessionStorage doesn't trigger a full index build on every home (or `/_dev/components`) mount. The engine only runs while the dropdown is open
- Adds `src/v1/styles/global-search.css` (`.td-search*` layout, grid row template, 2-line clamp, active-row accent, stale dim, `max-height: min(60vh, 36rem)`); registered in `src/v1/styles/index.css`
[...]

PR #629, opened 2026-06-02
feat(explorer-db-builder): generate javaagent index.json

vitorvasc:
The issue #24 specified a per-ecosystem `index.json` for the javaagent - "a flat list of all components with lightweight metadata for initial page load and client-side filtering" - but the pipeline shipped without it. The collector side has had its `index.json` for a while; this PR brings the javaagent to parity.

Implements the `index.json` requirement from #24.

## What's in this PR

- Adds `make_index_instrumentation` to `instrumentation_transformer.py`, extracting the lightweight fields for the index entry: `name`, `display_name`, `description`, `has_telemetry` (derived from whether telemetry is present), and `has_standalone_library`
- Adds `DatabaseWriter.write_index`, which writes `javaagent/index.json` as `{ "ecosystem": "javaagent", "components": [...] }`, sorted by `name` for deterministic output. Full instrumentation detail stays in the content-addressed component files; the index only carries what the browse/search UI needs up front

## What's not in this PR

- **The generated `index.json` artifact.** This is code only. The file is produced by the `Build Explorer Database` workflow, which is `workflow_dispatch`-only (manual). Once this merges, dispatching that workflow for the javaagent ecosystem will generate `index.json`

## Verification

- https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/630

---


lucacavenaghi97 (approved):

jaydeluca (approved):
nice catch. we'll probably need to make some updates to the frontend to add a new hook and use this instead of loading everything. I think there might be a few other fields that will be needed too, like the semantic conventions, features, and whether it has javaagent support. We can always tack those on when we switch over the implementation so we see exactly whats needed

vitorvasc:
@jaydeluca yep! this was indirectly part of #587. I think the new hook can be added there as we develop the global search, but it might be worth tracking the new fields in a separate PR and addressing those changes to the search component on the current javaagent page as well

PR #635, opened 2026-06-02
fix(explorer): edge function turned 304 into 404 for data files

### LINKED ISSUE #632: Config builder occasionally has issues loading
### Expected behavior

When loading the config builder page, no errors occur

### Actual behavior

Sometimes on first load, an error is displayed:

Image

### Steps to reproduce

This doesn't happen every time, but load the page with a fresh cache/browser (no indexdb), and occasionally will see the error. Refreshing the page fixes it

### Affected page

https://explorer.opentelemetry.io/java-agent/configuration/builder

### Evidence (screenshots, logs)  of the occuring bug

_No response_

lucacavenaghi97:
The agent-negotiation edge function turned the 304 from conditional revalidation into a 404, so the config builder intermittently failed to load versions.

- edge function passes non-200 responses through instead of forcing 404
- fetchWithRetry retries a 404 once with a cache-busting reload
- loadConfigVersions validates cached versions like loadVersions

Fixes #632

### DIFF
--- a/ecosystem-explorer/netlify/edge-functions/__tests__/agent-negotiation.test.ts
+++ b/ecosystem-explorer/netlify/edge-functions/__tests__/agent-negotiation.test.ts
@@ -1,0 +1,66 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


jaydeluca (approved):
nice, thanks for the quick fix!

PR #639, opened 2026-06-03
Add URL persistence for Collector Components page search

### LINKED ISSUE #622: Collector Components page search is not persisted in the URL
## Description

On the Collector Components page (`/collector/components`), the text search input state is stored only in local React state (`useState("")`) and is not synced to URL search params. This means the search query is lost on page refresh and cannot be shared via URL.

This is inconsistent with the Java Agent Instrumentation list page, which correctly persists all filters, including text search, in the URL using the `?search=` query parameter.

The type and distribution filters on the same Collector page are properly persisted via `useSearchParams`, making the search input the only filter that breaks this pattern.

**Affected file:**
`ecosystem-explorer/src/features/collector/collector-components-page.tsx` 

## Steps to Reproduce

1. Navigate to `/collector/components`
2. Type a search term (e.g. `Docker`) in the Search input field
3. Observe the URL in the address bar: it does not update with a `?search=Docker` parameter
4. Refresh the page
5. Observe that the search input is now empty and all components are shown again

## Expected Behavior

* The search query should be reflected in the URL as a search query parameter (e.g. `/collector/components?search=Docker`)
* Refreshing the page should restore the search term from the URL
* Sharing the URL with the search parameter should reproduce the same filtered view
* Browser back/forward navigation should restore previous search states
* This should behave identically to the Java Agent Instrumentation list page

## Actual Behavior

* The URL does not change when typing in the search input
* Refreshing the page resets the search to empty
* The search term cannot be shared via URL
* Browser back/forward navigation does not affect the search state

## Screenshots
1.Java Agent

Image

2.Collector

Image

yugeshweb:
Closes #622 

## What changed

Synced the search text input on the Collector Components page (`/collector/components`) to URL search params (`?search=...`).

Previously, the search query was stored only in local React state via `useState("")`. It now reads from and writes to `searchParams`, matching the pattern already used by the Java Agent instrumentation list page.

## Why

The search input was the only filter on the Collector Components page not persisted in the URL. The **Type** and **Distribution** dropdowns already used `searchParams`.

This meant:

- Search was lost on page refresh
- Filtered views couldn't be shared via URL
- Browser back/forward navigation didn't restore search state
- UX was inconsistent with the Java Agent instrumentation list page, which already persists all filters in the URL

## How it was tested

TypeScript compilation passes with zero errors:

  ```bash
  tsc -b --noEmit
  ```

Manual testing:
- Typed a search query
- Verified the URL updates to `?search=<query>`
- Refreshed the page
- Confirmed the search text and filtered results are preserved
- Used browser back/forward navigation
- Confirmed previous search states are restored

## Is this a breaking change?

No.

Previously, the `?search=` parameter was ignored. It is now read and applied. Existing URLs without `?search=` behave exactly as before.

## Screenshot
ss

- A local `searchQuery` state is intentionally retained alongside `searchParams` to avoid input flicker. Typing updates local state immediately while URL updates use `replace: true` to avoid polluting browser history.

- This follows the same derived-state-during-render pattern used in `java-instrumentation-list-page.tsx`

- The `syncedUrlSearch` guard handles external URL changes (such as browser back/forward navigation) by syncing them back into local state, matching the existing Java Agent page behavior.

## Checklist 

### DIFF
--- a/ecosystem-explorer/src/features/collector/collector-components-page.tsx
+++ b/ecosystem-explorer/src/features/collector/collector-components-page.tsx
@@ -88,7 +88,14 @@ function CollectorComponentsContent({ urlVersion }: { urlVersion?: string }) {
   const [searchParams, setSearchParams] = useSearchParams();
   const typeQuery = searchParams.get("type");
   const distributionQuery = searchParams.get("distribution");
-  const [searchQuery, setSearchQuery] = useState("");
+  const urlSearch = searchParams.get("search") ?? "";
+  const [searchQuery, setSearchQuery] = useState(urlSearch);
+
+  const [syncedUrlSearch, setSyncedUrlSearch] = useState(urlSearch);
+  if (syncedUrlSearch !== urlSearch) {
+    setSyncedUrlSearch(urlSearch);
+    setSearchQuery(urlSearch);
+  }
 
   const typeFilter = useMemo(() => getTypeFilter(typeQuery), [typeQuery]);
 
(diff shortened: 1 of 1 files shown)


jaydeluca (approved):
thanks @yugeshweb !

PR #645, opened 2026-06-04
Use indexes for list pages for better performance

jaydeluca:
## What changed

Leverages pre-computed indexes for the list pages instead of fanning out and loading all components individually.

The previous page loads were multiple seconds

## Why

Right now the java agent instrumentation list page takes multiple seconds to load

If i load it fresh right now, it takes over 13 seconds (even sometimes 30+)

image

## How it was tested

Build db and ran tests locally, seeing a significant improvement: (500ms)

image

## Is this a breaking change?

No, theres a fallback to allow this to continue working before we rebuild the db


lucacavenaghi97:
Nice improvement!

One thing I noticed while testing: `loadAllInstrumentations` is also used by the Configuration Builder page, and the slim bundle entries drop two fields it needs:

- `configurations`: with bundle data every module shows "No configurable options for this module"
- `disabled_by_default`: the default-disabled modules (e.g. `dropwizard_metrics`) show as "enabled by default", and the initial customization toggle gets inverted in the YAML

CI doesn't catch it because the committed data has no `bundle_hash` yet, so tests and the preview go through the fallback, which still returns full objects. It would surface after the first db rebuild post-merge.

I tried adding the two fields back in `make_list_instrumentation`: the bundle goes from ~10KB to ~23KB gzipped, so still cheap.

maryliag (approved):

PR #648, opened 2026-06-04
fix: Added validation of fresh API responses to prevent malformed data caching and crashes

### LINKED ISSUE #647: Type Safety: Add validation of fresh API responses to prevent malformed data caching in ecosystem explorer
## Summary
The OpenTelemetry Ecosystem Explorer has a type safety issue where fresh API responses are not validated before being cached, allowing malformed data (like empty objects) to be cached and served to consumers, potentially causing runtime crashes.

## Problem
The ```fetchWithCache()``` function in ```ecosystem-explorer/src/lib/api/fetch-with-cache.ts``` validates cached/stale data when a validator is provided, but it does not validate fresh responses before caching and returning them. This means malformed responses can be cached and served to consumers who expect specific data structures.

Specifically:

```collector-data.ts``` does not pass validators to ```loadVersions()```
```configuration-data.ts``` does not pass validators to ```loadConfigSchema()``` and ```loadConfigStarter()```
```javaagent-data.ts``` does not pass validators to ```loadGlobalConfigurations()```

## Steps to Reproduce
Serve an empty JSON object ```{}``` for ```/data/collector/versions-index.json```
Navigate to the Collector explorer page
The application will crash when trying to access ```.versions``` property on the empty object
This happens because the empty object gets cached and served without validation

## Proposed Fix
Update ```fetchWithCache()``` function to validate fresh responses before caching them
Add validators to all data loading functions in:
```collector-data.ts```: Add validator to ```loadVersions()```
```configuration-data.ts```: Add validators to ```loadConfigSchema()``` and ```loadConfigStarter()```
```javaagent-data.ts```: Add validator to ```loadGlobalConfigurations()```
The fix ensures that only properly structured data is cached and served to consumers, preventing runtime crashes due to malformed data.

### PR DESCRIPTION
**Pull request by MeloveGupta**
## What changed

Added validation of fresh API responses in the ```fetchWithCache``` function and all data loading functions to prevent malformed data (like empty objects) from being cached and served to consumers.

## Why

Fixes #647 
The OpenTelemetry Ecosystem Explorer was vulnerable to crashes when upstream APIs returned malformed data (such as empty JSON objects ```{}```) because the system only validated cached/stale data but not fresh responses. This could have led to consumers assuming expected data structures existed when they didn't, which could cause runtime errors.

## How it was tested

All data loading functions now include appropriate validators for their expected data structures.

## Is this a breaking change?

No

The fix ensures that fresh API responses are validated before being cached, preventing malformed data from being served to consumers.

## Checklist

### CODE REVIEW

File: ecosystem-explorer/src/lib/api/configuration-data.ts [hunk 1]
```diff
@@ -45,6 +49,9 @@ export async function loadConfigStarter(version: string): Promise<ConfigStarter
     `config-starter-${version}`,
     `${BASE_PATH}/defaults/sdk-configuration-defaults-${version}.json`,
     STORES.CONFIGURATION,
-    { allow404: true }
+    {
+      allow404: true,
+      validate: (d) => d === null || (d !== null && typeof d === "object" && !Array.isArray(d)),
```
jaydeluca: the second `d !== null` is always true

```suggestion
      validate: (d) => d === null || (typeof d === "object" && !Array.isArray(d)),
```

ecosystem-explorer/src/lib/api/fetch-with-cache.ts
```diff
@@ -170,6 +170,25 @@ export async function fetchWithCache<T>(
 
       const format = options?.format ?? "json";
       const data = format === "json" ? await response.json() : await response.text();
+
+      // Validate fresh data unconditionally after parsing, before any caching
+      if (options?.validate) {
+        if (!options.validate(data)) {
+          // More resilient approach: try to serve valid stale cache before throwing
```
jaydeluca:
```suggestion
```

ecosystem-explorer/src/lib/api/fetch-with-cache.ts
```diff
@@ -170,6 +170,25 @@ export async function fetchWithCache<T>(
...
+          if (isIDBAvailable()) {
+            const staleData = await getCached<T>(cacheKey, storeType, { allowExpired: true });
+            if (staleData !== null && options.validate(staleData)) {
+              console.warn("Fresh response failed validation, serving stale cache:", cacheKey);
+              return staleData;
+            }
+          }
+          // Only throw if no valid stale cache exists
```
jaydeluca:
```suggestion
```

ecosystem-explorer/src/lib/api/javaagent-data.ts
```diff
@@ -171,7 +173,8 @@ export async function loadGlobalConfigurations(): Promise<GlobalConfiguration[]>
   const data = await fetchWithCache<GlobalConfiguration[]>(
     "global-configurations",
     resolveDataPath(BASE_DIR, "global-configurations.json"),
-    STORES.GLOBAL_CONFIGURATIONS
+    STORES.GLOBAL_CONFIGURATIONS,
+    { validate: (d) => d !== null }
```
jaydeluca: The payload type is `GlobalConfiguration[]`, but the validator only rejects `null`. A malformed `{}` (the empty-object example called out in the PR description) passes `d !== null`, is cached, and is returned to consumers as a fake array — `.length`/`.map` then misbehave downstream.

jaydeluca (commented):
any ideas on how we could add some tests for this behavior?

MeloveGupta:
@jaydeluca I have updated the global configurations test fixture to return the expected array shape, so the strict GlobalConfiguration[] validation stays intact. The other changes you mentioned have also been made. Please let me know if anything else still needs adjustment.

MeloveGupta:
@jaydeluca I looked for the tests, and the best coverage would be around the validation boundary itself: to add tests that verify invalid fresh responses such as an empty object, primitive value, or null are rejected, and that the loader falls back to stale cached data when available.
We can also add a happy-path test to confirm a properly shaped array is accepted and returned as-is, with any generic cache fallback behavior covered if needed. 
According to me this should give us good coverage for both the validation logic and the stale-cache fallback flow.

jaydeluca:
> to add tests that verify invalid fresh responses such as an empty object, primitive value, or null are rejected, and that the loader falls back to stale cached data when available.
We can also add a happy-path test to confirm a properly shaped array is accepted and returned as-is, with any generic cache fallback behavior covered if needed.
According to me this should give us good coverage for both the validation logic and the stale-cache fallback flow.

Cool, if you want to test these ideas out and see if they seem like they provide value that would be great. If you find that they don't that's ok too

jaydeluca (approved):
thanks!

MeloveGupta:
Hi @jaydeluca , I just wanted to apologize for not working on the tests. Actually I got hooked up with some work...
I'd definitely work on the tests and open a PR soon :)

jaydeluca:
hey @MeloveGupta ! if you have something written feel free to submit a PR, otherwise no problem, thanks!

PR #665, opened 2026-06-07
Add Go instrumentation watcher for the ecosystem registry

mikeblum:
## Summary

The **Extract** step of the Go instrumentation watcher ETL pipeline. This branch clones `opentelemetry-go-contrib`, walks its module tree, derives descriptive metadata from each module path, and emits a versioned, content-addressed inventory at `ecosystem-registry/go/contrib/v{version}/instrumentation.yaml`.

Scope is the components that **instrument a developer's code** — instrumentation wrappers (gin, grpc, http…) and bridges (zap, logrus…). Pipeline-config components (exporters, propagators, samplers, detectors, processors) configure the SDK rather than a target library and are out of scope.

The **Transform** step (AST-based telemetry extraction) is stacked on top in `feat/golang-watcher-instrumentation`.

## What this PR does

| Layer | Responsibility |
|---|---|
| \`repo/\` | Clone + fetch \`opentelemetry-go-contrib\`; checkout at release tag or \`main\`; resolve per-module version tags |
| \`instrumentation/\` | Walk go.mod files; parse module path; derive \`Metadata\` (name, display name, type, target, stability…) |
| \`inventory/\` | Write versioned \`instrumentation.yaml\` envelopes; idempotent release guard; SNAPSHOT cleanup |
| \`cmd/watcher/\` | Orchestrates two versions per run: latest release tag + \`main\` SNAPSHOT |

## Watcher contract coverage

| # | Responsibility | Status |
|---|----------------|--------|
| 1 | Version management | ✅ latest release tag + \`main\` SNAPSHOT, per-module versions, \`VersionExists\` idempotency |
| 2 | Schema evolution | ✅ explicit \`file_format: 0.1\`, metadata derived from each module's own \`go.mod\` |
| 3 | Change detection | ⚠️ versioned + content-addressed for diffing; auto GitHub-issue creation deferred |
| 4 | Data regeneration | ✅ \`make sync\` per tag, idempotent releases, SNAPSHOT refreshed each run |
| 5 | Deterministic output | ✅ stable-sorted arrays, no timestamps, SHA-256/12 content hash — byte-stable across runs |

## Test plan

- [x] \`go test ./...\` — full suite across \`conf\`, \`instrumentation\`, \`inventory\`, \`repo\`
- [x] Table-driven tests throughout; \`Enabled()\`-based log level assertions
- [x] Production bug fixed: \`newLog(level)\` was silently overriding the level parameter with the \`"info"\` fallback when \`LOG_LEVEL\` was unset
- [x] e2e \`make sync\` → v1.44.0 + SNAPSHOT, byte-identical across two runs
- [ ] Reviewer validation


mikeblum:
## 📦 Rendered output preview

The inventory files themselves are not in this PR's diff — they're produced by the nightly CI automation (`ecosystem-registry/go/` is generated, like the other watchers). This comment shows what `make sync` actually emits so reviewers can evaluate the output shape without running it.

Generated from a local `make sync` against **opentelemetry-go-contrib**:
- Release: **`v1.44.0`** → `ecosystem-registry/go/contrib/v1.44.0/instrumentation.yaml` (sha256/12 `77c946b0d191`)
- Snapshot: **`v1.44.1-SNAPSHOT`** from `main` (sha256/12 `e2da7f6fd844`)
- **19 libraries**, 948 lines each. Byte-identical across repeated runs (determinism verified).

### Libraries (19)

Scope is components that **instrument a developer's code**: instrumentation wrappers + log bridges. Pipeline-config components (exporters, propagators, samplers, detectors, processors) are intentionally excluded — see PR description.

| display_name | type | module version |
|---|---|---|
| logr | bridge | v0.19.0 |
| logrus | bridge | v0.19.0 |
| prometheus | bridge | v0.69.0 |
| slog | bridge | v0.19.0 |
| zap | bridge | v0.19.0 |
| AWS | wrapper | v0.69.0 |
| Echo | wrapper | v0.69.0 |
| Gin | wrapper | v0.69.0 |
| gRPC | wrapper | v0.69.0 |
| Host | wrapper | v0.69.0 |
| HTTP | wrapper | v0.69.0 |
| HTTP Trace | wrapper | v0.69.0 |
| Lambda | wrapper | v0.69.0 |
| MongoDB (v1) | wrapper | v0.69.0 |
| MongoDB (v2) | wrapper | _(no release tag yet)_ |
| Mux | wrapper | v0.69.0 |
| RESTful | wrapper | v0.69.0 |
| Runtime | wrapper | v0.69.0 |
| X-Ray Config | wrapper | v0.69.0 |

> **Note for reviewers:** some libraries share a display name (e.g. MongoDB v1 and v2 drivers). The `name` field is derived from the full module path to keep records unique. A driver showing an empty `version` has no per-module release tag yet — it picks one up once go-contrib tags it.

### Example record — `otelecho` (spans + metrics fused)

```yaml
- name: instrumentation-github.com-labstack-echo-otelecho
  display_name: Echo
  description: ""
  source_path: instrumentation/github.com/labstack/echo/otelecho
  scope:
    name: go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho
  module:
    path: go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho
    version: v0.69.0
  target_module: github.com/labstack/echo
  go_min_version: 1.25.0
  library_link: https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/github.com/labstack/echo/otelecho
  instrumentation_type: wrapper
  installation:
    type: wrapper
    description: ""
  semantic_conventions:
    - HTTP_SERVER_SPANS
    - HTTP_SERVER_METRICS
  stability: experimental
  telemetry:
    - when: default
      spans:
        - kind: server
          attributes:
            - name: client.address
              type: string
            - name: echo.error
              type: string
            - name: http.request.method
              type: string
            - name: http.response.status_code
              type: int
            - name: http.route
              type: string
            - name: network.peer.address
              type: string
            - name: network.protocol.name
              type: string
            - name: network.protocol.version
              type: string
            - name: server.address
              type: string
            - name: server.port
              type: int
            - name: url.path
              type: string
            - name: url.scheme
              type: string
            - name: user_agent.original
              type: string
      metrics:
        - name: http.server.request.body.size
          type: histogram
          unit: By
          attributes:
            - name: http.request.method
              type: string
            - name: http.response.status_code
              type: int
            - name: http.route
              type: string
            - name: network.protocol.version
              type: string
            - name: url.scheme
              type: string
        - name: http.server.request.duration
          type: histogram
          unit: s
          attributes:
            - name: http.request.method
              type: string
            - name: http.response.status_code
              type: int
            - name: http.route
              type: string
            - name: network.protocol.version
              type: string
            - name: url.scheme
              type: string
        - name: http.server.response.body.size
          type: histogram
          unit: By
          attributes:
            - name: http.request.method
              type: string
            - name: http.response.status_code
              type: int
            - name: http.route
              type: string
            - name: network.protocol.version
              type: string
            - name: url.scheme
              type: string
```

### Example record — `otellogr` (log bridge)

```yaml
- name: bridges-otellogr
  display_name: logr
  description: ""
  source_path: bridges/otellogr
  scope:
    name: go.opentelemetry.io/contrib/bridges/otellogr
  module:
    path: go.opentelemetry.io/contrib/bridges/otellogr
    version: v0.19.0
  target_module: github.com/go-logr/logr
  go_min_version: 1.25.0
  library_link: https://pkg.go.dev/go.opentelemetry.io/contrib/bridges/otellogr
  instrumentation_type: bridge
  installation:
    type: import
    description: ""
  stability: experimental
```

_Generated by `make snapshot`; the nightly workflow integration is tracked separately._

mikeblum:
To address the `uv` expectation from CI ex. `uv sync --locked --all-extras --dev` I'm thinking one approach is a default `Makefile` in `watcher-common` that all the other watchers use for ci and we can have watchers with ecosystem-specific `Makefile`s. There are probably other things we need to tweak to make watchers polyglot writ large.

ecosystem-automation/golang-instrumentation-watcher/AGENTS.md
```diff
@@ -0,0 +1,49 @@
...
+convention used across this repo: the scope is the component path, abbreviated
+to the ecosystem and language.
+
+- **Scope:** `ecosystem-automation/golang` — e.g.
+  `feat(ecosystem-automation/golang): resolve go-contrib release tags`.
+- Use the standard types (`feat`, `fix`, `refactor`, `chore`, `docs`, `test`,
+  `style`, `perf`).
+- No AI attribution, `Co-Authored-By`, or tool links in commit messages.
+- Keep the log reviewable: one logical change per commit, not a play-by-play of
+  the editing session.
```
jaydeluca: personally I prefer AI to not make commits at all, so let's remove this. People who use ai to make commits can put this type of thing in a personal agents file

ecosystem-automation/golang-instrumentation-watcher/AGENTS.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Footguns
+
+- The `insturmentation` typo in the module path and directory name is fixed —
+  the module is `.../golang-instrumentation-watcher`. Don't reintroduce it.
+- Snapshot writes must clean up the prior snapshot and write a replacement in
+  the same run, or the frontend sees a missing snapshot until the next sync.
+- Don't hand-edit files under `ecosystem-registry/` — they are pipeline output
+  and the immutable historical record.
```
jaydeluca: im not sure we need this either

```suggestion

```

ecosystem-automation/golang-instrumentation-watcher/Makefile
```diff
@@ -0,0 +1,89 @@
...
+
+.PHONY: test
+test: ## 🧪 Run all tests
+	go test -test.v -race -covermode=atomic -coverprofile=coverage.out ./... && \
+	go tool cover -html=coverage.out -o coverage.html && \
+	echo "Coverage report saved to coverage.html" && \
+	rm -f coverage.out
+
+.PHONY: test-perf
+test-perf: ## ⚡ Run benchmark tests
```
jaydeluca: there's a ton of stuff in this file that I don't think are necessary, like benchmarks and security scans. Can you just scope this to only what's needed?

ecosystem-automation/golang-instrumentation-watcher/README.md
```diff
@@ -0,0 +1,158 @@
...
+      - # spans/metrics + attributes extracted via Go AST
+```
+
+`name` is the repo-relative module path (slashes normalized to hyphens) so it
+is globally unique even when two modules share a leaf directory (e.g. the v1 and
+v2 `otelmongo` drivers); `display_name` keeps the short, human-facing form.
+
+A **Weaver registry** (`registry/signals.yaml` + `attributes.yaml`) is also
+generated as an optional dev/validation artifact (`make dev` →
+`weaver registry check`), not as the canonical consumer output.
```
jaydeluca: i think we should start small and build up to this.  can we strip this down to just whats needed/included in this first step?

jaydeluca (commented):
i started reviewing this, but wanted to pause and ask if it's possible to start even smaller? it's a bit difficult to review 48 files and 8k lines of code

mikeblum:
> i started reviewing this, but wanted to pause and ask if it's possible to start even smaller? it's a bit difficult to review 48 files and 8k lines of code

struck all the weaver registry work and pushed to another branch. What's left is the `go-contrib` metadata.ymls.

ecosystem-automation/golang-instrumentation-watcher/Makefile
mikeblum: stripped this down to the basics

ecosystem-automation/golang-instrumentation-watcher/AGENTS.md
mikeblum: struck it. I saw AGENTS.md and CLAUDE.md in the repo and thought it was needed.

ecosystem-automation/golang-instrumentation-watcher/AGENTS.md
mikeblum: struck

ecosystem-automation/golang-instrumentation-watcher/README.md
mikeblum: kicked the weaver registry work to another follow up branch

ecosystem-registry/go/contrib/v1.44.0/instrumentation.yaml
mikeblum, line 1: I was unsure about needing both `go/contrib/vX.Y.Z` and `go/contrib/vX.Y.Z-SNAPSHOT`. It get why as its what maven and gradle expect but it feels a bit awkward and redundant here. Is the idea we symlink instead to appease the crawler?

jaydeluca: the `-SNAPSHOT` approach is basically just to track changes that are in main but haven't been released yet. It's not a hard requirement to have this incorporated into the watcher, but for java i found it useful to monitor changes between releases to make sure everything was working properly

jaydeluca: so you can skip it for now if you want

jaydeluca:
to fix all of your CI issues, you need to edit [this file](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/blob/main/pyproject.toml#L17-L20)

and under the "members" block, add:

```
exclude = [
    "ecosystem-automation/golang-instrumentation-watcher",
]
```

ecosystem-automation/golang-instrumentation-watcher/instrumentation/metadata.go
```diff
@@ -0,0 +1,144 @@
...
+		DisplayName:         inferDisplayName(leaf),
+		SourcePath:          sourcePath,
+		Scope:               metadata.Scope{Name: r.Path},
+		Module:              metadata.Module{Path: r.Path, Version: r.Version},
+		TargetModule:        inferTarget(r.Path, leaf),
+		GoMinVersion:        r.GoVersion,
+		LibraryLink:         "https://pkg.go.dev/" + r.Path,
+		InstrumentationType: instrType,
+		Installation:        metadata.Installation{Type: inferInstallType(instrType)},
+		Stability:           metadata.StabilityExperimental,
```
jaydeluca: should this really be hard coded to experimental?

ecosystem-automation/golang-instrumentation-watcher/conf/log.go
```diff
@@ -0,0 +1,59 @@
...
+	"os"
+	"strconv"
+)
+
+const (
+	// ENV_LOG_LEVEL is the environment variable that overrides the log level.
+	ENV_LOG_LEVEL = "LOG_LEVEL"
+
+	// LOG_LEVEL_INFO is the default log level used when [ENV_LOG_LEVEL] is unset.
+	LOG_LEVEL_INFO = "info"
```
jaydeluca: this is unused, is it needed in the followup?

ecosystem-automation/golang-instrumentation-watcher/instrumentation/parser.go
```diff
@@ -0,0 +1,17 @@
+package instrumentation
+
+var displayNameMap = map[string]string{
```
jaydeluca, line 3: any ideas on how this could eventually make it's way upstream to avoid us having to maintain a map?

ecosystem-automation/golang-instrumentation-watcher/metadata/metadata.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// Stability is the maturity level of an instrumentation library. Its YAML
+// tokens are "development", "experimental", and "stable".
+type Stability int
+
+const (
+	StabilityDevelopment  Stability = iota // early-stage; API may change
+	StabilityExperimental                  // feature-complete but not yet production-ready
+	StabilityStable                        // stable API, production-ready
```
jaydeluca: is this based on something? in java we only have experimental and stable

ecosystem-registry/go/contrib/v1.44.1-SNAPSHOT/instrumentation.yaml
```diff
@@ -0,0 +1,948 @@
...
+    go_min_version: 1.25.0
+    library_link: https://pkg.go.dev/go.opentelemetry.io/contrib/bridges/otellogrus
+    instrumentation_type: bridge
+    installation:
+      type: import
+      description: ""
+    stability: experimental
+  - name: bridges-otelslog
+    display_name: slog
+    description: ""
```
jaydeluca: should probably conditionally render these only when there is an actual description

ecosystem-registry/go/contrib/v1.44.1-SNAPSHOT/instrumentation.yaml
```diff
@@ -0,0 +1,948 @@
...
+    stability: experimental
+    telemetry:
+      - when: default
+        spans:
+          - kind: server
+            attributes:
+              - name: cloud.resource_id
+                type: string
+              - name: faas.invocation_id
+                type: string
```
jaydeluca: this should come in the followup i think, guessing you need to regenerate these files

ecosystem-automation/golang-instrumentation-watcher/repo/repo_test.go
```diff
@@ -0,0 +1,236 @@
...
+func setupGitRepo(t *testing.T, tmpDir string, cmds [][]string) {
+	t.Helper()
+	for _, cmd := range cmds {
+		c := exec.Command(cmd[0], cmd[1:]...)
+		c.Dir = tmpDir
+		if err := c.Run(); err != nil {
+			t.Fatalf("failed to run %v: %v", cmd, err)
+		}
+	}
+}
```
jaydeluca, line 34: The git-backed tests (`TestInfo`, `TestTagsAt`, `TestCheckout`) create temp repos that inherit the developer's **global** git config. On a machine with commit signing enabled (`commit.gpgsign=true` + `gpg.format=ssh`), every `git commit --allow-empty` tries to sign with the SSH key and fails with `exit status 128`. Setting local `user.email`/`user.name` (which the tests already do) isn't enough — signing config is inherited from global/system config. CI passes today only because the CI environment has no signing configured.

Suggest isolating each git invocation from global/system config in `setupGitRepo`:

```go
c := exec.Command(cmd[0], cmd[1:]...)
c.Dir = tmpDir
// Isolate from the developer's global/system git config so commits don't inherit
// settings like commit.gpgsign (SSH/GPG signing), which would make `git commit`
// fail or block on a key in these hermetic tests.
c.Env = append(os.Environ(),
	"GIT_CONFIG_GLOBAL=/dev/null",
	"GIT_CONFIG_SYSTEM=/dev/null",
)
```

This keeps the per-repo `user.email`/`user.name` (local config) while dropping inherited signing settings, making the tests hermetic regardless of the contributor's machine.

ecosystem-automation/golang-instrumentation-watcher/Makefile
```diff
@@ -0,0 +1,50 @@
...
+	go tool cover -html=coverage.out -o coverage.html && \
+	echo "Coverage report saved to coverage.html" && \
+	rm -f coverage.out
+
+.PHONY: test-integration
+test-integration: ## 🌐 Run integration tests (clones upstream repos)
+	go test -tags=integration -race ./...
+
+.PHONY: pre-commit
+pre-commit: fmt tidy lint test ## ✅ Run all checks
```
jaydeluca, line 46: `pre-commit: fmt tidy lint test` depends on a `lint` target that doesn't exist. The catch-all pattern rule below (`%:` → `@:`) silently absorbs `lint` as a do-nothing target, so `make pre-commit` reports success while running no linter at all. There's some `gofmt` drift in `walker.go` that this would otherwise catch.

Either add a real target, e.g.:

```makefile
.PHONY: lint
lint: ## 🔍 Vet and check formatting
	go vet ./...
	test -z "$$(gofmt -l .)" || { echo "unformatted files:"; gofmt -l .; exit 1; }
```

or drop `lint` from `pre-commit` so the gap isn't hidden behind a passing run.

ecosystem-automation/golang-instrumentation-watcher/Makefile
```diff
@@ -0,0 +1,50 @@
+##r 🔭 opentelemetry-ecosystem-explorer/golang-instrumentation-watcher 🔭
```
jaydeluca: Stray `r` in the header comment: `##r 🔭 ...` → `## 🔭 ...`.

ecosystem-automation/golang-instrumentation-watcher/repo/repo.go
```diff
@@ -0,0 +1,235 @@
...
+	"https://github.com/open-telemetry/opentelemetry-go-contrib.git",
+}
+
+// RepoInfo identifies a checked-out repository and its current commit. It is
+// returned by [CheckoutAt].
+type RepoInfo struct {
+	Name    string // repository name (e.g. "opentelemetry-go-contrib")
+	Path    string // absolute filesystem path to the checkout
+	Head    string // short SHA of the checked-out commit
+	SHA     string // full commit SHA
```
jaydeluca, line 36: The struct field is documented as the **full** commit SHA, but `info()` (line 116) truncates it to 8 chars (`SHA: sha[:shaLength]`, `shaLength = 8`), so `SHA` ends up a near-duplicate of `Head` (`rev-parse --short`). The full 40-char value from `--format=%H` is fetched and then thrown away.

Pick one: keep the full SHA (matches the field's stated contract and is more useful as provenance in the inventory pipeline) by dropping the `[:shaLength]` slice, or fix the doc comment to say "short (8-char) commit SHA". As-is, `Head` and `SHA` carry essentially the same information.

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library.go
```diff
@@ -0,0 +1,86 @@
...
+		if err != nil {
+			continue
+		}
+		for _, pkg := range packages {
+			lib, err := analyzeLibrary(pkg.GoModPath)
+			if err != nil || lib == nil {
+				continue
+			}
+			libraries = append(libraries, *lib)
+		}
```
jaydeluca: Scan errors are swallowed silently here: a failed `Walk` (`if err != nil { continue }`) skips an entire subtree, and a failed `analyzeLibrary` drops a single library — both with no log line.

Per the schema discipline in `ecosystem-automation/AGENTS.md` ("All expected files must be present... Partial writes leave the registry in an inconsistent state that persists in Git history"), a transient FS or parse error would produce a smaller library set that gets committed and reads as *libraries removed* in the registry diff. `ScanRepo` logs nothing today, so this would be invisible.

Let's fail hard on `Walk` errors — a missing `instrumentation/` subtree is not something we want to silently ship as an empty inventory.

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library.go
```diff
@@ -0,0 +1,86 @@
...
+	default:
+		// Only the subtrees that instrument a developer's code: instrumentation
+		// wrappers (gin, grpc, http…) and bridges (zap, logrus…). The other
+		// go-contrib components (exporters, propagators, samplers, detectors,
+		// processors) configure the SDK pipeline rather than instrument a target
+		// library, so they have no target_module and are out of scope here.
+		for _, sub := range []string{"instrumentation", "bridges"} {
+			scanPaths = append(scanPaths, filepath.Join(repoPath, sub))
+		}
+	}
```
jaydeluca: The `case repo.RepoGo:` branch (whole-tree scan) is unreachable on this branch — only `opentelemetry-go-contrib` is ever cloned (`repo.repos` has a single entry). Can we drop `RepoGo` and this branch until we actually tackle scanning `opentelemetry-go`?

ecosystem-automation/golang-instrumentation-watcher/instrumentation/walker.go
```diff
@@ -0,0 +1,60 @@
...
+
+func omitDirectory(name string) bool {
+	for _, dir := range omitDirectories {
+		if name == dir {
+			return true
+		}
+	}
+	return false
+}
+
```
jaydeluca: Not gofmt-clean — trailing blank line at end of file. `make fmt` (or `gofmt -w`) fixes it. (Related to the `lint` no-op in the Makefile.)

jaydeluca (commented):
A few things to address before merge:

1. **Git-backed tests aren't hermetic** — they inherit the developer's global git config and fail under SSH/GPG commit signing (`exit status 128`). I have a one-line fix; left it as an inline comment on `repo_test.go`. Tests pass in CI today only because CI has no signing config, but this bites any contributor who signs commits.
2. **`make pre-commit` runs no linter** — the `lint` target doesn't exist and is silently absorbed by the catch-all rule, so the check passes while linting nothing (this is how the `gofmt` drift in `walker.go` slipped through).
3. **`RepoInfo.SHA` is documented as the full SHA but truncated to 8 chars.**

Plus a couple of optional cleanups noted inline: silent error-swallowing in `ScanRepo` vs. the "all files present" schema rule, and the unreachable `RepoGo` scan branch.

Note: no CI gate enforces `gofmt`/`go vet` yet since the watcher isn't wired into nightly CI — worth a minimal Go job (vet + gofmt check + `go test`) when it is.

ecosystem-automation/golang-instrumentation-watcher/instrumentation/metadata.go
mikeblum: in Go pre-1.x modules are considered experimental but not necessarily unstable. Even widely used ones like `k8s.io/client-go` at [v0.36.2](https://github.com/kubernetes/client-go/releases/tag/v0.36.2). Its expected that there are no silent breaking changes after 1.x and if thre are to bump the mid-var.

ecosystem-automation/golang-instrumentation-watcher/instrumentation/parser.go
mikeblum: My approach is to get this branch running as a POC before going to the Go SIG to work out how we might push these upstream. I think it will be a tighter loop if there's working code to get feedback on vs a doc / rfc.

ecosystem-automation/golang-instrumentation-watcher/metadata/metadata.go
mikeblum: I was referencing this doc here: https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/0232-maturity-of-otel.md#development

I think development applies for Go as the compile-time and eBPF OTel insturmentation projects are very much under active development.

ecosystem-registry/go/contrib/v1.44.1-SNAPSHOT/instrumentation.yaml
mikeblum: yep will re-run and clean these up. Realized after the fact that i pushed up some WIP artifacts.

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library.go
mikeblum: Agreed. Added tests for this as well.

mikeblum: Nice catch. Added tests around this to catch regressions as we add the transformation layer back.

ecosystem-automation/golang-instrumentation-watcher/Makefile
mikeblum: fixed this up and made `make pre-commit` mimic the CI with `bun run format`

ecosystem-automation/golang-instrumentation-watcher/repo/repo_test.go
mikeblum: fixed! talk about works on my machine 🫠

ecosystem-registry/go/contrib/v1.44.1-SNAPSHOT/instrumentation.yaml
mikeblum: yep struck

ecosystem-automation/golang-instrumentation-watcher/instrumentation/metadata.go
jaydeluca: so you are saying that as of now, there are zero stable isntrumentations?

mikeblum: I thought there weren't any but upon review there's https://pkg.go.dev/go.opentelemetry.io/contrib which is funny as its the parent "module" that holds all the other modules. Pushed up a fix with tests.

ecosystem-automation/golang-instrumentation-watcher/conf/log.go
mikeblum: removed

jaydeluca (approved):
LGTM, thanks!

One thing to pick up in a followup: the stability inference added in "fix infer stability" is effectively a no-op in the real pipeline. ParseModule always returns an empty Version (by design — a module's own go.mod doesn't record its release version), so DeriveMetadata → inferStability(r.Version) always evaluates to experimental. The real version only lands later via ApplyModuleVersions, which sets Module.Version but never recomputes Stability, so the case "v1": StabilityStable branch can never fire in production. TestDeriveMetadata passes because it seeds Version directly — an input shape the production caller never produces.

mikeblum:
> LGTM, thanks!
> 
> One thing to pick up in a followup: the stability inference added in "fix infer stability" is effectively a no-op in the real pipeline. ParseModule always returns an empty Version (by design — a module's own go.mod doesn't record its release version), so DeriveMetadata → inferStability(r.Version) always evaluates to experimental. The real version only lands later via ApplyModuleVersions, which sets Module.Version but never recomputes Stability, so the case "v1": StabilityStable branch can never fire in production. TestDeriveMetadata passes because it seeds Version directly — an input shape the production caller never produces.

Yea I'm looking to use this with the exemplar apps / pinned modules in the go.mod for generating the instrumentation yamls. But yes atm it's a no-op on this branch.

PR #672, opened 2026-06-08
ci: persist otelbot App token so automated PR pushes trigger workflows

vitorvasc:
## What

The **Build Explorer Database** and **Nightly Registry Update** workflows checked out with the default `GITHUB_TOKEN` and persisted it as the git credential, so the `git push -f` to the automated-update branch authenticated as `GITHUB_TOKEN`. The `GH_TOKEN: <otelbot App token>` env on the push step only applies to the `gh` CLI (`gh pr create`/`gh pr edit`), **not** to `git push`.

Pushes authenticated with `GITHUB_TOKEN` [do not trigger downstream workflows](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow). So when a second run force-updated an **already-open** PR, CodeQL never re-ran for the new head and the PR stayed blocked on the "Code scanning results" merge rule (most recently #660). The first run only worked because `gh pr create` runs with the App token and fires the `pull_request: opened` event.

## Fix (both workflows)

- Mint the otelbot App token **before** checkout and persist it as the checkout credential on the primary repo, so `git push -f` authenticates as the App and triggers CodeQL on the PR. Forks have no App secret and keep falling back to `GITHUB_TOKEN`.
- Add `permission-contents: write` to the App token. `create-github-app-token` scopes the token to exactly the requested `permission-*` inputs, and the previous config only requested `pull-requests: write` — so the token lacked the `contents: write` the push needs.
- Reorder steps so `Configure git (primary repository)` (which runs `use-cla-approved-bot.sh` against the checked-out repo) now runs after checkout.

## Scope

Only the two PR-creating workflows are affected. The `screenshots-*` workflows push to the `otelbot/screenshots` data branch with `GITHUB_TOKEN` **on purpose** (they must not trigger downstream runs) and are left unchanged.

## Verification

- `python3 -c yaml.safe_load` parses both files.
- Project Prettier (`prettier --check`) passes on both files.

---

### DIFF
--- a/.github/workflows/build-explorer-database.yml
+++ b/.github/workflows/build-explorer-database.yml
@@ -39,14 +39,6 @@ jobs:
       pull-requests: write # Create or update the automated database-update PR
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout code
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-        with:
-          token: ${{ secrets.GITHUB_TOKEN }}
-          # Persisted credentials are required so the later `git push -f origin
-          # HEAD:${BRANCH}` step can authenticate.
-          persist-credentials: true
-
       - name: Detect repository type (if on a fork we use different git config)
         id: repo_check
         env:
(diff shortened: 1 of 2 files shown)


jaydeluca (approved):

jaydeluca:
@vitorvasc this doesn't seem to work for the nightly registry update: https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/actions/runs/27181496549/job/80292438913

vitorvasc:
@jaydeluca I'll take a look at it 👀

vitorvasc:
@jaydeluca TIL we have a different bot running in the .io repo that has permissions to commit and push to the branches there. I tried doing the same thing here, but the otelbot app we're using is missing those permissions. 😅 

Just opened #678 and changed the behavior to trigger CodeQL by closing/reopening the PR using the same current bot

PR #673, opened 2026-06-09
Fix/658 structured list properties

### LINKED ISSUE #658: [Config Builder] Support structured list properties
### Expected behavior

The java.common.service_peer_mapping configuration is a "structured list", meaning that each entry should have the [structure](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/524c1edca6f9681adff6fe97d228181bf9640907/declarative-config-bridge/src/main/java/io/opentelemetry/instrumentation/config/bridge/ServicePeerMapping.java#L29-L33) of:

```
instrumentation/development:
  java:
    common:
      service_peer_mapping:
        peer: 1.1.1.1
        service_name: test-service
```

This is tricky because it is different from the standard system property structure, which is a map. We probably need to include some custom mappings. url template rules have a similar structured list
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/instrumentation/http/RegexUrlTemplateCustomizerInitializer.java

I have a [PR to identify these types of declarative configs](https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/18883), but I'm still working out the pattern. We could add a targeted shim/converter in the explorer until upstream is sorted

### Actual behavior

Right now, we are rendering the inputs as standard map properties, so it results in:

Image

### Steps to reproduce

go to https://explorer.opentelemetry.io/java-agent/configuration/builder

Navigate to the vertx_http_client instrumentation, open the `java.common.service_peer_mapping` config and set values. See the results in the YAML

### Affected page

https://explorer.opentelemetry.io/java-agent/configuration/builder

### Evidence (screenshots, logs)  of the occuring bug

url template rules have a similar structured list
https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/javaagent-tooling/src/main/java/io/opentelemetry/javaagent/tooling/instrumentation/http/RegexUrlTemplateCustomizerInitializer.java

DCchoudhury15:
## What changed

Added support for structured list configurations in the declarative config builder.

### Backend (`explorer-db-builder`)
- Added schema metadata for `java.common.service_peer_mapping`
- Added schema metadata for declarative configs ending with `url_template_rules`
- Updated schema mapping to recognize `structured_list` configs and expose the injected schema as the list item schema

### Frontend (`ecosystem-explorer`)
- Added support for `declarative_type` and `declarative_schema`
- Added a dedicated `StructuredListRenderer` for rendering structured list entries
- Updated customization handling to initialize structured lists correctly
- Updated validation logic to accept array values for structured list configurations

## Why

Fixes #658.

`service_peer_mapping` and `url_template_rules` are structured configurations, but were previously rendered as generic key/value maps. This made it unclear which fields were expected and resulted in an inconsistent editing experience.

This change allows these configurations to be rendered using their schema definition, producing a more intuitive UI and correct YAML output.

## Screenshot

image

### Manual
- Opened `java.common.service_peer_mapping` in the config builder
- Added a sample mapping entry
- Verified the generated YAML matches the expected structure

```yaml
instrumentation/development:
  java:
    common:
      service_peer_mapping:
        - peer: 1.1.1.1
          service_name: test-service

### CODE REVIEW

**Comment by DCchoudhury15** [comment]
@jaydeluca pr up for review

File: ecosystem-automation/explorer-db-builder/src/explorer_db_builder/declarative_name_corrections.py [hunk 1]
```diff
@@ -51,15 +51,55 @@ def apply_declarative_name_corrections(inventory: dict[str, Any]) -> dict[str, A
...
                     config["declarative_name"] = corrected
                     logger.debug(
                         "Corrected declarative_name %r -> %r for config %r",
-                        current,
+                        original_name,
                         corrected,
                         config.get("name"),
                     )
 
+                current_name = config.get("declarative_name")
```
jaydeluca, line 66: could you update the tests in test_declarative_name_corrections.py to account for these too?

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/schema_ui_mapper.py
```diff
@@ -37,6 +37,10 @@ def _generate_label(key: str) -> str:
 
 def _classify_node(node: dict[str, Any]) -> str:
     """Classify a schema node into a UI control type."""
+
+    if node.get("declarative_type") == "structured_list":
```
jaydeluca: i could be missing something, but i dont think the changes in this file are needed. I removed it and re-ran the build and everything still works

jaydeluca (commented):
just a couple comments from myself and copilot to address but this looks good, nice work.

If you are interested, there are a few other places we could update to use this new declarative_type and schema as well, like the instrumentation detail page configuration tab (https://explorer.opentelemetry.io/java-agent/instrumentation/vertx-http-client-4.0) and the config explorer: https://explorer.opentelemetry.io/java-agent/configuration

I can create an issue to handle those in a followup if you'd rather just focus on the current scope of this PR, just figured i'd mention it

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/schema_ui_mapper.py
DCchoudhury15: Ah, my apologies , you're absolutely right.
I updated `schema_ui_mapper.py` because I assumed `structured_list` still needed to be mapped to a standard `list` control on the backend. I completely overlooked that `StructuredListRenderer` already handles the declarative type directly on the frontend, so the change wasn't needed at all.
Sorry for the unnecessary noise in the PR. I've reverted the file. Thanks for catching that!

DCchoudhury15:
> just a couple comments from myself and copilot to address but this looks good, nice work.
> 
> If you are interested, there are a few other places we could update to use this new declarative_type and schema as well, like the instrumentation detail page configuration tab (https://explorer.opentelemetry.io/java-agent/instrumentation/vertx-http-client-4.0) and the config explorer: https://explorer.opentelemetry.io/java-agent/configuration
> 
> I can create an issue to handle those in a followup if you'd rather just focus on the current scope of this PR, just figured i'd mention it

Thanks @jaydeluca! I'm really glad the implementation looks good. 

I would absolutely love to work on extending this feature to the instrumentation detail page and the config explorer! Please go ahead and create the follow-up issue you can tag or assign me directly to it, and I'll start to work on it as soon as this PR is merged. 

I've pushed the fixes covering the unit tests and the edge cases Copilot flagged, so this should be ready for another look! Thanks for your time and patience.

ecosystem-explorer/src/features/java-agent/configuration/components/instrumentation-config-field.tsx
```diff
@@ -225,6 +225,106 @@ function KeyValueMapRenderer({
...
+  schema,
+}: ControlRendererProps & { schema: NonNullable<Configuration["declarative_schema"]> }) {
+  const items = Array.isArray(value) ? (value as ConfigValues[]) : [];
+  const properties = Object.entries(schema.properties);
+
+  return (
+    <div className="w-full max-w-xl space-y-2" aria-label={ariaLabel}>
+      {items.map((item, idx) => {
+        const stableKey =
+          typeof (item as Record<string, unknown>)._id === "string"
```
jaydeluca: this `_id` is leaking into the YAML output now

image


could we keep `_id` purely as a render key and not store it in the value? We could Track stable keys with a `useRef`/`WeakMap` keyed by item identity. If that ends up being too complex we can just accept index keys (the previous behavior) since these lists are short and append-only.

ecosystem-explorer/src/types/javaagent.ts
```diff
@@ -110,6 +110,17 @@ export interface Configuration {
...
   examples?: string[];
+  /** When set to "structured_list", each entry is an object described by declarative_schema. */
+  declarative_type?: "structured_list";
+  /** Per-item schema used when declarative_type is "structured_list". */
+  declarative_schema?: {
+    type: "object";
+    required?: string[];
+    properties: Record<
+      string,
+      { type: "string" | "boolean" | "number" | "integer"; description?: string; default?: unknown }
```
jaydeluca: RIght now the code path for numbers wouldn't work, we don't have the proper handling. I think we'd need to add something similar to `NumberRenderer`. Since the only 2 configs with structure don't have numbers and it complicates things, we can just slim this down and i'll create a followup issue to handle numbers

```suggestion
      { type: "string" | "boolean"; description?: string; default?: string | boolean }
```

ecosystem-explorer/src/features/java-agent/configuration/components/instrumentation-config-field.tsx
DCchoudhury15: Sorry about that! I added `_id` for stable keys based on a Copilot suggestion, but I didn't realize it would leak into the YAML output.
You're right that `idx` is sufficient here. I've reverted the change, removed `_id`, and verified locally that the YAML output is clean now.
Thanks for catching it!

jaydeluca (approved):
nice, thanks!

PR #683, opened 2026-06-09
Move Java Agent search to the index and add a slim list type

### LINKED ISSUE #657: Use a dedicated type for the javaagent list bundle entries
The javaagent list bundle entries and the full instrumentation files are both typed as `InstrumentationData`, where almost every field is optional. TypeScript can't tell the two shapes apart, so a consumer reading a field the bundle doesn't carry compiles fine and only fails at runtime.

The collector side already handles this with a dedicated `IndexComponent` type, separate from `CollectorComponent`. The javaagent side should do the same: an `InstrumentationListEntry` interface for the bundle shape, with `loadAllInstrumentations` returning it (fallback projected to the same shape). The current consumers all type-check against the slim shape, so it's signature-only changes in ~10 files, plus `InstrumentationModule.coveredEntries` switching to the new type.

With this, dropping a field from the bundle becomes a compile error instead of a silent runtime regression.

lucacavenaghi97:
Java Agent global search now reads index.json (with precomputed search_terms) instead of the list bundle, so it no longer depends on fields the slim bundle drops. Also adds a dedicated InstrumentationListEntry type for the bundle/list path (mirrors the collector IndexComponent) so list and config-builder consumers type-check against the slim shape.

search_terms is populated by a db rebuild. Until then the search source falls back to name/display_name/description, so nothing breaks at merge.

Closes #657


vitorvasc (approved):

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/instrumentation_transformer.py
```diff
@@ -180,6 +180,87 @@ def _transform_0_3_to_0_5(inventory_data: dict[str, Any]) -> dict[str, Any]:
...
+        # scalars (e.g. minimum_java_version), and drop blanks.
+        if value is None or isinstance(value, (dict, list)):
+            return
+        text = str(value).strip()
+        if text:
+            terms.add(text)
+
+    add(instrumentation.get("library_link"))
+    add(instrumentation.get("source_path"))
+    add(instrumentation.get("minimum_java_version"))
```
jaydeluca: I think we can probably limit this to only distinctive, identifier-like strings like config name/declarative_name, metric name, scope.name, the maven coordinate (range stripped), semantic_conventions, features, and descriptions. We can probably drop the individual attribute names, types etc and everything else

jaydeluca (approved):

PR #691, opened 2026-06-12
fix(collector-watcher): exclude comments from schema cas hash

### LINKED ISSUE #583: Exclude comments from collector schema hash generation
In the most recent registry update (#582) I noticed that the schema_hash changed for all the components, which is due to a [new comment being added](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/582/files#diff-326a9c6ad45b2ba45dbcb77bad703baba291e8724c80ac927c7bba227e64e665R23-R24) 

We should update the [logic that generates the hashes](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/blob/main/ecosystem-automation/collector-watcher/src/collector_watcher/schema_copier.py#L64) to exclude comments so that we only bump the hash on actual changes

Pittu-Sharma:
## What changed

Updated `CollectorSchemaCopier.compute_schema_hash` to parse YAML schemas into a pure Python data structure before serializing them back into a deterministic string for hashing. Added a unit test to guarantee that comments and whitespaces are inherently ignored.

## Why

Fixes #583.
Previously, if a developer added a comment to their component's `metadata-schema.yaml`, it would alter the file's text and incorrectly trigger a brand-new CAS file allocation in the registry. This fix ensures that only actual logical changes to the schema result in a new hash.

## How it was tested

- Ran the newly added `test_compute_schema_hash_ignores_comments` unit test to verify deterministic behavior.
- Manually tested by generating CAS hashes for a YAML file with and without comments, validating they produce the same hash.
- Full `uv run pytest` test suite passed locally.

## Is this a breaking change?

No.

## Checklist


Pittu-Sharma:
@jaydeluca, Sir please review the PR once. Waiting for your suggestion. If any changes are required in the PR. Please do let me know.

jaydeluca (approved):
thanks. Btw, when I ran the full backfill locally I noticed that the results are not correct, not due to this PR but due to another bug in the backfill, so I will be doing a fast follow to fix that

Issue #372, opened 2026-05-06
[UI/UX] Phase 3: Ecosystem landing


vitorvasc:
Part of #84. Depends on #370.

This issue tracks the per-ecosystem landing pages (Collector, Java Agent, …). Each ecosystem gets a contextual hero with a "Latest release" delta card, an interactive **Pipeline anatomy diagram** that doubles as a one-click filter into the list page, and three quick-entry cards.

See [`projects/84-ui-ux-design/02-ecosystem-landing.md`](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/blob/main/projects/84-ui-ux-design/02-ecosystem-landing.md) for the full plan.

The components will be shipped behind the `V1_REDESIGN` feature flag and depend on the foundation primitives from #370 + the `listFilters` URL contract from Phase 4 (so the pipeline-anatomy stages can deep-link into the filtered list). The two phases can run in parallel once that contract is locked.

## PRs

- [x] PR 1 - `listFilters` URL contract (`list-filters.ts`) - the deep-link parser/serializer shared with Phase 4; unblocks `<PipelineAnatomy>` stage links - #685 
- [x] PR 2 - `<ReleaseCard>` component (version, release date, deltas) - #686
- [x] PR 3 - `<PipelineAnatomy>` component (array-driven, mobile-responsive, deep-links into the list) - #687
- [x] PR 4 - `<QuickEntryRow>` component - #688
- [x] PR 5 - `EcosystemConfig` TypeScript schema (hero copy, release-fetch source, pipeline stages, quick-entry items) - #708 
- [x] PR 6 - `/collector` route - composes the above with Collector config - #708 
- [x] PR 7 - `/java-agent` route - Java Agent config, alternate "categories" diagram - #708 
- [x] PR 8 - Release-data integration (read from the index-based data layer; fall back to static config) - #711
- [x] PR 9 - Empty/error states + visual regression - #758

LoveChauhan-18:
Hi, I would like to work on this issue. Please assign it to me!

vitorvasc:
Hey @LoveChauhan-18, thanks for your interest!

I've been actively working on the redesign locally and most of the components are already implemented, my apologies. The main challenge at this point is breaking it into smaller, but reviewable PRs and handling conflicts carefully as we merge, so we don't break any existing or in-flight features.

[...]

PR #708, opened 2026-06-18
feat(v1): add Collector and Java Agent landing pages

vitorvasc:
## What changed

- Adds a declarative `EcosystemConfig` schema and a single `EcosystemPage` composition that renders every per-ecosystem landing: a hero with a release card, the pipeline-anatomy strip, and a quick-entry row. 
- Ships the Collector and Java Agent configs and points the `/collector` and `/java-agent` routes at the new V1 landings. 
- The route swap lives inside `V1App`, so the pages surface only behind `V1_REDESIGN`.

## Why

Contributes to #372. This bundles PRs 5, 6, and 7 from that issue so the routes render something real rather than landing a types-only contract on its own.

## How it was tested

- `bun run test` - 1057 pass, including 6 new cases in `ecosystem-page.test.tsx` (hero/release, the five Collector stage deep-links, quick-entries, breadcrumbs, Java Agent categories + release)
- `bun run typecheck`, `bun run lint`, `prettier --check` - clean
- Manual, with `V1_REDESIGN` on: `/collector` and `/java-agent` render the landings; clicking a pipeline stage deep-links into the filtered list

## Is this a breaking change?

No. The route swaps live inside `V1App`, gated by `V1_REDESIGN` (off by default), so the legacy pages keep serving every user until the flag flips.

- Java Agent category stages deep-link via `?search=` (the param the instrumentation list actually reads). Match quality for terms like "framework" is a follow-up once a real category facet exists.
- The "Compare releases" quick-entry is gated on `JAVA_RELEASE_COMPARISON`, so it stays hidden when that route is disabled instead of 404ing.
- Per-ecosystem copy is hardcoded in `configs.tsx` (the content layer). Moving it into locale files for non-English ecosystems is a later step.

## Checklist

---

> Co-reviewed by Claude Opus 4.8.

### DIFF
--- a/ecosystem-explorer/src/v1/features/ecosystem/configs.tsx
+++ b/ecosystem-explorer/src/v1/features/ecosystem/configs.tsx
@@ -28,6 +28,7 @@ import { ArrowLeftRight, Box, GitCompare, Layers } from "lucide-react";
 import { TYPE_STRIPE_COLORS } from "@/components/ui/type-stripe-colors";
 import { OtelLogo } from "@/components/icons/otel-logo";
 import { JavaIcon } from "@/components/icons/java-icon";
+import { isEnabled } from "@/lib/feature-flags";
 import { filtersToHref } from "@/v1/lib/list-filters";
 import type { EcosystemConfig } from "./types";
 
@@ -35,8 +36,6 @@ const collectorComponentsPath = "/collector/components";
 const javaInstrumentationPath = "/java-agent/instrumentation";
 
 export const collectorConfig: EcosystemConfig = {
-  id: "collector",
-  slug: "collector",
   name: "OpenTelemetry Collector",
   hero: {
     eyebrow: "Infrastructure · Vendor-agnostic agent",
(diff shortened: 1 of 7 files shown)


ecosystem-explorer/src/v1/features/ecosystem/configs.tsx
```diff
@@ -0,0 +1,246 @@
...
+  },
+  pipelineTitle: "Instrumentation categories",
+  pipelineLead:
+    "Java auto-instrumentations are grouped by category. Click any category to filter the instrumentation list.",
+  pipelineNoFlow: true,
+  stages: [
+    {
+      id: "http",
+      label: "HTTP",
+      count: "32",
```
jaydeluca, line 172: what is the long term plan for this type of thing? are you keeping track of all the areas we'll need to swap out for dynamic updates?

jaydeluca (approved):

ecosystem-explorer/src/v1/features/ecosystem/configs.tsx
vitorvasc: they were mostly commented in each file to revisit when cleaning up after the development phase; but I also added it to the `NEXT-STEPS.md` just to be safe

PR #712, opened 2026-06-19
fix(config-builder): Improve instrumentation toggle UX and state clarity

### LINKED ISSUE #654: [Config Builder] Instrumentation enabled toggles
Right now we render modules with their enabled/disabled by default status, and a "Customize" button to change it. I find myself thinking that the "Customize" button means "customize the entire module", but then has the effect of disabling the instrumentation when clicking it.

Default view:

Image

Clicking "customize" toggles the other config:

Image

I'm thinking it might be more intuitive if we remove the "Customize" button entirely, and just render them with the Enabled / Disabled based on their default. The challenge is that clicking customize also adds the instrumentation to either the disabled/enabled list, so it's complicated.

I will need to think more about this but wanted to create an issue as a placeholder while I was thinking of it.

Adii-45:
## What changed

This PR refactors the instrumentation customization workflow in the Config Builder to make instrumentation state management more intuitive and easier to understand.

### UI/UX Improvements

- Removed the ambiguous "Customize" action from instrumentation rows
- Made the instrumentation row itself the primary interaction for expanding/collapsing details
- Added explicit Enabled / Disabled controls within the expanded instrumentation view
- Improved visual separation between:
  - instrumentation state management
  - instrumentation configuration/customization
- Simplified status presentation to reduce visual noise
- Improved visibility of the current instrumentation state

### State Management Improvements

- Consolidated instrumentation state handling into a single source of truth
- Removed confusing behavior where customization actions implicitly changed instrumentation state
- Ensured instrumentation status persists correctly across expand/collapse interactions
- Kept generated YAML output synchronized with UI state changes

### Testing

- Updated affected component tests
- Updated reducer and hook tests where necessary
- Verified instrumentation enable/disable state updates correctly
- Verified generated YAML reflects the current instrumentation state
- Verified state persists correctly when expanding/collapsing instrumentation rows

## Why

The previous UX overloaded the "Customize" action with multiple responsibilities:

1. Adding an instrumentation to the customization list
2. Implicitly modifying its enabled/disabled state

This made the workflow difficult to understand and resulted in confusing state transitions.

This update separates configuration actions from state-management actions and makes instrumentation status explicit and directly controllable.

Fixes #654

## How it was tested

- Ran the full test suite successfully
- Verified build passes successfully
- Manually tested:
  - Enable → Disable → Enable flows
  - Expand/collapse behavior
  - Reset-to-default behavior
  - YAML output synchronization
  - Multiple instrumentation modules

## Is this a breaking change?

No.

This change only affects the Config Builder UI and does not change generated configuration formats or public APIs.

## Special notes for reviewer

The goal of this PR was to preserve existing functionality while making instrumentation state management more explicit and intuitive!
The implementation intentionally minimizes changes outside the instrumentation configuration flow!

## Screenshots

| Enabled State | Disabled State |
|---------------|----------------|
|  |  |


Adii-45:
Hi @jaydeluca,

I've incorporated the fixes and updates in the issue and PR. All CI checks are now passing!

Whenever you have some time, I'd appreciate any feedback or a review.. Thanks!!

Adii-45:
Addressed all Copilot review comments and follow-up fixes! All checks are now passing - feedback and further review are welcome!

Adii-45:
Hi @jaydeluca, just following up on this PR — would appreciate a review whenever you have a chance. Thanks!

ecosystem-explorer/src/features/java-agent/configuration/components/instrumentation-row.tsx
```diff
@@ -154,6 +141,26 @@ export function InstrumentationRow({
           className="bg-surface-card px-4 py-4 sm:px-5 sm:py-5"
           data-yaml-section-key="instrumentation/development"
         >
+          <div className="border-border/40 mb-6 flex items-center justify-between border-b pb-4">
+            <div>
+              <h5 className="text-foreground text-sm font-semibold">Instrumentation Status</h5>
```
jaydeluca: A few new user-facing strings here are hardcoded rather than going through `t()`:
- `"Instrumentation Status"` (here, line 146)
- `"Explicitly enable or disable this instrumentation."` (line 148)
- `"Reset to default"` (line 159)
- the `"Default:"` prefix (line 108)

Localisation was recently implemented across the app, so we now need to make sure new strings like these go through `t()` and stay in sync across language files. Could you move these into `builder.row.*` keys and add the matching keys to both `en` and `es` `java-agent.json`? For line 108 you can likely reuse the existing `builder.row.enabledByDefault` / `disabledByDefault` keys.

ecosystem-explorer/src/hooks/configuration-builder-reducer.ts
```diff
@@ -74,7 +98,7 @@ export function configurationBuilderReducer(
       delete remainingErrors[pathKey];
       return {
         ...state,
-        values: setByPath(state.values, action.path, action.value),
+        values: cleanInstrumentation(setByPath(state.values, action.path, action.value)),
```
jaydeluca, line 101: Routing reset-to-default through `SET_VALUE` with `value: null` + `cleanInstrumentation` is a clean way to avoid the `enabled: null` / empty-`{}` leak — but the old `SET_CUSTOMIZATION` suite had an explicit "removes a module when status is none" test that was removed and not replaced, so this cleanup is now only covered indirectly. Could you add a direct reducer test: `SET_VALUE` at `[...instrumentation, "cassandra", "enabled"]` with `value: null` asserts `state.values.distribution` is `undefined` (and that a sibling customized module survives)? That locks in the exact regression this path is guarding against.

ecosystem-explorer/src/hooks/configuration-builder-reducer.ts
```diff
@@ -35,32 +35,56 @@ export const INITIAL_STATE: ConfigurationBuilderState = {
...
+    if (moduleVal && typeof moduleVal === "object" && !Array.isArray(moduleVal)) {
+      const modObj = { ...moduleVal } as ConfigValues;
+      if (modObj.enabled === null || modObj.enabled === undefined) {
+        delete modObj.enabled;
+        changed = true;
+      }
+      if (Object.keys(modObj).length === 0) {
+        delete inst[moduleName];
+        changed = true;
+      } else if (changed) {
```
jaydeluca: Minor readability: the `else if (changed)` write-back guard works (a modified non-empty `modObj` is always written back because `changed` was just set `true` above), but it leans on `changed` doubling as both the "something changed" flag and the per-module write guard, which reads as a bug at first glance. A one-line comment, or just unconditionally assigning `inst[moduleName] = modObj` when it's non-empty, would make this easier to follow.

ecosystem-explorer/src/features/java-agent/configuration/components/instrumentation-row.tsx
```diff
@@ -97,41 +96,29 @@ export function InstrumentationRow({
...
+        <div className="flex items-center gap-4">
+          <div className="flex flex-col items-end text-right">
+            <span
+              className={`text-sm font-bold ${isEnabled ? "text-emerald-500" : "text-muted-foreground"}`}
             >
-              <Plus className="h-3 w-3" aria-hidden="true" />
-              {t("builder.row.customize")}
-            </button>
-          )}
+              {isEnabled ? t("builder.row.enabled") : t("builder.row.disabled")}
```
jaydeluca, line 111: Nice that the enabled/disabled state is surfaced prominently here. One enhancement worth considering: alongside the status, show a count of how many configuration options each instrumentation exposes (e.g. a small "N options" badge). Right now it's hard to tell at a glance which instrumentations have customization beyond enable/disable — a count would let users spot the configurable ones without expanding each row. The number could come from the same per-module config data we already aggregate (`aggregateConfigurations(module)` / the owned config paths used by `useCustomizedModules`).

jaydeluca (commented):
**Orphan locale keys** (can't comment inline since `public/locales/` isn't in the diff): now that the "Customize" action is gone, these keys in **both** `en/java-agent.json` and `es/java-agent.json` have no remaining references — `builder.row.customize`, `builder.row.customizeAriaLabel`, `builder.row.removeTooltip`, `builder.row.enabledByDefault`, `builder.row.disabledByDefault`. Removed UI strings should have their keys deleted from every language file in the same commit. (`enabledByDefault`/`disabledByDefault` could instead be reused for the new "Default:" line — see the inline comment.)

This is looking really good, thanks! left a few other comments and theres merge conflicts but this is a nice improvement

Adii-45:
@jaydeluca Thanks for the review!

I've addressed the feedback, cleaned up the locale keys, added the requested reducer test coverage, and resolved the merge conflicts!

Would appreciate another look whenever you have a chance... Thanks!

jaydeluca (approved):
LGTM, thanks!

PR #723, opened 2026-06-22
fix: use route parameter as telemetry comparison identifier

### LINKED ISSUE #721: Telemetry Comparison view uses incorrect instrumentation identifier, preventing comparison data from loading
## Summary

The Telemetry Comparison view appears to pass an incorrect instrumentation identifier to `TelemetryComparisonSection`, which prevents the comparison hook from loading telemetry diff data.

## Problem

In `src/features/java-agent/instrumentation-detail-page.tsx`, the instrumentation route parameter is obtained via:

```tsx
const { param } = useParams<{ param: string }>()
```

However, when rendering `TelemetryComparisonSection`, the component receives:

```tsx
instrumentationName={name ?? ""}
```

The variable `name` is not declared, imported, or destructured within the component, while `param` is the route-derived instrumentation identifier already used elsewhere in the page.

This results in an invalid/empty instrumentation name being passed to the comparison flow.

## Impact

When opening the Telemetry Comparison view, the comparison hook exits early because it does not receive a valid instrumentation identifier.

As a result, telemetry comparison data is never loaded and the comparison view remains empty.

## Reproduction Steps

1. Navigate to any Java Agent instrumentation detail page.
2. Open the **Telemetry** section.
3. Enable the **Comparison** view.
4. Observe that comparison results are not loaded.

## Expected Behavior

The Telemetry Comparison view should receive the current instrumentation identifier from the route and load telemetry diff data between the selected versions.

## Actual Behavior

The comparison flow receives an invalid instrumentation identifier and returns before any comparison data is fetched.

## Suggested Fix

Pass the route parameter (`param`) to `TelemetryComparisonSection` instead of the currently referenced variable.

Example:

```tsx
instrumentationName={param ?? ""}
```

## Notes

I traced the execution path from the instrumentation detail route through the comparison component and verified that:

- The route parameter is available as `param`
- `TelemetryComparisonSection` currently receives `name`
- `name` is not declared/imported/destructured in the component
- The comparison hook depends on a valid instrumentation identifier before loading comparison data

This appears to be a small, isolated, user-facing bug that should be straightforward to fix and cover with a regression test.

Adii-45:
## What changed

This PR fixes the telemetry comparison view by passing the route-derived instrumentation identifier (`param`) to `TelemetryComparisonSection` instead of the incorrectly referenced variable.

#### `src/features/java-agent/instrumentation-detail-page.tsx`

Changed:

```tsx
instrumentationName={name ?? ""}
```

to:

```tsx
instrumentationName={param ?? ""}
```

This ensures the comparison view receives the correct instrumentation identifier from the current route.

#### `src/features/java-agent/instrumentation-detail-page.test.tsx`

Added a regression test that verifies:

- The Telemetry Comparison view receives the route parameter as `instrumentationName`
- The correct identifier is passed through when the comparison flow is enabled
- Future regressions in this data flow are detected by the test suite

## Why

The comparison flow depends on a valid `instrumentationName` to load telemetry comparison data.

The route parameter is already available via:

```tsx
const { param } = useParams<{ param: string }>();
```

However, `TelemetryComparisonSection` was receiving:

```tsx
instrumentationName={name ?? ""}
```

instead of the route-derived identifier.

As a result, the comparison component could receive an invalid/empty identifier and fail to load comparison data correctly.

Fixes #721

## How it was tested

- Verified the Telemetry Comparison view receives the expected route identifier
- Added a regression test covering the comparison flow
- Ran TypeScript type checking successfully
- Ran the affected test suite successfully

## Is this a breaking change?

No.

## Special notes for reviewers

This is a small, isolated fix that only affects the identifier passed to the telemetry comparison view. No application architecture, state management, routing behavior, or APIs were modified.

## Checklist

- [x] I wrote this PR description myself

### DIFF
--- a/ecosystem-explorer/src/features/java-agent/instrumentation-detail-page.test.tsx
+++ b/ecosystem-explorer/src/features/java-agent/instrumentation-detail-page.test.tsx
@@ -15,6 +15,7 @@
  */
 import { describe, it, expect, vi, beforeEach } from "vitest";
 import { render, screen, within, fireEvent } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 import { MemoryRouter, Route, Routes, useNavigate } from "react-router-dom";
 import { InstrumentationDetailPage } from "./instrumentation-detail-page";
 import type { InstrumentationData } from "@/types/javaagent";
@@ -28,6 +29,10 @@ vi.mock("@/components/ui/back-button", () => ({
   BackButton: () => <button>Back</button>,
 }));
 
+vi.mock("./components/telemetry-comparison/telemetry-comparison-section", () => ({
+  TelemetryComparisonSection: vi.fn(() => <div data-testid="telemetry-comparison-mock" />),
+}));
+
 vi.mock("react-router-dom", async () => {
   const actual = await vi.importActual("react-router-dom");
   return {
@@ -37,6 +42,7 @@ vi.mock("react-router-dom", async () => {
 });
 
 import { useVersions, useInstrumentation } from "@/hooks/use-javaagent-data";
+import { TelemetryComparisonSection } from "./components/telemetry-comparison/telemetry-comparison-section";
 
 const mockVersionsData = {
   versions: [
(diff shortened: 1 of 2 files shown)


Adii-45:
Hi @jaydeluca,

The fix and regression test are in place, and all checks are passing!
Would appreciate a review whenever you have a chance. Thanks!

jaydeluca (approved):
thanks!

PR #724, opened 2026-06-22
test: add nightly e2e acceptance test for declarative config builder

### LINKED ISSUE #662: End to end acceptance test for declarative config builder
I'm not yet sure how good of an idea this is, but I was thinking it would be cool to have a test that is not run as part of normal builds, but runs on a nightly basis that does the following:

- Using a browser test it navigates the site to the configuration builder and applies a handful of configurations, and then generates an output Yaml file.
- Have a test java app that is then launched with the generated configuration file, and we do some basic assertions to make sure it doesn't crash due to improper syntax
- Opens an issue if it fails

We could use https://github.com/grafana/oats as the framework for testing the agent. We can probably re-use most of the code from [this example](https://github.com/open-telemetry/opentelemetry-java-examples/tree/main/javaagent-declarative-configuration), just fed with the yaml file we generate with the explorer

Pittu-Sharma:
## What changed

Added an end-to-end acceptance testing pipeline for the declarative configuration builder. This pipeline uses a Playwright script (`generate-test-config.mjs`) to automate the browser UI and extract the generated YAML configuration. The configuration is then passed to the `smoke-test-spring-boot` container and validated natively using the `grafana/oats` testing framework via a new GitHub Action workflow (`nightly-declarative-config-test.yml`).

## Why

Fixes #662. 

The configuration builder lacked automated validation. This ensures that the YAML configurations generated by the UI are valid, successfully parseable by the OpenTelemetry Java Agent, and correctly emit telemetry to an OTLP endpoint, catching any builder regressions automatically.

## How it was tested

- Ran the `generate-test-config.mjs` Playwright script locally to verify successful extraction of the `test-config.yaml`.
- Verified the `docker-compose.yml` mounts the configuration into the `smoke-test-spring-boot` container and exports to the `lgtm` stack on port 4318.
- Validated that `oats.yaml` asserts the expected `/greeting` span when hitting the container.
- Verified passing `pre-commit` hooks locally (formatting, `ruff`, `eslint`, etc.).

## Is this a breaking change?

No.

I have added appropriate JSDoc documentation to the automation script and `.gitignore` exceptions for the runtime artifacts. The workflow triggers automatically on a schedule but can also be triggered manually via `workflow_dispatch` if you want to test it.

## Checklist


acceptance-tests/declarative-config-acceptance/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+version: "3"
+services:
+  app:
+    image: ghcr.io/open-telemetry/opentelemetry-java-instrumentation/smoke-test-spring-boot:latest
+    environment:
+      JAVA_TOOL_OPTIONS: "-javaagent:/app/opentelemetry-javaagent.jar"
+      OTEL_JAVAAGENT_CONFIGURATION_FILE: "/app/test-config.yaml"
```
jaydeluca: `OTEL_JAVAAGENT_CONFIGURATION_FILE` points the agent at a [`.properties` file](https://opentelemetry.io/docs/zero-code/java/agent/configuration/), not the declarative YAML — so the agent ignores `test-config.yaml` and falls back to env-var autoconfig (`OTEL_EXPORTER_OTLP_ENDPOINT` below makes spans flow regardless). I confirmed this by putting garbage in the config and the test still passed. Use `OTEL_CONFIG_FILE` (`otel.config.file`), which is what the generated YAML's own header tells you to use:

```suggestion
      OTEL_CONFIG_FILE: "/app/test-config.yaml"
```

acceptance-tests/declarative-config-acceptance/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+version: "3"
+services:
+  app:
+    image: ghcr.io/open-telemetry/opentelemetry-java-instrumentation/smoke-test-spring-boot:latest
```
jaydeluca: `:latest` doesn't exist for this image (`manifest unknown` at runtime) — ghcr only publishes `jdkNN-<timestamp>` tags. Pin a real one:

```suggestion
    image: ghcr.io/open-telemetry/opentelemetry-java-instrumentation/smoke-test-spring-boot:jdk17-20260623.27999598977
```

acceptance-tests/declarative-config-acceptance/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+version: "3"
```
jaydeluca: `version: "3"` is obsolete under Compose v2 and emits a warning — safe to drop.

acceptance-tests/declarative-config-acceptance/oats.yaml
```diff
@@ -0,0 +1,17 @@
...
+  # The smoke-test-spring-boot container exposes a /greeting endpoint
+  - path: /greeting
+
+expected:
+  traces:
+    - trace:
+        spans:
+          - name: "GET /greeting"
+            attributes:
+              http.route: "/greeting"
```
jaydeluca: This isn't valid OATS v2 schema — the binary panics with `field trace not found in type ExpectedTraces`. Use a `traceql` selector with `equals`/`attributes` ([OATS](https://github.com/grafana/oats)):

```suggestion
    - traceql: '{ span.http.route = "/greeting" }'
      equals: "GET /greeting"
      attributes:
        http.route: "/greeting"
```

.github/workflows/nightly-declarative-config-test.yml
```diff
@@ -0,0 +1,108 @@
...
+      - name: Generate config file
+        run: |
+          cd ecosystem-explorer
+          bun run scripts/generate-test-config.mjs ../acceptance-tests/declarative-config-acceptance/test-config.yaml
+
+      - name: Download latest OpenTelemetry Java Agent
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          AGENT_URL=$(gh release view --repo open-telemetry/opentelemetry-java-instrumentation latest --json assets -q '.assets[] | select(.name == "opentelemetry-javaagent.jar") | .url')
```
jaydeluca: `gh release view … latest` reads `latest` as a **tag**, which this repo doesn't have — the step fails with `release not found`, so `$AGENT_URL` is empty. Omit the arg (no tag = latest release):

```suggestion
          AGENT_URL=$(gh release view --repo open-telemetry/opentelemetry-java-instrumentation --json assets -q '.assets[] | select(.name == "opentelemetry-javaagent.jar") | .url')
```

(Or simpler, replace the curl entirely with `gh release download --repo … --pattern 'opentelemetry-javaagent.jar' --dir acceptance-tests/declarative-config-acceptance`.)

.github/workflows/nightly-declarative-config-test.yml
```diff
@@ -0,0 +1,108 @@
...
+          echo "Downloading latest agent from: $AGENT_URL"
+          curl -L -o acceptance-tests/declarative-config-acceptance/opentelemetry-javaagent.jar "$AGENT_URL"
+
+      - name: Run OATS acceptance tests
+        run: |
+          cd acceptance-tests/declarative-config-acceptance
+          docker run --rm -v /var/run/docker.sock:/var/run/docker.sock -v "$(pwd):/workspace" -w /workspace grafana/oats:latest test
```
jaydeluca: `grafana/oats:latest` isn't a runnable image — `docker run` fails with `pull access denied … repository does not exist`. [OATS](https://github.com/grafana/oats) is a Go binary that orchestrates docker-compose + the lgtm stack itself. Install and run it instead, pinned to a version:

```bash
go install github.com/grafana/oats@v0.7.0
oats -timeout 5m acceptance-tests/declarative-config-acceptance/
```
See [`.mise/tasks/oats-tests.sh`](https://github.com/open-telemetry/opentelemetry-java-examples/blob/main/.mise/tasks/oats-tests.sh) for a working example.

ecosystem-explorer/scripts/generate-test-config.mjs
```diff
@@ -0,0 +1,143 @@
...
+        .getByRole("button", { name: /Expand Exporter/i })
+        .first()
+        .click({ timeout: 5000 });
+      await page.getByText("otlp_http").first().click({ timeout: 5000 });
+    } catch (e) {
+      console.log(
+        "Could not toggle OTLP exporter, proceeding with default",
+        e?.message || String(e)
+      );
+    }
```
jaydeluca, line 112: Two things here:

(a) **Select schema version `1.0.0` before extracting the YAML.** The builder defaults to `1.1.0`, but released agents reject it — with the config actually loaded (see the docker-compose comment), agent 2.29.0 fails to start: `Unsupported file format '1.1'. Supported formats include 0.4, 1.0`. Drive the version selector (`#schema-version-select`) to `1.0.0` so the output is `file_format: "1.0"`. Worth a comment noting this means we don't catch the default (1.1) being ahead of released agents.

(b) **Don't swallow the exporter-toggle failure.** "Proceed with default" means a broken selector passes silently — and the default already emits `otlp_http`, so a regression here goes unnoticed. Assert the extracted YAML contains the expected exporter block before writing, and re-throw on selector failure. (Selectors also hardcode English UI text in an i18n'd app.)

.github/workflows/nightly-declarative-config-test.yml
```diff
@@ -0,0 +1,108 @@
...
+
+      - name: Install Playwright browsers
+        run: |
+          cd ecosystem-explorer
+          bunx playwright install chromium
+
+      - name: Generate config file
+        run: |
+          cd ecosystem-explorer
+          bun run scripts/generate-test-config.mjs ../acceptance-tests/declarative-config-acceptance/test-config.yaml
```
jaydeluca, line 53: The generated config lands in the same dir as `oats.yaml`. OATS scans the test dir and parses every `*.yaml` as a test case (it panics on a malformed config). Write `test-config.yaml` outside the OATS dir, or point OATS at the specific `oats.yaml`.

.github/workflows/nightly-declarative-config-test.yml
```diff
@@ -0,0 +1,108 @@
...
+  contents: read # Default minimum; the run-acceptance-test job declares its own elevated scopes
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  run-acceptance-test:
+    name: Run Acceptance Test
+    runs-on: ubuntu-latest
```
jaydeluca, line 18: Add `timeout-minutes:` to the job — with no bound a hung Playwright/OATS step runs to the 6h default. Also pin OATS (`0.7.0`) and the app image; the agent staying "latest" is correct (that's the point).

.github/workflows/nightly-declarative-config-test.yml
```diff
@@ -0,0 +1,108 @@
...
+
+      - name: Build explorer
+        run: |
+          cd ecosystem-explorer
+          bun run build
+
+      - name: Install Playwright browsers
+        run: |
+          cd ecosystem-explorer
+          bunx playwright install chromium
```
jaydeluca: Use `bunx playwright install --with-deps chromium` for reliability across runner image changes.

jaydeluca (changes requested):
I tried testing this out locally and it's not working. I worked through it to investigate why and found a few issues that need to be fixed before this can be merged. Details inline.

Reference for a working OATS setup that mirrors this use case: [opentelemetry-java-examples / javaagent-declarative-configuration/oats](https://github.com/open-telemetry/opentelemetry-java-examples/tree/main/javaagent-declarative-configuration/oats) and its runner [`.mise/tasks/oats-tests.sh`](https://github.com/open-telemetry/opentelemetry-java-examples/blob/main/.mise/tasks/oats-tests.sh).

Pittu-Sharma:
Thanks for the thorough review and for catching those issues! I've gone ahead and pushed updates to address all your feedback:

- **OATS & Docker-Compose:**
  - Removed the obsolete `version: "3"` and pinned the Spring Boot test app image to the specific `jdk17` tag instead of `latest`.
  - Switched the environment variable to use `OTEL_CONFIG_FILE` so the declarative config is properly loaded by the agent.
  - Updated the assertions in `oats.yaml` to use the correct OATS v2 `traceql` syntax instead of the invalid `trace` schema.

- **Workflow Reliability:**
  - Replaced the brittle `curl`/`grep` parsing for the agent JAR with a clean `gh release download` command.
  - Switched the OATS run step to install the runner directly (`go install github.com/grafana/oats@v0.7.0`) instead of attempting to run the non-existent docker image.
  - Added the `--with-deps` flag to the Playwright installation and added a `timeout-minutes: 15` limit to the job to prevent unbounded hangs.

- **Config Generation Script:**
  - The script now explicitly selects schema version `1.0.0` to ensure compatibility with released agents. 
  - It also no longer swallows the exporter-toggle failure; it will throw an error if the selector fails and assert that the extracted YAML actually contains the `otlp_http` exporter block.

Let me know if there's anything else you'd like me to tweak before we merge!

jaydeluca (approved):
thanks!

PR #728, opened 2026-06-24
fix(telemetry): prevent inverted comparison results for oldest version

### LINKED ISSUE #727: Telemetry Comparison inverts added/removed results when viewing the oldest instrumentation version
## Summary

The Telemetry Comparison view initializes with an incorrect default comparison range when the currently selected instrumentation version is the oldest available version.

As a result, telemetry differences are calculated in the reverse chronological direction, causing additions and removals to be displayed with inverted semantics.

## Problem

When viewing an instrumentation that has multiple released versions and selecting the oldest available version, enabling **Comparison** mode automatically creates a comparison between:

- From = latest version
- To = oldest version

instead of using a forward comparison from the currently selected version.

This causes the generated telemetry diff to be evaluated in the opposite direction from expected version progression.

## Impact

Because the comparison direction is reversed:

- Telemetry introduced in newer releases may appear as **Removed**.
- Telemetry removed in newer releases may appear as **Added**.
- The comparison results become misleading despite loading successfully.
- Users can draw incorrect conclusions about telemetry changes between releases.

The issue is especially confusing because no warning or indication is shown that the comparison is being generated in reverse.

## Reproduction Steps

1. Navigate to a Java Agent instrumentation with multiple released versions.
2. Select the oldest available version.
3. Open the **Telemetry** tab.
4. Enable **Comparison** mode.
5. Observe the automatically selected comparison range and resulting telemetry diff.

## Expected Behavior

The default comparison should follow normal version progression and present telemetry additions and removals in a direction consistent with release history.

## Actual Behavior

The comparison is initialized using a reversed version range, causing telemetry additions and removals to appear inverted relative to expected release progression.

## Notes

This issue is reproducible through normal UI interactions and does not require manual URL modification or invalid input.

Adii-45:
## Summary

Fixes #727

When the currently selected instrumentation version is the oldest available version, the telemetry comparison view initialized with a reversed comparison range!

This caused telemetry additions and removals to appear with inverted semantics!

## Changes

- Corrected the fallback `defaultFromVersion` selection for the oldest-version case
- Reused the existing same-version warning flow when no older comparison target exists
- Added regression test coverage for the oldest-version scenario

## Validation

- Added regression tests
- All tests passing
- Typecheck passing
- Lint passing
- Manually verified through the UI

### DIFF
--- a/ecosystem-explorer/src/features/java-agent/components/telemetry-comparison/telemetry-comparison-section.test.tsx
+++ b/ecosystem-explorer/src/features/java-agent/components/telemetry-comparison/telemetry-comparison-section.test.tsx
@@ -50,6 +50,7 @@ const versions: VersionInfo[] = [
 
 describe("TelemetryComparisonSection — defaultFromVersion", () => {
   beforeEach(() => {
+    vi.clearAllMocks();
     vi.mocked(useTelemetryComparisonModule.useTelemetryComparison).mockReturnValue(
       noopHookReturn
     );
--- a/ecosystem-explorer/src/features/java-agent/components/telemetry-comparison/telemetry-comparison-section.test.tsx
+++ b/ecosystem-explorer/src/features/java-agent/components/telemetry-comparison/telemetry-comparison-section.test.tsx
@@ -51,9 +51,7 @@ const versions: VersionInfo[] = [
 describe("TelemetryComparisonSection — defaultFromVersion", () => {
   beforeEach(() => {
     vi.clearAllMocks();
-    vi.mocked(useTelemetryComparisonModule.useTelemetryComparison).mockReturnValue(
-      noopHookReturn
-    );
+    vi.mocked(useTelemetryComparisonModule.useTelemetryComparison).mockReturnValue(noopHookReturn);
   });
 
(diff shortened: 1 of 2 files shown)


Adii-45:
@jaydeluca

I've addressed the feedback and all checks are passing now..!
Would appreciate a review whenever you have the opportunity.. Thank you!

Adii-45:
@jaydeluca
Just following up on this PR... Would appreciate a review whenever you have a chance. Thank you!

jaydeluca:
@Adii-45 i am in the process of reviewing this right now. 

One note for future contributions: please avoid repeatedly pinging maintainers for review or updates. Most of the project is maintained by volunteers, and repeated nudges can create extra noise and pressure rather than helping move things faster. This PR hadn't even been open 24 hours and you've already pinged me twice about it.

It’s completely fine to follow up occasionally if a PR has gone stale, but in general the best approach is to wait for maintainers to review when they have capacity. Thanks for understanding and for contributing.

jaydeluca (approved):

Adii-45:
Thanks for the clarification, and apologies for the repeated pings. I'll keep that in mind for future contributions.. Appreciate the review!

PR #735, opened 2026-06-27
fix(java-agent): collapse cross-condition telemetry moves into a single changed entry

### LINKED ISSUE #732: Java Agent Release Comparison add/removed bug
### Expected behavior

When looking at the Telemetry Added/Removed, I should only see deltas between the releases.

### Actual behavior

when looking at [2.27.0 compared to 2.28.1](https://explorer.opentelemetry.io/java-agent/releases?from=2.27.0&to=2.28.1), i see the same internal span being marked as both removed and added for JAXRS

Image

When looking at the [individual page](https://explorer.opentelemetry.io/java-agent/instrumentation/jaxrs-1.0) it shows that the reason is that the backing information was corrected to show that the internal span is no longer emited by default, but only when the config option is selected

Image

Image

For these comparison diffs, we need some kind of middleware logic that allows us to override and fix discrepancies that are caused by an issue with the data itself.

For example, the new behavior of 2.28.1 is actually something that should be backfilled for all previous instrumentations too. we need a way to handle that, whether we do it in the explorer-db-builder or have something in the comparison logic.

The comparison logic probably needs a way to differentiate added/removed in a way that is aware of the `when` condition too

Adii-45:
## Summary

Fixes #732

This PR fixes an issue in the Java Agent release comparison where telemetry that only moved between `when` conditions was rendered as two separate entries (Removed + Added) instead of a single Changed entry.

For example, the JAX-RS 1.x INTERNAL span moved from:

- `default` (v2.27.0)
- `otel.instrumentation.common.experimental.controller-telemetry.enabled=true` (v2.28.1)

Previously this appeared as two separate cards even though the span itself was unchanged.

## What changed

- Added cross-condition span and metric deduplication during release comparison.
- Detect structurally identical Added/Removed pairs and collapse them into a single `changed` entry.
- Preserve the destination `when` condition so it can be surfaced in the UI.
- Added a contextual "Now only emitted when ..." message for condition-only changes.
- Left genuine additions, removals, and content changes unchanged.
- Added regression tests covering:
  - cross-condition span moves
  - cross-condition metric moves
  - genuine additions
  - genuine removals
  - real content changes

## Result

Before:

- Removed
- Added

After:

- Changed
- "Now only emitted when `<condition>`"

This better represents telemetry that has moved between conditions without implying that it was removed and re-added.

### DIFF
--- a/ecosystem-explorer/public/locales/en/java-agent.json
+++ b/ecosystem-explorer/public/locales/en/java-agent.json
@@ -291,6 +291,9 @@
     "metricRemoved": "This metric is no longer emitted in the comparison version.",
     "spanRemoved": "This span is no longer emitted in the comparison version.",
     "spanAdded": "This span is newly emitted in the comparison version.",
+    "whenChanged": {
+      "label": "Now only emitted when"
+    },
     "moduleAddedTooltip": "This module was added in the new version",
     "moduleRemovedTooltip": "This module was removed in the new version",
     "moduleChangedTooltip": "This module has telemetry or configuration changes",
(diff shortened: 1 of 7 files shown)


jaydeluca (approved):
nice, thanks!

PR #741, opened 2026-06-29
feat(v1): add a language switcher to the v1 navbar

vitorvasc:
## What changed

The v1 navbar gets a language switcher (Radix dropdown) next to the theme
toggle, reusing the same `td-*` chrome that mirrors opentelemetry.io's Bootstrap
dropdown. A new `src/i18n/languages.ts` holds the supported locales and
namespaces as one source of truth, so the i18n init config, the legacy header
`<select>`, and the new v1 switcher all derive from a single list and stay in
sync.

## Why

Part of the v1 redesign (#84). The v1 navbar shipped a theme toggle but no way
to switch language, while the legacy header already had one.

## How it was tested

- `bun run typecheck`: clean
- `bun run test -t "LanguageToggle"`: 3 tests pass; the nav-bar test also asserts the toggle renders

## Is this a breaking change?

No

- The v1 chrome stays English-only by design for now (Docs link, theme and language labels). Option labels use endonyms ("English", "Español") so they read the same in any locale and stay out of the translation files; wiring the rest through i18next is a separate pass.
- `SUPPORTED_LANGUAGES` and `NAMESPACES` mirror the on-disk `public/locales/en` and `es` files. No locale keys were added or removed.

/cc @eruizdechavez 

## Checklist

### DIFF
--- a/ecosystem-explorer/src/components/layout/header.tsx
+++ b/ecosystem-explorer/src/components/layout/header.tsx
@@ -19,6 +19,7 @@ import { Menu, X, Sun, Moon, Monitor, ChevronDown } from "lucide-react";
 import { Link, useLocation } from "react-router-dom";
 import { useTranslation } from "react-i18next";
 import { OtelLogo } from "@/components/icons/otel-logo";
+import { LANGUAGES } from "@/i18n/languages";
 import { useTheme, type ThemeMode } from "@/theme-context";
 
 const NAV_ITEMS = [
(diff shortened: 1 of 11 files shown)


maryliag (approved):

eruizdechavez (approved):
Looks nice! 

If the new version is not going to support multiple languages initially, have you considered wrapping the language selector in a feature flag again? You can use [my initial i18n pr](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/649) as reference.

jaydeluca (approved):

PR #744, opened 2026-06-29
test(i18n): added locale key parity test

### LINKED ISSUE #736: add a test for en/es locale key sync
## Component / area:
ecosystem-explorer - i18n / tests

## Problem this solves:
Right now keeping the en and es locale json files in sync is basically just a manual review step. AGENTS.md says reviewers should check that every key added to en also gets added to es (and that removed keys get cleaned up from both), but there's nothing actually enforcing that. So if someone adds a key and forgets es, or deletes a key from en but leaves it behind in es, nothing catches it, it just sits there until someone happens to notice.

## Proposed solution:
Add a test that goes through every file under public/locales, flattens out the keys for each namespace, and checks that en and es have the exact same set of keys. Should be pretty simple, just reading json and comparing key sets, no need to touch any app code.

## Acceptance criteria:
fails if a key exists in en but not es, or in es but not en, for any namespace
passes against the current locale files (they're in sync right now, I checked)
doesn't need to be hardcoded to en/es specifically so it still works if another language gets added later

MeloveGupta:
## What changed

added a test that checks the en and es locale json files actually have the same keys. basically if someone adds a key to en and forgets to add it to es (or deletes one and leaves it behind in the other), this will catch it instead of nobody noticing.

## Why

Fixes #736 
right now keeping en/es in sync is just a manual review thing, AGENTS.md tells reviewers to check it by hand but nothing actually enforces it, so stuff can slip through.

## How it was tested

ran ```bun run test src/i18n/locale-parity.test.ts```, all 9 tests pass. also tested that it actually catches drift, deleted a key from one of the es files on purpose, reran it, got a clean failure showing exactly which key was missing, then put the file back and reran to confirm it goes green again. ran the full test suite too, didn't break anything else.

## Is this a breaking change?

No

just a new test file, doesn't touch any app code. it reads the language folders dynamically off public/locales/ instead of hardcoding en/es so it'll keep working automatically if a third language gets added later.

## Checklist

### DIFF
--- a/ecosystem-explorer/src/i18n/locale-parity.test.ts
+++ b/ecosystem-explorer/src/i18n/locale-parity.test.ts
@@ -1,0 +1,89 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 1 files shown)


jaydeluca (approved):
thanks!

PR #748, opened 2026-06-30
fix(java-agent): add telemetry when-condition corrections

### LINKED ISSUE #732: Java Agent Release Comparison add/removed bug

Adii-45:
## Summary

Closes #732

This PR implements the second telemetry correction pass described in #732!

Historically, some generated Java agent registry data placed telemetry under incorrect `when` conditions. This change introduces a dedicated correction pass that runs during inventory generation and relocates telemetry to the correct `when` block before metadata backfill and explorer output generation.

## Changes

- Add a dedicated telemetry `when` correction module
- Add a declarative YAML correction seed
- Execute the correction pass during `load_and_augment_inventory()`
- Apply ignore-condition folding before telemetry corrections
- Support version-gated corrections where required
- Add unit tests covering:
  - ignore-condition folding
  - version guards
  - telemetry relocation
  - block cleanup
  - all-version corrections
  - metric corrections

## Testing

- ✅ Added unit tests for the new correction pass
- ✅ Verified correction logic locally
- ✅ Existing test suite passes

## Notes

Per maintainer guidance, this PR intentionally excludes generated explorer output (`public/data/...`). Generated artifacts will be produced and verified separately after merge.


jaydeluca (approved):
thanks! i found one other small thing while testing related to the service peer mapping config and pushed that up, but everything looks good on my tests locally

PR #755, opened 2026-06-30
fix(workflows): add js instrumentation watcher to nightly registry update

### LINKED ISSUE #752: Nightly registry update workflow does not run js-instrumentation-watcher
## Summary

The nightly registry update workflow currently does not execute the JavaScript instrumentation watcher, which means JavaScript registry data is never refreshed after the initial import.

## Current behavior

The `nightly-registry-update.yml` workflow runs the other registry synchronization jobs (collector, Java agent, .NET, configuration, etc.), but there is no step that executes the `js-instrumentation-watcher`.

The watcher itself already exists in the repository and exposes a runnable entry point, but it is never invoked by the nightly workflow.

As a result, newly released JavaScript instrumentation packages are not reflected in the Explorer unless the registry is updated manually.

## Evidence

Verified locally:

- `ecosystem-automation/js-instrumentation-watcher` exists.
- The watcher exposes a CLI entry point via its `pyproject.toml`.
- `.github/workflows/nightly-registry-update.yml` does not invoke `js-instrumentation-watcher`.
- Other registry watchers are executed from the same workflow.

## Expected behavior

The nightly registry update workflow should also execute the JavaScript instrumentation watcher so that the JavaScript registry stays synchronized with upstream releases.

## Possible fix

Add a step to invoke the JavaScript watcher in `nightly-registry-update.yml`, similar to the existing registry synchronization jobs, and expose its result for the existing notification/reporting flow if needed.

Adii-45:
## Summary

This PR updates the nightly registry update workflow to execute the JavaScript instrumentation watcher alongside the existing registry synchronization jobs.

### Changes

- Add a `js-instrumentation-watcher` step to the `synchronize-inventory` job.
- Expose the watcher result as a workflow output.
- Add a notification job for JavaScript watcher failures following the existing pattern used by the other registry watchers.

This keeps the JavaScript registry synchronized with upstream releases and aligns its workflow with the existing collector, Java agent, .NET, and configuration watchers.

Closes #752

### DIFF
--- a/.github/workflows/nightly-registry-update.yml
+++ b/.github/workflows/nightly-registry-update.yml
@@ -30,6 +30,7 @@ jobs:
       collector_result: ${{ steps.collector_watcher.outcome }}
       java_result: ${{ steps.java_instrumentation_watcher.outcome }}
       dotnet_result: ${{ steps.dotnet_instrumentation_watcher.outcome }}
+      js_result: ${{ steps.js_instrumentation_watcher.outcome }}
       configuration_result: ${{ steps.configuration_watcher.outcome }}
     steps:
       - name: Detect repository type (if on a fork we use different git config)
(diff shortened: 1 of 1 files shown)


jaydeluca (approved):
thanks!

PR #756, opened 2026-07-01
feat(about): add ecosystem coverage statistics

### LINKED ISSUE #754: Add stats around ecosystem coverage to about page
### Component / area

The [About](https://explorer.opentelemetry.io/about) page should have a new section "Ecosystems"

### Problem this solves

As a user exploring the ecosystem explorer, I would like to know how many versions of the java agent are mapped, and how many instrumentation libraries. For example: 5 versions of the java agent, and the sum of the distinct set of the libraries in each.

### Proposed solution

Add a new step the explorer-db-builder that produces a new data artifact, something like "ecosystem-stats.json" that has a handful of stats that are calculated during build time.

For the java agent and collector ecosystems, count how many different versions are supported, and the sum of the unique set of libraries/components tracked in any version. Write it to the summary file, and then reference that file in the about page.

### Acceptance criteria

When viewing the about page, I see somethign to the effect of:

Ecosystems:

Java Agent:
4 Versions
250 instrumentation libraries

Collector:
5 Versions
203 Components

### Additional context

_No response_

Adii-45:
## Summary

This PR adds ecosystem coverage statistics to the About page by generating build-time metadata for each ecosystem and rendering it in a new **"Ecosystems"** section.

The statistics are computed during the existing explorer-db-builder pipeline, written as a new `ecosystem-stats.json` artifact, and consumed by the frontend instead of using hardcoded values.

Fixes #754!

---

### Builder

- Added generation of a new `ecosystem-stats.json` artifact for each ecosystem.
- Compute:
  - Number of supported versions.
  - Number of unique Java instrumentation libraries.
  - Number of unique Collector components.
- Integrated artifact generation into the existing Java Agent and Collector build pipelines.
- Added unit and integration tests covering the new builder logic.

### Frontend

- Added a new **Ecosystems** section to the About page.
- Fetch ecosystem statistics from the generated build artifacts.
- Display:
  - Java Agent
    - Version count
    - Instrumentation library count
  - Collector
    - Version count
    - Component count
- Added loading/error handling following existing data-fetching patterns.
- Added i18n entries and frontend tests for the new section.

---

## Result

The About page now displays live build-generated ecosystem coverage information instead of hardcoded values.

Example:

- **Java Agent**
  - 5 Versions
  - 263 Instrumentation Libraries

- **Collector**
  - 7 Versions
  - 271 Components

(The exact numbers depend on the current registry contents.)

---

## Testing

- Added builder unit tests.
- Added frontend tests.
- Verified generated `ecosystem-stats.json` artifacts.
- Verified the About page renders the generated statistics correctly.
- Verified translations render correctly.
- Ran the existing project test suite successfully.

---

### About page – Ecosystems section

The About page now includes a new **Ecosystems** section displaying build-generated coverage statistics for the Java Agent and Collector.

About page with Ecosystems section

### DIFF
--- a/ecosystem-automation/explorer-db-builder/src/explorer_db_builder/ecosystem_stats.py
+++ b/ecosystem-automation/explorer-db-builder/src/explorer_db_builder/ecosystem_stats.py
@@ -22,7 +22,7 @@ reflects the latest version, so these unions must run over every loaded version 
 from typing import Any
 
 
-def count_unique_library_names(inventories: list[dict[str, Any]]) -> int:
+def count_unique_java_library_names(inventories: list[dict[str, Any]]) -> int:
     """Count unique instrumentation names across all versions.
 
     Libraries and custom instrumentations are counted together, matching how every other
@@ -45,7 +45,7 @@ def count_unique_library_names(inventories: list[dict[str, Any]]) -> int:
     return len(names)
 
 
-def count_unique_component_ids(components_by_version: list[list[dict[str, Any]]]) -> int:
+def count_unique_collector_component_ids(components_by_version: list[list[dict[str, Any]]]) -> int:
     """Count unique collector component ids across all versions.
 
     Args:
--- a/ecosystem-automation/explorer-db-builder/src/explorer_db_builder/ecosystem_stats.py
+++ b/ecosystem-automation/explorer-db-builder/src/explorer_db_builder/ecosystem_stats.py
@@ -1,0 +1,64 @@
+# Copyright The OpenTelemetry Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""Aggregate cross-version counts for the About page's ecosystem-stats.json.
+
+Produces the counts written to ecosystem-stats.json during the javaagent and collector
+builds. Pure (no I/O); the respective DatabaseWriter writes the file. index.json only
+reflects the latest version, so these unions must run over every loaded version instead.
+"""
+
+from typing import Any
+
+
+def count_unique_library_names(inventories: list[dict[str, Any]]) -> int:
+    """Count unique instrumentation names across all versions.
+
+    Libraries and custom instrumentations are counted together, matching how every other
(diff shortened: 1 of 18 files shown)


ecosystem-automation/explorer-db-builder/src/explorer_db_builder/ecosystem_stats.py
```diff
@@ -0,0 +1,64 @@
...
+
+Produces the counts written to ecosystem-stats.json during the javaagent and collector
+builds. Pure (no I/O); the respective DatabaseWriter writes the file. index.json only
+reflects the latest version, so these unions must run over every loaded version instead.
+"""
+
+from typing import Any
+
+
+def count_unique_library_names(inventories: list[dict[str, Any]]) -> int:
```
jaydeluca: Since the shape of these inputs is dependent on the ecosystem, we should make the methods more explicit as to which ones they are for
```suggestion
def count_unique_java_library_names(inventories: list[dict[str, Any]]) -> int:
```

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/ecosystem_stats.py
```diff
@@ -0,0 +1,64 @@
...
+    for inventory in inventories:
+        for key in ("libraries", "custom"):
+            for item in inventory.get(key) or []:
+                name = item.get("name")
+                if name:
+                    names.add(name)
+    return len(names)
+
+
+def count_unique_component_ids(components_by_version: list[list[dict[str, Any]]]) -> int:
```
jaydeluca:
```suggestion
def count_unique_collector_component_ids(components_by_version: list[list[dict[str, Any]]]) -> int:
```

ecosystem-explorer/src/hooks/use-ecosystem-stats.ts
```diff
@@ -0,0 +1,88 @@
...
+  collector: CollectorStatsShape;
+}
+
+export interface UseEcosystemStatsOptions {
+  javaAgentStatsUrl?: string;
+  collectorStatsUrl?: string;
+}
+
+async function fetchJson<T>(url: string): Promise<T> {
+  const res = await fetch(url);
```
jaydeluca: everywhere else we do `fetchWithCache` and use `resolveDataPath` for the URLs. I think it's ok in this case not to cache the data since it's so small, so we don't need the cache part, but can you add a comment that it is intentional? Also, we should update the paths to use `resolveDataPath`. 

```suggestion
  // Intentionally not using cache for this data
  const res = await fetch(url);
```

jaydeluca (commented):
looks good! just a couple small changes needed

Adii-45:
Thanks for the review, @jaydeluca!

I’ve addressed all the requested changes and updated the tests accordingly. All checks are passing now..!

Please let me know if there’s anything else that should be adjusted. Thanks!

jaydeluca (approved):
nice, thanks!

PR #763, opened 2026-07-02
feat/adding all buttons for instrumentations

### LINKED ISSUE #631: [Config builder] Add all button for instrumentations
As someone using the declarative configuration builder, I would like a button on the instrumentation tab for "add all" which adds all configuration options and default values to the YAML output for the instrumentation modules

ThyTran1402:
## What changed

Adds an **"Add all configs"** button to the Instrumentation tab of the declarative Configuration Builder. It writes every instrumentation-module configuration option (across all scopes which general. *`, `java.common.*`, and module-owned) with its default value into the YAML output in a single action. 
The add is **merge-safe**: it only fills configs that aren't already set, so any values the user has customized are preserved. The click is gated behind a confirmation dialog showing the count, and it always adds all modules regardless of the active search/status filter.

## Why

Resolved: #631 

Previously the only way to populate instrumentation config options was to expand each module and click "Override" on individual fields. For users who want the full set of options and defaults as a starting point, that is tedious across ~200+ modules. This adds a one-click bulk action while guaranteeing it never clobbers existing edits.

## How it was tested

- New unit tests: reducer `MERGE_DEFAULTS` (merge-safety, no-op identity, shared-parent paths, list/map preservation), the `buildInstrumentationDefaultEntries` helper (scope filtering, cross-module dedupe, `parseDefault` mapping), the `mergeDefaults` hook action, and the `InstrumentationBrowser` button (label distinctness, confirm accept/cancel, filter-independence, disabled-when-empty).
- New integration test: switching to the Instrumentation tab, clicking "Add all configs", and asserting the `instrumentation/development:` section appears in the YAML preview.
- Full suite green locally: `bun run test` (782 passed) and `bun run test:integration` (34 passed), plus `bun run typecheck` and `bun run lint` clean.

## Is this a breaking change?
No

## Checklist


jaydeluca:
@ThyTran1402 just noting that there are some failing tests. I'm going to put this into draft just for housekeeping purposes, feel free to re-open when things are green. Thanks!

ThyTran1402:
Hi @jaydeluca,

I fixed all of the CI errors. Can you please take a look when you have time?

Thank you.

lucacavenaghi97:
Hi @jaydeluca ! I'll be able to do the review in the next few days

ThyTran1402:
Hi @lucacavenaghi97,

I think the PR is ready for your review.

Thank you!

ecosystem-explorer/src/lib/instrumentation-default-entries.ts
```diff
@@ -0,0 +1,62 @@
...
+      if (!opts.includeScopes.includes(cfg.scope)) continue;
+      // aggregateConfigurations already deduped per module by declarative_name;
+      // dedupe again across modules so shared general.*/java.common.* configs
+      // land once. Keyed by serialized path since path uniquely maps 1:1 to
+      // declarative_name.
+      const key = cfg.path.join(".");
+      if (byPathKey.has(key)) continue;
+      byPathKey.set(key, {
+        path: cfg.path,
+        value: parseDefault(cfg.entry.type, cfg.entry.default),
```
lucacavenaghi97: The `structured_list` case is missing here. The field does `isStructuredList ? [] : parseDefault(...)` (`instrumentation-config-field.tsx:365`), so `java.common.service_peer_mapping` (map + structured_list) ends up as `{}` instead of `[]`. On 2.30.0 that config is in 43 modules, and after the click all 43 show the "imported value not a map" warning. It goes away as soon as the user adds an entry, but it's not great to see it appear right after clicking the button.

I'd pull that branch into a helper called from both places, so they can't drift again.

ecosystem-explorer/src/hooks/configuration-builder-reducer.ts
```diff
@@ -225,6 +225,57 @@ export function configurationBuilderReducer(
...
+      for (const { path, value } of action.entries) {
+        if (getByPath(values, path) !== undefined) continue;
+        values = setByPath(values, path, value);
+        changed = true;
+      }
+      if (!changed) return state;
+      return { ...state, values, isDirty: true };
+    }
+
+    case "SET_OVERRIDE": {
```
lucacavenaghi97: Nothing dispatches `SET_OVERRIDE`, and nothing uses `setOverride` from the hook (the one in `section-expansion-context` is a different thing). If it's for another issue I'd keep it out of here.

Also note the shape isn't the one the rest of the builder uses (`instrumentation.<module>.enabled` boolean, not two arrays), and line 263 deletes the whole `distribution` key, not just the instrumentation part.

ecosystem-explorer/src/features/java-agent/configuration/components/instrumentation-browser.tsx
```diff
@@ -118,6 +132,16 @@ export function InstrumentationBrowser({
     [setCustomization]
   );
 
+  const handleAddAll = useCallback(() => {
+    if (defaultEntries.length === 0) return;
+    const ok = window.confirm(
```
lucacavenaghi97, line 138: Hardcoded string, this needs to go through `t()` with the key added to `es` too (rule is in `ecosystem-explorer/AGENTS.md`).

On the count: 24 of the 141 have an empty default, `stripEmpties` drops them and only 117 make it into the YAML. Writing nothing when the default is empty is correct, but the field still marks itself as customized so you get a Reset button with no output. I'd skip those upfront.

ecosystem-explorer/src/features/java-agent/configuration/components/instrumentation-browser.tsx
```diff
@@ -132,6 +156,15 @@ export function InstrumentationBrowser({
...
           ) : null}
         </h3>
+        <button
+          type="button"
+          onClick={handleAddAll}
+          disabled={defaultEntries.length === 0}
+          className="border-border/60 bg-card text-foreground hover:bg-card/80 flex items-center gap-1 rounded-md border px-3 py-1 text-xs disabled:cursor-not-allowed disabled:opacity-50"
+        >
+          <CopyPlus className="h-3 w-3" aria-hidden="true" />
+          Add all configs
```
lucacavenaghi97: Same for the label. And this is `HeaderActionButton` (`preview-card.tsx:42`) rewritten with slightly different classes: as it stands it ends up without `cursor-pointer` and without the app focus ring (I checked, it gets the browser default outline). I'd export that one instead of duplicating it.

Last thing: on the Instrumentation tab this button sits next to the preview's "Add all", which does something else. "Add all instrumentation configs" would be clearer.

lucacavenaghi97 (commented):
Tried this locally on agent 2.30.0. It does what #631 asks: the merge doesn't overwrite anything, common and general are in, experimental options are in, and it ignores the active filter. 

Three things I'd fix, commented below. Then a separate note: if you click "Add all configs" on 2.30.0 and then switch the agent selector to 2.26.1, the YAML keeps 30 options that version doesn't know about (graphql, cassandra, ...). That comes from `PRUNE_INSTRUMENTATIONS` only pruning `distribution.javaagent.instrumentation`, so it's not on this PR, but with one click it becomes the normal case. I'll open an issue for it.

ecosystem-explorer/src/lib/instrumentation-default-entries.ts
ThyTran1402: Yup. Thank you for your feedback. I addressed it and added more tests.

ecosystem-explorer/src/hooks/configuration-builder-reducer.ts
ThyTran1402: Hmmm I believe SET_OVERRIDE already exists on main by another PR, and on main it's live: `instrumentation-browser.tsx` calls `setOverride` from `useConfigurationBuilder`.  This PR replaced that mechanism with `setCustomization`, which dispatches plain SET_VALUE writing `instrumentation.<module>.enabled` `booleans`, and rewired the browser to use it. That orphaned `setOverride/SET_OVERRIDE`. 

I also dropped all of distribution and address the shape mismatch.

ecosystem-explorer/src/features/java-agent/configuration/components/instrumentation-browser.tsx
ThyTran1402: Addressed

ecosystem-explorer/src/features/java-agent/configuration/components/instrumentation-browser.tsx
ThyTran1402: Addressed with structured_list options are now intentionally excluded from bulk-add, while individual Override still seeds them with `[]` for editing.

lucacavenaghi97:
LGTM, thanks for your work! @ThyTran1402

lucacavenaghi97 (approved):

jaydeluca (approved):

PR #766, opened 2026-07-02
fix(javaagent): prevent false "No README found" flash on initial render

### LINKED ISSUE #764: Standalone Library briefly shows a false “No README found” state before the README begins loading
## Component / area

Java Agent → Instrumentation Detail Page → Standalone Library

---

## Problem this solves

When opening the **Standalone Library** tab, the UI can briefly render the **"No README found"** error state before transitioning into the loading spinner, even though the README ultimately loads successfully.

This creates a short but incorrect error flash during the initial render and is inconsistent with the loading behavior used by the other data-fetching hooks in the same feature.

---

## Evidence

The `useLibraryReadme` hook initializes its state with:

```ts
{
  data: null,
  loading: false,
  error: null,
}
```

Unlike the other hooks in the same file (`useVersions`, `useInstrumentation`, and `useInstrumentation`), which begin with `loading: true`.

Because `loading` is initially `false`, the first render reaches the consumer before the effect starts the fetch and updates the loading state.

The `StandaloneLibraryTab` component checks the loading state before rendering the markdown content. During this first render it can therefore enter the error/"not found" branch before the loading spinner appears.

After the effect runs, the hook updates `loading` to `true`, the spinner renders, and finally the README is displayed once the request completes successfully.

---

## Why this appears to be unintended

- Other data-fetching hooks in the same file initialize with `loading: true`.
- The surrounding hooks follow a consistent loading → success/error lifecycle.
- No TODOs, comments, planning documents, or existing issues indicate that this brief error flash is intentional.
- The behavior appears to be an inconsistency in the initial hook state rather than a deliberate UX decision.

---

## Suggested fix

Initialize `useLibraryReadme` with `loading: true` so the first render immediately reflects the loading state, matching the behavior of the other hooks in the file and avoiding the transient false "No README found" state.

This keeps the loading lifecycle consistent across the related hooks and removes the unnecessary error flash.

Adii-45:
## What changed

This PR initializes `useLibraryReadme` with `loading: true` instead of `loading: false`.

It also adds a regression test covering the hook's initial loading state and the normal successful loading path.

## Why

The `StandaloneLibraryTab` checks the hook's loading state before rendering the markdown content.

Because `useLibraryReadme` previously initialized with `loading: false`, the first render could briefly enter the "No README found" state before the effect started the fetch and switched the hook into the loading state.

Initializing the hook in the loading state keeps its lifecycle consistent with the other data-fetching hooks in the same file and avoids the transient incorrect UI.

Fixes #764!

## How it was tested

- Added regression tests for `useLibraryReadme`
  - verifies the hook starts in the loading state
  - verifies the normal successful loading path still works
- Ran the new tests successfully.

## Is this a breaking change?

No.

## Notes for reviewers

This intentionally keeps the change minimal:
- one-line initialization change
- regression tests only
- no behavior changes beyond preventing the transient incorrect render

### DIFF
--- a/ecosystem-explorer/src/hooks/use-javaagent-data.test.ts
+++ b/ecosystem-explorer/src/hooks/use-javaagent-data.test.ts
@@ -1,0 +1,57 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { renderHook, waitFor } from "@testing-library/react";
+import { useLibraryReadme } from "./use-javaagent-data";
+
+vi.mock("@/lib/api/javaagent-data", () => ({
+  loadLibraryReadme: vi.fn(),
+}));
+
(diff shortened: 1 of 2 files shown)


jaydeluca (approved):
thanks!

PR #782, opened 2026-07-06
fix(collector): prevent infinite loading on Collector detail page

### LINKED ISSUE #774: Collector detail page remains stuck on loading when versions fetch fails and no ?version= is present
## Component/Area

Collector feature → Collector detail page (`CollectorDetailPage`) / version loading

## Expected behavior

If the Collector versions index fails to load, the Collector detail page should surface an error state (similar to other Collector pages) instead of remaining indefinitely in a loading state!

## Actual behavior

When the Collector versions request fails and the URL does not include a `?version=` query parameter, the Collector detail page never leaves its loading state.

The page remains on the loading spinner indefinitely and no error UI or recovery path is shown.

## Steps to reproduce

1. Open a Collector detail page without a `?version=` query parameter (for example, navigate directly to a component detail URL).
2. Simulate a failure while loading the Collector versions index (e.g. block the request or force it to fail).
3. Observe that the page remains on the loading spinner indefinitely.
4. Compare this with the Collector list page under the same failure, which correctly renders an error state.

## User impact

Users encountering a versions-fetch failure on a Collector detail page are left with a permanent loading screen and no indication of what went wrong or how to recover!

Since the Collector detail route is now publicly available following the recent Collector feature rollout, this affects a production user path rather than a gated feature.

## Notes

During investigation I verified that:

- the relevant loading logic remains unchanged
- the Collector detail page still has no dedicated regression test covering this scenario
- similar Collector and Java Agent pages already surface an error state for equivalent failures

Adii-45:
## What changed

This PR fixes the Collector detail page getting stuck in an infinite loading state when the versions fetch fails and the URL does not contain a `?version=` query parameter.

Specifically, it:

- reads the existing `versionsError` returned by `useCollectorVersions()`
- stops waiting for a version once the versions request has failed
- routes the page into the existing error UI instead of remaining in the loading state
- adds regression tests covering this scenario while preserving the existing success and loading paths

Fixes #774

---

## Why

The Collector detail page previously ignored the `error` returned from `useCollectorVersions()`.

When the versions request failed and there was no explicit `?version=` in the URL, the page could never resolve a version, causing it to remain on the loading spinner indefinitely without any recovery UI.

Other Collector pages and the equivalent Java Agent page already surface this failure correctly. This change aligns the Collector detail page with the existing error-handling behavior used elsewhere.

---

## How it was tested

- Added regression tests covering versions-fetch failures
- Verified existing loading behavior remains unchanged
- Verified successful version resolution still works
- Verified existing component-fetch error handling is preserved
- Ran targeted Collector tests
- Ran `tsc --noEmit`
- Ran ESLint

---

## Breaking change?

No

---

## Notes for reviewers

This change is intentionally scoped to the Collector detail page.

No hook or API changes were required because `useCollectorVersions()` already exposed the required error state. The page simply wasn't consuming it..!

The implementation reuses the existing error rendering path, keeping behavior consistent with other Collector pages while adding regression coverage for the previously untested scenario!

### DIFF
--- a/ecosystem-explorer/src/features/collector/collector-detail-page.test.tsx
+++ b/ecosystem-explorer/src/features/collector/collector-detail-page.test.tsx
@@ -1,0 +1,164 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { render, screen } from "@testing-library/react";
+import { MemoryRouter, Route, Routes } from "react-router-dom";
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { CollectorDetailPage } from "./collector-detail-page";
+
+vi.mock("@/hooks/use-collector-data", () => ({
+  useCollectorVersions: vi.fn(),
+  useCollectorComponent: vi.fn(),
+}));
+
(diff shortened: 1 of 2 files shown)


jaydeluca (approved):
thanks!

Issue #762, opened 2026-07-02
Collector Readme rendering


jaydeluca:
### Component / area

Collector component detail pages

### Problem this solves

If there is additional context or information included in the readme, it becomes discoverable.

### Proposed solution

We will first need to update our collector watcher to pull in the readme docs into the registry. Then we will need to incorporate it into the detail pages, rendering the markdown like we do for the java pages. We need to also incorporate a step with the explorer-db-builder to make the data accessible in the explorer.

We might also need to see if we should do any processing, for example stripping out duplicative information or headers. That piece will need some investigation if it's worth it / feasible to do in a consistent manner.

### Acceptance criteria

If a collector component has a readme, we should render it, similar to how we are rendering standalone java instrumentation library readmes (like in https://explorer.opentelemetry.io/java-agent/instrumentation/apache-dbcp-2.0).

MeloveGupta:
Hey @jaydeluca , I would love to work on this...
Could you assign it to me???

MeloveGupta:
Hey @jaydeluca ,
Yesterday I went through how the java standalone library readme thing works end to end since the issue says it has to do something similar. Here is what I found out:

- ```java-instrumentation-watcher``` fetches readmes remotely off github's tree api (```readme_extractor.py```), regex matches on ```instrumentation/*/library/README.md```
- gets stored content-addressed (```name-hash.md```) by ```JavaagentInventoryManager``` in ```watcher_common```, which also rebuilds a name -> hash map by scanning that dir back
- db-builder reads that map, writes the final markdown file (```DatabaseWriter.write_markdown```) and stamps a markdown_hash field onto the instrumentation record
- frontend only shows the tab if ```has_standalone_library``` and ```markdown_hash``` are both there, lazy loads the actual content through a hook when you open the tab. 

But collector doesn't line up as simple. There are two things:

- ```collector-watcher``` doesn't hit a remote api at all, it clones the repos locally and scans the filesystem (```component_scanner.py```), so grabbing the readme is actually simpler here, just check if ```README.md``` exists next to the component
- bigger thing - collector_watcher's ```InventoryManager``` doesn't extend the shared ```BaseInventoryManager``` that java's readme storage logic lives on top of. Because collector's inventory is keyed by (distribution, version), not just version, since there's core + contrib. All of java's readme storage methods assume a single version key.

So I don't think I can just inherit the readme storage logic the way it's built right now. 2 ways that I thought to move ahead:
[...]

PR #785, opened 2026-07-08
feat(db-builder): publish collector component READMEs and stamp markdown_hash

MeloveGupta:
## What changed

db-builder now reads the readme data collector-watcher collects (from the last PR #775 ), it publishes the actual markdown files into the database, and stamps a ```markdown_hash``` field onto matching components so the frontend can eventually fetch them. I have also added a ```has_readme``` boolean to the slim index shape, similar to how java has ```has_standalone_library``` there, so a future list view could show some kind of indicator without needing the full component payload.

## Why

Part 2/3 of #762 

This is PR 2 of 3, this is the db-builder piece that actually makes that data usable.

## How it was tested

I ran the full explorer-db-builder suite, 348 passed (336 existing + 12 new). ruff check and format both clean.
I caught two things by actually running: I had to update an existing test that asserts the exact allowed keys on the slim bundle shape since I added ```has_readme``` to it, and one of my own new tests asserted ```".."``` should get fully stripped from a sanitized filename, ran it, it failed, the sanitizer only strips ```/``` not dots (matches the exact same sanitize behaviour from the watcher PR), so I fixed my test assertion instead of the code.

I also specifically tested that a readme-loading failure for one distribution doesn't take down the whole build, removed the try/except on purpose first to confirm it actually matters. Without it, a single OSError while loading one distribution's readmes killed the entire db-builder run across every version, not just that one distribution. I put it back and confirmed the whole suite is clean again.

## Is this a breaking change?

No

```write_markdown``` and ```_sanitize_name``` on ```CollectorDatabaseWriter``` are ported straight from ```DatabaseWriter``` (javaagent side), same content-addressing, same "never fail db generation on a readme error" comment. I didn't try to unify the two classes since that's the same generalization question that came up in the watcher PR.

## Checklist

### CODE REVIEW

**Comment by MeloveGupta** [comment]
Hey @jaydeluca @vitorvasc ,
In this PR, copilot flagged two things and I have fixed both, I've added regression tests for them as well.
All the tests are green, once this PR gets merged, I'll start with the part 3(frontend rendering) and get #762 fully wrapped up.
Also please let me know if any other further adjustments are required from my side on this PR :)

File: ecosystem-automation/explorer-db-builder/src/explorer_db_builder/collector_builder.py [hunk 1]
```diff
@@ -84,7 +91,35 @@ def _process_version(
...
+        try:
+            readme_map = inventory_manager.load_component_readme_map(distribution, version)
+            for component_name, markdown_hash in readme_map.items():
+                try:
+                    content = inventory_manager.load_component_readme_content(
+                        distribution, version, component_name, markdown_hash
+                    )
+                    if content is not None:
+                        db_writer.write_markdown(component_name, markdown_hash, content)
+                        published_readmes[component_name] = markdown_hash
```
jaydeluca: `write_markdown` swallows `OSError` internally and returns `None`, so this stamps `published_readmes[component_name]` regardless of whether the file was actually written. If the disk write fails, the component still gets `markdown_hash` → `has_readme: True` in the index/bundle, but no markdown file exists on disk — an internally inconsistent DB that becomes a 404 once PR 3 fetches it.

This also contradicts the docstring above (lines 70-75): _"Only components whose README was actually loaded and published **successfully** get a markdown_hash."_ A swallowed write failure isn't a successful publish.

Suggest making `write_markdown` return a success signal and stamping only on real success:

```suggestion
                    if content is not None and db_writer.write_markdown(
                        component_name, markdown_hash, content
                    ):
                        published_readmes[component_name] = markdown_hash
```

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/collector_database_writer.py
```diff
@@ -36,6 +37,39 @@ def __init__(self, database_dir: str = "ecosystem-explorer/public/data/collector
...
+        try:
+            with open(file_path, "w", encoding="utf-8") as f:
+                f.write(content)
+            self.files_written += 1
+            self.total_bytes += len(content.encode("utf-8"))
+            logger.debug("Wrote markdown for '%s' with hash %s", safe_name, markdown_hash)
+        except OSError as e:
+            logger.error("Failed to write markdown for '%s': %s", safe_name, e)
+            # README publishing failures must never fail DB generation, matching
+            # the same principle already established for the javaagent writer.
```
jaydeluca, line 80: Because this swallows `OSError` and returns `None`, two things follow:

1. The caller can't tell a write succeeded from a write that failed (see the stamp-regardless issue in `collector_builder.py`).
2. Combined with `load_component_readme_content` also swallowing `OSError` and returning `None` (`inventory_manager.py:317-319`), the per-component `try/except OSError` in `_process_version` is effectively dead code — nothing inside it raises `OSError` at runtime.

Recommend returning `True`/`False` (or the written `Path` / `None`) so the caller can stamp `has_readme` only when the file is really on disk. Keeping the "never fail DB generation" behavior is fine — just surface the outcome instead of hiding it.

ecosystem-automation/explorer-db-builder/tests/test_collector_builder.py
```diff
@@ -379,3 +384,221 @@ def test_returns_1_when_all_versions_empty(self, tmp_path):
...
+
+        exit_code = run_collector_builder(inventory_manager=manager, db_writer=db_writer)
+
+        assert exit_code == 0
+        with open(tmp_path / "collector" / "index.json") as f:
+            index_data = json.load(f)
+        by_name = {c["name"]: c for c in index_data["components"]}
+
+        assert by_name["goodreceiver"]["has_readme"] is True
+        assert by_name["badreceiver"]["has_readme"] is False
```
jaydeluca: This passes only because it monkeypatches `write_markdown` to _raise_ — the real writer swallows `OSError` and returns normally. In production the same disk failure would stamp `has_readme: True` (with no file), so this assertion is the opposite of what actually happens.

Once `write_markdown` returns a success signal (see the writer comment), please drive the real path instead of replacing the method — e.g. patch `builtins.open` to raise inside the real `write_markdown`, then assert `by_name["badreceiver"]["has_readme"] is False` and that no markdown file was written for it. That way the test guards the actual contract.

jaydeluca (commented):
Thanks @MeloveGupta!

One correctness gap to sort before merge: a silent markdown write failure still stamps `has_readme: True` with no file on disk, because `write_markdown` swallows `OSError` and the caller stamps regardless. The added regression test papers over this by patching `write_markdown` to raise, which the real writer never does. Suggested fix + test change are in the inline comments.

The sanitized-vs-raw name lookup is latent-only for collector names (lowercase-alnum), so that's a follow-up, not a blocker.

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/collector_builder.py
MeloveGupta: done, applied basically what you suggested...

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/collector_database_writer.py
MeloveGupta: done, I've updated it...

ecosystem-automation/explorer-db-builder/tests/test_collector_builder.py
MeloveGupta: done, I've applied the fix...

MeloveGupta:
hey @jaydeluca , I've addressed all the issues you mentioned and updated the PR, could you review it and let me know if any further adjustments are required from my side...

jaydeluca (approved):
thanks!

PR #807, opened 2026-07-14
fix(telemetry): prevent stale when condition in comparison results

### LINKED ISSUE #795: Telemetry Comparison displays incorrect diff after changing when condition while versions are loading
## Component/Area

Java Agent → Instrumentation Detail → Telemetry → Comparison

Note: This is different from #727. That issue involved the comparison direction for the oldest selected version, whereas this issue is a race condition caused by changing the when condition while version data is still loading!

---

## Expected behavior

The comparison results should always correspond to the currently selected `when` condition, even if the user changes the `when` condition while a version change is still loading.

---

## Actual behavior

Under a specific interaction sequence, the comparison results can be computed using a stale `when` condition while the UI reflects the newly selected one..!

This produces an inconsistent comparison where the displayed diff does not correspond to the currently selected `when` condition!

---

## Steps to reproduce

1. Open an instrumentation with multiple `when` conditions (for example, `spring-kafka-2.7`)
2. Open the **Telemetry → Comparison** tab
3. Enable network throttling (for example, Chrome DevTools → Slow 3G) so the version request remains in flight long enough
4. Change the **To Version** selector
5. Before the version request completes, change the `when` condition
6. Wait for the version request to finish

---

## Expected result

The comparison should always be generated using the latest selected `when` condition.

---

## Actual result

The comparison is generated using the previous when condition while the UI reflects the newly selected condition, resulting in an inconsistent comparison!

---

## Notes

During investigation I verified:

- The issue reproduces on the latest `main`.
- The behavior is isolated to the interaction window while the version request is in flight.
- Refreshing or triggering another comparison recomputes the correct result.
- The issue appears localized to the comparison state update rather than the backend data itself.

Adii-45:
## Summary

This PR fixes a stale state issue in the Java Agent Telemetry Comparison view..!

When a version change is still loading, changing the selected `when` condition could cause the comparison to be computed using the previous condition while the UI displayed the newly selected one. This resulted in inconsistent comparison results until another interaction triggered a recomputation!

The comparison logic now always uses the latest selected `when` condition when the asynchronous comparison resolves!

## Changes

- Use the latest `when` condition during asynchronous comparison resolution
- Keep the fix scoped to the comparison hook without changing the API or metadata model
- Add a regression test covering the interaction where the `when` condition changes while a version request is still in flight

## Testing

- Reproduced the issue locally on the latest `main`
- Verified that the comparison now reflects the latest selected `when` condition after the request completes
- Added a regression test for the reported scenario

Fixes #795

### DIFF
--- a/ecosystem-explorer/src/features/java-agent/components/agent-explore-landing.test.tsx
+++ b/ecosystem-explorer/src/features/java-agent/components/agent-explore-landing.test.tsx
@@ -51,7 +51,7 @@ describe("AgentExploreLanding", () => {
     });
     expect(configRefLink).toHaveAttribute(
       "href",
-      "https://opentelemetry.io/docs/zero-code/java/agent-config/"
+      "https://opentelemetry.io/docs/zero-code/java/agent/configuration/"
     );
 
     const githubLink = screen.getByRole("link", {
--- a/ecosystem-explorer/src/features/java-agent/hooks/use-telemetry-comparison.test.ts
+++ b/ecosystem-explorer/src/features/java-agent/hooks/use-telemetry-comparison.test.ts
@@ -1,0 +1,91 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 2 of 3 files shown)


jaydeluca (approved):
thanks!

Issue #373, opened 2026-05-06
[UI/UX] Phase 4: List pages


vitorvasc:
Part of #84. Depends on #370.

This issue tracks the list pages, the biggest UX win of the redesign. Replaces the current Search + Type + Version filter row with a real left-rail facet panel (Type, Signal, Stability, Distribution, Version), a density toggle (Cards / Compact / Table), active-filter chips, and full URL-persisted state. 

See [`projects/84-ui-ux-design/03-list-page.md`](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/blob/main/projects/84-ui-ux-design/03-list-page.md) for the full plan.

The components will be shipped behind the `V1_REDESIGN` feature flag and depend on the foundation primitives from #370 (NavBar, SubNav, theme, StatusPill, TypeStripe, Card). PR 1 below is the shared URL contract — it must land first because Phase 3's pipeline-anatomy stages also depend on it.

## PRs

- [x] PR 0 - `listFilters` parser/serializer - shipped in #685 (shared with Phase 3's deep links)
- [x] PR 1 - Facet primitives — `<CheckboxFacet>`, `<SearchFacet>`, `<SelectFacet>` - #784
- [x] PR 2 - `<FacetPanel>` — sticky on desktop, drawer on mobile; composes the primitives - #803 
- [x] PR 3 - List controls — `<ActiveFilterChips>`, `<DensityToggle>`, `<SortDropdown>`, `<Pagination>`, `<EmptyState>`, `<FacetDrawerToggle>` - #786 
- [x] PR 4 - Result views — `<CompactList>`, `<CardView>`, `<TableView>` - #787 
- [x] PR 5 - List route — `/<ecosystem>/components`, URL ↔ state wiring, plus the #645 index-shape adaptation - #834 
- [x] PR 6 - Empty/error/loading polish + visual regression on each density - #841

SurbhiAgarwal1:
I'd love to take this on! Based on the requirements in 03-list-page.md, here is my proposed implementation plan to build out the faceted browse experience:

Proposed Implementation Plan
1. URL-Driven State Foundation I'll start by implementing a typed ListFilters parser/serializer to ensure the URL query params are the single source of truth for all filtering, sorting, and density preferences. This will ensure every browse state is fully deep-linkable and shareable.

2. Component Architecture

[...]

PR #841, opened 2026-07-20
feat(v1): add list visual regression + drawer a11y (Phase 4 - PR 6)

vitorvasc:
Phase 4 close-out: the two follow-ups PR 5 left for last — the mobile facet drawer's modal behavior and the list page's visual-regression coverage — plus light-theme contrast fixes the new axe captures surfaced.

Contributes to #84 and #373.

Stacked on #834 — the diff shows its commit until it merges; only `20f3906c` is this PR's.

## What's in this PR

- The open facet drawer is now a real modal: `role="dialog"` + `aria-modal`, focus moves into the drawer on open and returns to `FacetDrawerToggle` on close, Tab wraps inside it, body scroll locks, and a scrim click dismisses it (alongside Escape and the close button). The panel renders as a `div` with an explicit `complementary`/`dialog` role switch because ARIA forbids `role="dialog"` on `aside`
- `scripts/take-screenshots.mjs` captures the collector list once per density (compact, `?density=cards`, `?density=table`) and the open drawer on the mobile viewport, each with an axe report
- Light-theme WCAG AA fixes for what the new captures flagged (ratios are worst case, over the slate list-row background):
  - `GlowBadge` light text shades: orange/green 600→800 (2.97/2.72 → 6.09/6.03), blue/red 600→700 (4.29/3.81 → 5.59/5.12). Dark shades unchanged
  - Light `--muted-foreground-hsl` darkened from 47% to 40% lightness — slate-500 only clears 4.5:1 on pure white, and the list rows and `bg-muted` chips sit on tinted surfaces (3.95 → 5.14 on the hovered row)
  - Active `DensityToggle` label reads in `foreground` instead of `primary`, which landed at 2.3:1 on the tinted background in dark mode
- 6 new tests: drawer dialog semantics, focus trap + restore, scroll lock, scrim dismissal, and the list-page round trip through the toggle

## What's not in this PR

- The remaining light-theme axe findings sit on other phases' surfaces (StatsBand language badge, pipeline-anatomy stage counts, legacy pages' heading order) and predate this PR
- `error-boundary.tsx` uses the same failing `text-red-600` pattern but never renders in a capture; left for the end-of-redesign cleanup

## Verification

- `bun run typecheck` / `bun run lint` — clean
- `bun run test` — 1281 tests pass (140 files), including the 6 new drawer tests
- `bun run screenshots:capture` with the V1 + showcase flags — 26 page×theme axe scans; the 8 list-page reports (3 densities + drawer × 2 themes) show zero violations

---

> Co-authored with Claude Fable 5.


jaydeluca (approved):
this is looking soooo good !

PR #879, opened 2026-07-21
feat(java-agent): add status filter for release comparison

### LINKED ISSUE #874: [Java Agent Release Comparison] Filter module changes
As a user using the release comparison tool (https://explorer.opentelemetry.io/java-agent/releases) I want to be able to filter the details in teh "changes summary" tab. For example, if I want to see just the module changes where there were changes to metrics, or filter to just added/removed modules.

Image

Adii-45:
## Summary

Closes #874

This PR adds a status filter to the Java Agent Release Comparison page, allowing the Module Changes list to be filtered by change type.

## What's included

- Added a Status filter for the Module Changes section
- Supports:
  - All Statuses
  - Added
  - Changed
  - Removed
- Uses the existing application DropdownMenu pattern for a consistent UI
- Preserves URL state so filters survive refresh and navigation
- Added/updated unit and integration tests

## Testing

- Verified Added/Changed/Removed filtering
- Verified URL synchronization
- Verified refresh persistence
- Verified browser back/forward navigation
- Ran:
  - `bun run typecheck`
  - `bun run lint`
  - `bun run test`

Screenshot 2026-07-21 at 10 08 35 PM

### DIFF
--- a/ecosystem-explorer/src/features/java-agent/java-release-comparison-page.tsx
+++ b/ecosystem-explorer/src/features/java-agent/java-release-comparison-page.tsx
@@ -57,9 +57,15 @@ export function JavaReleaseComparisonPage() {
   useEffect(() => {
     if (versions.length > 0 && (!searchParams.get("from") || !searchParams.get("to"))) {
       setSearchParams(
-        {
-          from: fromVersion || versions[Math.min(1, versions.length - 1)].version,
-          to: toVersion || versions[0].version,
+        (prev) => {
+          const params = new URLSearchParams(prev);
+          if (!params.has("from")) {
+            params.set("from", fromVersion || versions[Math.min(1, versions.length - 1)].version);
+          }
+          if (!params.has("to")) {
+            params.set("to", toVersion || versions[0].version);
+          }
+          return params;
         },
         { replace: true }
       );
@@ -75,11 +81,21 @@ export function JavaReleaseComparisonPage() {
   } = useReleaseComparison(fromVersion, toVersion, validVersionStrings);
 
   const handleFromVersionChange = (version: string) => {
-    setSearchParams({ from: version, to: toVersion });
+    setSearchParams((prev) => {
+      const params = new URLSearchParams(prev);
+      params.set("from", version);
+      params.set("to", toVersion);
+      return params;
+    });
   };
 
   const handleToVersionChange = (version: string) => {
-    setSearchParams({ from: fromVersion, to: version });
+    setSearchParams((prev) => {
+      const params = new URLSearchParams(prev);
+      params.set("from", fromVersion);
+      params.set("to", version);
+      return params;
+    });
   };
 
   const statusFilter = useMemo(() => parseStatusParam(searchParams.get("status")), [searchParams]);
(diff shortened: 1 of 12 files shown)


Adii-45:
I think this PR is ready for review... Please let me know if any changes are needed... Thanks!

ecosystem-explorer/src/features/java-agent/java-release-comparison-page.tsx
```diff
@@ -25,8 +25,21 @@ import { useVersions } from "@/hooks/use-javaagent-data";
 import { useReleaseComparison } from "./hooks/use-release-comparison";
 import { ReleaseVersionSelector } from "./components/release-comparison/release-version-selector";
 import { InstrumentationDiffCard } from "./components/release-comparison/instrumentation-diff-card";
+import {
+  ModuleStatusFilter,
+  type ModuleChangeStatus,
+} from "./components/release-comparison/module-status-filter";
 import { GlowBadge } from "@/components/ui/glow-badge";
 
+const VALID_MODULE_STATUSES: ModuleChangeStatus[] = ["added", "changed", "removed"];
```
jaydeluca, line 38: I had also mentioned adding filters for telemetry types too, so that you could pair it with this one and filter down to like only modules that changed and had a metric change. 

We can model it just like we do on the java instrumentation list page, with buttons:

Image

Adii-45: Thanks! Implemented the telemetry filter following the Java Instrumentation page pattern... It now works together with the existing Status filter to narrow down the results. Happy to make any further changes if needed!
Screenshot 2026-07-23 at 7 34 57 AM

jaydeluca: looks good! but i think theres a bug, when I select metrics, i still see items that just have span changes (see "JDBC"):

image

Adii-45: Fixed in the latest commit... The root cause was that the filter checked for the presence of metrics rather than actual changed metrics, so modules with only unchanged metric entries could still match! It now filters only real telemetry changes, and I've added a regression test for the JDBC scenario...!

ecosystem-explorer/src/features/java-agent/java-release-comparison-page.tsx
```diff
@@ -292,16 +407,30 @@ export function JavaReleaseComparisonPage() {
...
+                            onChange={handleTelemetryFilterChange}
+                          />
+                          <ModuleStatusFilter
+                            value={statusFilter}
+                            onChange={handleStatusFilterChange}
+                          />
+                        </>
+                      )}
+                      <div className="bg-muted/50 text-foreground/70 rounded-full px-4 py-1 text-xs font-bold">
+                        {t("releaseComparison.modulesImpacted", {
```
jaydeluca, line 426: can you look into why we see differnt numbers between "modules changed" and the "modules impacted"? I would think by default they should be the same?

Image

jaydeluca: nevermind, i misunderstood, and the 3 numbers add up to the lower one. ignore me.

jaydeluca (approved):
thanks!

ecosystem-explorer/src/features/java-agent/java-release-comparison-page.tsx
Adii-45: I dug into this a bit, and from what I found the two numbers are currently measuring slightly different things!

For example, if a comparison has:
- Added modules: 8
- Removed modules: 6
- Changed modules: 22

Then:
- **Modules Changed** = 22 (only modules that existed in both versions and actually changed)
- **Modules Impacted** = 8 + 6 + 22 = 36 (added + removed + changed)

So the difference seems to come from what each metric is counting rather than an incorrect calculation...

If your expectation is that these should always show the same value, I'm happy to dig further and investigate what the intended behavior should be, or put together a follow-up fix (PR) if the current semantics aren't what we want...

PR #886, opened 2026-07-23
Perf/875 java release comparison

### LINKED ISSUE #875: [Java Agent Release Comparison] Can performance be improved?
Opening https://explorer.opentelemetry.io/java-agent/releases?from=2.28.1&to=2.29.0 results in a long wait for the comparison results to populate. Can we improve this?

vanshika2720:
## What changed

Optimized Java Agent release comparison by avoiding unnecessary work for unchanged modules during release comparisons. Added coverage for custom-to-library and library-to-custom module moves and comparison race conditions.

## Why

Fixes #875

The release comparison previously processed unchanged modules unnecessarily. This change skips unchanged modules while preserving correct behavior for module moves and concurrent comparison scenarios.

## How it was tested

- Ran the existing test suite and added tests covering custom-to-library and library-to-custom module moves and comparison race conditions.
- Verified the release comparison manually at:
  `/java-agent/releases?from=2.28.1&to=2.29.0`
- Verified that the comparison renders successfully and displays the expected module change summary.
- Compared the release comparison on `main` and this branch using the same release versions (`2.28.1` → `2.29.0`).
- On `main`, the comparison made 339 detail requests.
- On this branch, the comparison made 174 detail requests.

### Before
Screenshot 2026-07-23 at 11 16 40 AM

### After
Screenshot 2026-07-23 at 11 17 05 AM

## Is this a breaking change?

No

The performance optimization skips modules that are unchanged between the selected releases. The added tests cover the edge cases where modules move between custom and library categories and ensure comparison results remain correct under race conditions.

## Checklist

### DIFF
--- a/ecosystem-explorer/scripts/benchmarks/measure-release-comparison.mjs
+++ b/ecosystem-explorer/scripts/benchmarks/measure-release-comparison.mjs
@@ -1,0 +1,270 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Benchmarks Java Agent release-comparison performance: compares a server
+// running `main` against a server running a PR branch, on the same
+// comparison URL, across simulated network latencies.
+//
+// Usage:
+//   bun scripts/benchmarks/measure-release-comparison.mjs [options]
+//
+// Options (all optional, shown with defaults):
+//   --main-url=http://localhost:5180
+//   --pr-url=http://localhost:5181
+//   --from=2.28.1
+//   --to=2.29.0
+//   --runs=5                per branch, per latency
+//   --latencies=0,40,80     comma-separated milliseconds
+import { chromium } from "playwright";
+
+function parseArgs(argv) {
+  const opts = {
+    mainUrl: "http://localhost:5180",
+    prUrl: "http://localhost:5181",
+    from: "2.28.1",
+    to: "2.29.0",
+    runs: 5,
+    latencies: [0, 40, 80],
+  };
+  for (const arg of argv) {
+    const [key, value] = arg.replace(/^--/, "").split("=");
+    if (key === "main-url") opts.mainUrl = value;
+    else if (key === "pr-url") opts.prUrl = value;
+    else if (key === "from") opts.from = value;
+    else if (key === "to") opts.to = value;
+    else if (key === "runs") opts.runs = Number(value);
+    else if (key === "latencies") opts.latencies = value.split(",").map(Number);
+  }
+  return opts;
+}
+
+const opts = parseArgs(process.argv.slice(2));
+const urlPath = `/java-agent/releases?from=${opts.from}&to=${opts.to}`;
+
+// A stuck browser.close() or CDP call can pin the event loop past a
+// Promise.race timeout (there's no real cancellation in JS), so this is the
+// actual backstop against the script hanging forever.
+const totalMeasurements = 2 * opts.latencies.length * opts.runs;
+const WATCHDOG_MS = totalMeasurements * 30000 + 30000;
+setTimeout(() => {
+  console.error(`watchdog fired after ${WATCHDOG_MS}ms - forcing exit`);
+  process.exit(1);
+}, WATCHDOG_MS);
+
+function classify(url) {
+  if (url.includes("/versions-index.json")) return "versions-index";
+  if (/\/versions\/[^/]+-index\.json$/.test(url)) return "manifest";
+  if (url.includes("/instrumentations/") && url.endsWith(".json")) return "detail";
+  if (url.includes("/bundles/")) return "bundle";
+  return "other";
+}
+
+function withTimeout(promise, ms, label) {
+  return Promise.race([
+    promise,
+    new Promise((_, reject) =>
+      setTimeout(() => reject(new Error(`${label} timed out after ${ms}ms`)), ms)
+    ),
+  ]);
+}
+
+async function measureOnce(baseUrl, latencyMs) {
+  const browser = await chromium.launch();
+  try {
+    return await withTimeout(
+      measureWithBrowser(browser, baseUrl, latencyMs),
+      25000,
+      "measureWithBrowser"
+    );
+  } finally {
+    // Give up on a hung close() after 5s rather than block the retry loop;
+    // the watchdog above is the real backstop against a leaked handle.
+    await withTimeout(browser.close(), 5000, "browser.close").catch(() => {});
+  }
+}
+
+async function measureWithBrowser(browser, baseUrl, latencyMs) {
+  // Fresh, unpersisted context per run: no HTTP cache, no IndexedDB, so every
+  // run issues the same network requests a reader's first visit would.
+  const context = await browser.newContext();
+  const page = await context.newPage();
+
+  if (latencyMs > 0) {
+    const client = await context.newCDPSession(page);
+    await client.send("Network.emulateNetworkConditions", {
+      offline: false,
+      latency: latencyMs,
+      downloadThroughput: (20 * 1024 * 1024) / 8, // 20 Mbps, generous - isolate latency
+      uploadThroughput: (10 * 1024 * 1024) / 8,
+    });
+  }
+
+  const responses = [];
+  page.on("response", async (res) => {
+    const url = res.url();
+    if (!url.includes("/data/javaagent/")) return;
+    let bytes = 0;
+    try {
+      bytes = (await res.body()).length;
+    } catch {
+      // response body unavailable (e.g. redirected/aborted) - count as 0
+      // bytes, the request is still counted.
+    }
+    responses.push({ url, kind: classify(url), bytes });
+  });
+
+  const t0 = performance.now();
+  await page.goto(`${baseUrl}${urlPath}`, { waitUntil: "domcontentloaded" });
+
+  // #panel-changes only renders once the diff has finished loading (see
+  // java-release-comparison-page.tsx), so waiting for it plus real content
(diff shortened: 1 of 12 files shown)


jaydeluca:
i just tested it on the deploy preview and its still very slow, im not noticing any improvement. 

The results you shard in your screenshot show that there wasn't any actual difference, doesnt it? Looks like before it was around 3000ms and now its still around 3000ms in your test results?

image

vanshika2720:
@jaydeluca I investigated further and found that localhost's near-zero latency and a fixed page-boot cost were masking the benefit. With simulated 40ms and 80ms latency, the PR showed ~29% and ~41% faster comparison loading, respectively, while reducing detail requests from 339 to 174. I’ll add the reproducible benchmark details to the PR.

vanshika2720:
@jaydeluca  I dug into this further. The original screenshot was from a near-zero-latency localhost run, where the fixed page boot cost dominated, so both branches appeared to take roughly the same time.

I ran a controlled benchmark with **5 runs per branch** for the same comparison (`2.28.1 → 2.29.0`):

| Latency | `main` median | PR median | Improvement |
| --- | ---: | ---: | ---: |
| 0ms | 945ms | 942ms | 0.3% faster |
| 40ms | 3651ms | 2116ms | 42.0% faster |
| 80ms | 5786ms | 3773ms | 34.8% faster |

The optimization reduces detail requests from **339 → 174**. At 0ms latency, the difference is negligible, which explains the original screenshot. With network latency, the reduced request count results in a significant improvement in comparison loading time.

I've attached the benchmark output for reference.
Screenshot 2026-07-26 at 3 09 16 PM

ecosystem-explorer/src/features/java-agent/hooks/use-release-comparison.ts
```diff
@@ -51,34 +64,53 @@ export function useReleaseComparison(
...
+          // Genuinely still waiting on the version list - keep spinning.
+          setLoading(true);
           setError(null);
-          return;
+        } else {
+          // The version list finished loading (or failed) and is still
+          // empty, so it isn't coming - stop waiting instead of spinning
+          // forever on a request that will never resolve.
+          setLoading(false);
+          setError(new Error("Failed to load the list of available versions."));
```
jaydeluca, line 81: This English string reaches the user — it's interpolated into `t("releaseComparison.errorLoading", { message })`, so Spanish readers get "Error al cargar datos de comparación: Failed to load the list of available versions."

The cause is that the page takes only `{ data, loading }` from `useVersions()` (`java-release-comparison-page.tsx:65`) though the hook also returns `error`. Consume it there and render a localized message, rather than having this hook invent an error about data it doesn't own. That also drops the need for the `versionsLoading` parameter.

jaydeluca (commented):
I replayed the skip logic against the committed manifests: 2.28.1 → 2.29.0 skips 165 of 251 modules, 339 → 174 detail requests. That checks out, and dropping unchanged entries is safe since the page already filters `status === "unchanged"` (`java-release-comparison-page.tsx:183`).

Three findings that can't be inline:

**1. The wall clock is concurrency-bound, which is why the page still feels slow.** 174 requests through one pool of 8 is ~22 serialised round trips; the old path had 16 slots over 339 fetches, ~31 waves. That ~30% gap is what your benchmark measured — halving requests couldn't halve latency, because `MAX_INSTRUMENTATION_FETCH_CONCURRENCY = 8` is now the binding constraint. Please note that in the description so nobody trims requests further for diminishing returns. The real lever is a precomputed diff/metrics bundle in the data pipeline, like the list page's `loadInstrumentationBundle` — I can open that as a follow-up.

**2. `javaagent-data.ts:30-36` is stale** (unchanged lines, so not commentable inline). It says the constant caps the `loadAllInstrumentations` fan-out *fallback* and that the primary path uses a bundle. `loadComparisonDetails` is a third consumer on a primary path.

**3. No page-level test opens the metrics tab.** `java-release-comparison-page.test.tsx` isn't in the diff; it doesn't mock `useAggregateMetrics` and never sets `activeTab` to `"metrics"`, so tests pass only because `enabled` stays false. The tab went from synchronous to async — one test that clicks it would catch both UI issues below.

The dashboard still lists Copilot's metrics error-message thread as waiting on you, but `05c2b7ad` already fixed it. A reply there will route this back to reviewers.

ecosystem-explorer/src/features/java-agent/java-release-comparison-page.tsx
```diff
@@ -481,57 +490,83 @@ export function JavaReleaseComparisonPage() {
                     </h2>
                     <div className="bg-muted/50 text-foreground/70 rounded-full px-4 py-1 text-xs font-bold">
                       {t("releaseComparison.totalMetrics", {
-                        count: diff.aggregateMetrics.length,
+                        count: aggregateMetrics?.length ?? 0,
```
jaydeluca, line 493: The badge is outside all three new loading/error/success guards, so it shows "0 Total Metrics in 2.29.0" next to the spinner for the whole ~250-request load. Move it into the success branch or show a placeholder while `metricsLoading`.

ecosystem-explorer/src/features/java-agent/hooks/use-aggregate-metrics.ts
```diff
@@ -0,0 +1,77 @@
...
+ * @param enabled Whether to load. Stays idle (and keeps prior results) while false.
+ */
+export function useAggregateMetrics(version: string, enabled: boolean) {
+  // Tagged with the version it was loaded for so a version change is never
+  // shown alongside data loaded for the previous one: React runs this effect
+  // after paint, so the first render after `version` changes would otherwise
+  // repaint with the old `metrics` value for one frame before the fetch below
+  // even starts.
+  const [state, setState] = useState<{ version: string; metrics: AggregateMetric[] } | null>(null);
+  const [loading, setLoading] = useState(false);
```
jaydeluca, line 40: On the render where `activeTab` flips to `"metrics"`, `enabled` is true but the effect hasn't run: `loading` false, `metrics` null, `error` null, so no branch renders and the panel is blank with no spinner until the effect commits.

Init `loading: true` (the convention in `use-collector-data.ts`) is wrong here since the hook is gated — derive it instead: `loading: loading || (enabled && !state && !error)`.

ecosystem-explorer/src/features/java-agent/hooks/use-aggregate-metrics.test.ts
```diff
@@ -0,0 +1,141 @@
...
+    const { result, rerender } = renderHook(
+      ({ enabled }) => useAggregateMetrics("2.10.0", enabled),
+      { initialProps: { enabled: false } }
+    );
+
+    expect(javaagentData.loadAllInstrumentationDetails).not.toHaveBeenCalled();
+
+    rerender({ enabled: true });
+
+    await waitFor(() => expect(result.current.loading).toBe(false));
```
jaydeluca, line 73: Since `loading` starts `false`, this `waitFor` is satisfied by the initial state and can return before the fetch settles — the assertions after it pass on `act`'s microtask flush, not on the wait. Same on line 86. Wait on the result instead: `expect(result.current.metrics).not.toBeNull()` / `expect(result.current.error).not.toBeNull()`.

ecosystem-explorer/src/features/java-agent/java-release-comparison-page.tsx
```diff
@@ -94,7 +95,15 @@ export function JavaReleaseComparisonPage() {
...
-  } = useReleaseComparison(fromVersion, toVersion, validVersionStrings);
+  } = useReleaseComparison(fromVersion, toVersion, validVersionStrings, versionsLoading);
+
+  // The metrics tab is a whole-release rollup and costs far more to load than
+  // the diff, so it stays unloaded until the reader opens that tab.
+  const {
+    metrics: aggregateMetrics,
+    loading: metricsLoading,
+    error: metricsError,
+  } = useAggregateMetrics(toVersion, activeTab === "metrics" && !!diff);
```
jaydeluca, line 106: This makes the metrics tab slower for readers who open it — it used to be free (the diff had loaded every module) and is now ~250 requests behind a spinner. The deferral is still the right call, but the description should say so rather than reading as an unqualified win.

ecosystem-explorer/src/lib/api/javaagent-data.ts
```diff
@@ -204,6 +204,100 @@ export async function loadAllInstrumentationDetails(
...
+  return identities;
+}
+
+export interface ComparisonDetails {
+  /** Detail for modules that differ between the two versions — the `from` side. */
+  fromData: InstrumentationData[];
+  /** Detail for modules that differ between the two versions — the `to` side. */
+  toData: InstrumentationData[];
+  /** Ids skipped because both manifests point at the identical detail file. */
+  unchangedIds: string[];
```
jaydeluca, line 232: Only read in tests — `useReleaseComparison` discards it. Either drop it, or surface it: "165 modules unchanged" would be useful next to the change count.

ecosystem-explorer/src/features/java-agent/utils/release-diff.ts
```diff
@@ -301,33 +356,8 @@ export function compareReleases(
...
-
-  const aggregateMetrics = Array.from(metricToInstrumentations.entries())
-    .map(([name, data]) => ({
-      name,
-      description: data.description,
-      emittedBy: data.emittedBy.sort(),
-    }))
-    .sort((a, b) => a.name.localeCompare(b.name));
+  const aggregateMetrics =
+    options.includeAggregateMetrics === false ? [] : computeAggregateMetrics(safeToData);
```
jaydeluca, line 360: `aggregateMetrics` is still typed as populated but is always `[]` on the comparison path, and `[]` is indistinguishable from "this release emits no metrics". `aggregateMetrics?: AggregateMetric[]`, or a separate return type for the opted-out call, puts that in the type system instead of the option's doc comment.

ecosystem-explorer/scripts/benchmarks/measure-release-comparison.mjs
```diff
@@ -0,0 +1,270 @@
...
+//   bun scripts/benchmarks/measure-release-comparison.mjs [options]
+//
+// Options (all optional, shown with defaults):
+//   --main-url=http://localhost:5180
+//   --pr-url=http://localhost:5181
+//   --from=2.28.1
+//   --to=2.29.0
+//   --runs=5                per branch, per latency
+//   --latencies=0,40,80     comma-separated milliseconds
+import { chromium } from "playwright";
```
jaydeluca, line 31: This will rot: no `package.json` entry, no README, no CI use, and it needs two manually started servers on hard-coded ports. Add a `bench:release-comparison` script or a header note on bringing the servers up.

PR #894, opened 2026-07-24
Fix/893 smoother page loads

### LINKED ISSUE #893: [Explorer] Smoother page loads
When loading pages for the first time, things often load in strange orders and things jump around a bit before being presented to the user. 

The https://explorer.opentelemetry.io/java-agent page is an example, where the "news" section loads after the other panels sometimes.

We should investigate better ways to handle async data loading and the way we present pages to the user so it doesn't feel like 8 different components are all loading and updating the page independently

malydev:
## Description

This PR improves the initial loading experience of the Java Agent announcements section.

Previously, `java-agent-announcements.tsx` returned `null` while the announcements were loading. As a result, no space was reserved for the section, and the entire block appeared later, pushing the content below it and causing a noticeable layout shift.

The section now renders immediately with its heading and skeleton cards while the announcements are loading. This keeps the layout stable and provides the user with visible loading feedback.

## Changes

* Render the announcements section while data is loading instead of returning `null`.
* Display the section heading immediately.
* Add skeleton cards that reserve the expected layout space.
* Render at least three skeleton cards to represent a complete row.
* Use the announcement count when it is greater than three.

## Before

The announcements section was completely absent during loading. Once the data became available, the section appeared suddenly and shifted the rest of the page content.

## After

The section appears immediately with loading placeholders. The skeleton cards reserve space until the announcements are available, reducing layout shifts and making the loading process clearer.

## Demo

<div align="center">
  <video src="https://github.com/user-attachments/assets/3d69bbfe-1ea5-4b00-99ec-8ae2e00948bf" controls width="640"></video>
</div>

## Testing

* Started the Ecosystem Explorer locally.
* Opened the Java Agent page.
* Tested the initial page load with network throttling enabled.
* Confirmed that the announcements section and heading appear during loading.
* Confirmed that the skeleton cards are replaced by the announcements after the data loads.
* Confirmed that the section no longer appears late and pushes the content below it.

Part of #893


jaydeluca:
hey @malydev are you able to sign the CLA? thanks!

jaydeluca (approved):
thanks!

Issue #876, opened 2026-07-21
Collector watcher not pulling all metadata


jaydeluca:
I noticed that #336 is not working, likely because we aren't properly pulling the data with the watcher.

Looking at the memory limiter processor metadata, there is telemetry:

https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/memorylimiterprocessor/metadata.yaml#L19-L28

But the page shows none:

<img width="1380" height="678" alt="Image" src="https://github.com/user-attachments/assets/98ea06cb-cd86-4adf-8a72-b1afbd34c4b5" />

Looking at our registry, we don't have any telemetry data tracked at all:

https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/blob/main/ecosystem-registry/collector/core/v0.156.0/processor.yaml#L22-L38

We need to update the collector-watcher to pull in all of the available metadata, including telemetry.

We will then need to run a backfill, which will be done after the change is merged, so do not include the re-generated metadata in the PR.

vanshika2720:
Hey @jaydeluca , I would love to work on this...
Could you assign it to me???

jaydeluca:
@vanshika2720 thanks let me know if you have any questions

vanshika2720:
@jaydeluca I have one question should the parser preserve all currently unknown metadata fields generically for forward compatibility, or would you prefer only adding support for the fields that exist in the current upstream schema (like telemetry)?

jaydeluca:
we are tracking the schema of the files separately, so I think it's safe for us to handle it generically and allow any unknown fields

Adii-45:
@jaydeluca I went through this issue and checked the follow-up work in #897! It looks like the original metadata loss issue has been fully resolved on main..

I believe this issue can be closed now, but please correct me if I’m missing anything!

PR #897, opened 2026-07-26
fix(collector-watcher): preserve unknown metadata.yaml fields

vanshika2720:
MetadataParserV1 previously copied only a fixed allowlist of metadata fields, causing valid fields such as `telemetry`, `tests`, `sem_conv_version`, and `config` to be silently dropped from upstream `metadata.yaml` files. The same allowlist behavior also applied to nested fields inside `status`, `attributes`, and `metrics`.

Since the registry tracks the schema of each file separately, the parser can safely preserve unknown fields generically. Known fields retain their existing normalization, while unknown fields are now passed through verbatim. `telemetry.metrics` reuses the existing metric normalization logic.

Part of #876

## What changed

Updated `MetadataParserV1` to preserve all metadata fields from upstream `metadata.yaml` files instead of filtering them through a fixed allowlist. Known fields keep their existing normalization, while unknown top-level and nested fields are preserved generically, including `telemetry`.

Added and updated tests covering telemetry preservation, unknown fields, nested fields, and deterministic output.

## Why

The collector-watcher was dropping valid metadata fields that were not explicitly handled by the parser, including `telemetry`. This prevented the complete upstream metadata from being stored in the registry.

The parser now preserves unknown fields for forward compatibility, as the metadata schema is tracked separately.

## How it was tested

- Added and updated tests for telemetry and unknown metadata field preservation.
- Added regression coverage for nested unknown fields and deterministic output.
- Verified against the real `memorylimiterprocessor` metadata structure.
- `189` tests passed.
- `ruff check` passed.
- `ruff format --check` passed.

Registry metadata regeneration/backfill is intentionally not included and can be performed after this PR is merged, as requested in the issue.

## Is this a breaking change?

No.

This PR is scoped to `collector-watcher` and its tests. It fixes the metadata parsing layer so fields such as `telemetry` are preserved for the subsequent registry backfill.

## Checklist


vanshika2720:
@jaydeluca I also checked the data flow. This PR is focused on fixing collector-watcher to preserve telemetry, as requested in this issue. The backfill is intentionally deferred until after merge.

Exposing the backfilled telemetry in the UI appears to require separate downstream changes. Would you like me to handle that as a follow-up PR as well?

ecosystem-automation/collector-watcher/src/collector_watcher/metadata_parser.py
```diff
@@ -104,6 +113,19 @@ def parse(self, raw: dict[str, Any]) -> dict[str, Any] | None:
...
+        if "telemetry" in raw:
+            parsed["telemetry"] = self._parse_telemetry(raw["telemetry"])
+
+        # Anything else on the file - present today or added upstream later -
+        # has no sanitization or ordering requirement of its own, so it is
+        # copied through unchanged instead of being silently dropped. This is
+        # what lets the parser absorb future mdatagen schema growth without a
+        # code change per field; only fields needing bespoke handling (above)
+        # get one.
+        for key, value in raw.items():
```
jaydeluca: i think we could probably remove the lines above this that simply re-assign values and don't do any additional processing (like `_parse_telemetry` or `_parse_attributes` ones could say)

Also, the result here is not deterministically ordered, so we should also ensure we sort the results to fix that

ecosystem-automation/collector-watcher/src/collector_watcher/metadata_parser.py
```diff
@@ -70,8 +70,17 @@ def get_schema_version(self) -> str:
...
+    list ordering) to: type, display_name, description, status, attributes,
+    metrics, resource_attributes, telemetry. Every other field present in the
+    source file — `tests`, `sem_conv_version`, `config`, `entities`, `events`,
+    `feature_gates`, or anything the upstream mdatagen schema adds later — is
+    passed through verbatim rather than dropped, so new top-level fields do
+    not require a code change here. The same pass-through-unless-normalized
+    rule applies one level down inside `status`, each `attributes`/
+    `resource_attributes` entry, and each `metrics` entry: only the specific
+    sub-fields that need sanitization or stable ordering are enumerated, and
+    anything else on those objects is preserved as-is.
```
jaydeluca: could we also incorporate an "exclusion list" filter of some sort? im thinking for starters we can ignore all items under the `test` node

ecosystem-automation/collector-watcher/tests/test_metadata_parser.py
```diff
@@ -271,6 +271,70 @@ def test_parse_complete_metadata(temp_component_dir):
...
+      sum:
+        value_type: int
+        monotonic: true
+"""
+    create_metadata_file(temp_component_dir, content)
+    parser = MetadataParser(temp_component_dir)
+    metadata = parser.parse()
+
+    assert metadata is not None
+    # The field the issue reports as missing.
```
jaydeluca: can we remove some of the non-useful commentary in these comments? basically anythign referencing the issue could be removed

ecosystem-automation/collector-watcher/src/collector_watcher/metadata_parser.py
vanshika2720: Removed the redundant reassignments,the merged output is now sorted deterministically, including nested metadata structures.

ecosystem-automation/collector-watcher/src/collector_watcher/metadata_parser.py
vanshika2720: Added an exclusion list at the parser level, currently excluding tests. It can be extended as additional fields are identified.

ecosystem-automation/collector-watcher/tests/test_metadata_parser.py
vanshika2720: @jaydeluca Cleaned up the issue-specific commentary from the tests while keeping comments that explain the test intent.

ecosystem-automation/collector-watcher/src/collector_watcher/metadata_parser.py
```diff
@@ -70,40 +70,59 @@ def get_schema_version(self) -> str:
...
+    sanitization or stable ordering are enumerated, and anything else on
+    those objects is preserved as-is. `EXCLUDED_FIELDS` names top-level
+    fields dropped entirely instead of passed through.
     """
 
+    #: Top-level metadata.yaml fields dropped from parsed output entirely,
+    #: rather than passed through generically. `tests` is fixture/test-only
+    #: config with no value to the registry. Extend this set to exclude
+    #: additional fields in the future.
+    EXCLUDED_FIELDS: frozenset[str] = frozenset({"tests"})
```
jaydeluca: `EXCLUDED_FIELDS` only applies at the top level, which leaves one field I think we want to also drop: `status.disable_codecov_badge`. The db-builder forwards `status` wholesale (`"status": status` in `collector_transformer.py`), so it lands in the frontend database but is not useful.

The other two new `status` sub-fields are worth keeping: `status.warnings` and `status.deprecation`.

Let's add support for dotted paths so the set can name nested fields (`frozenset({"tests", "status.disable_codecov_badge"})`), threading a path prefix through `_merge_known_fields`.

ecosystem-automation/collector-watcher/src/collector_watcher/metadata_parser.py
```diff
@@ -114,31 +133,44 @@ def _sanitize_description(description: str) -> str:
...
+        unrecognized or future schema fields are never silently dropped.
+        `exclude` names fields to drop entirely rather than pass through.
+        Output keys are sorted alphabetically so the same input always
+        produces the same key order, independent of the source YAML's field
+        order, keeping the content-addressed registry output deterministic.
+        """
+        merged = {**source, **normalized}
+        for key in exclude:
+            merged.pop(key, None)
+        return dict(sorted(merged.items()))
```
jaydeluca: The sort is one level deep at each of the four call sites, so nested maps still serialize in upstream file order. Concretely, these are not yet order-stabilized: `events` keys, `status.codeowners`, keys inside `entities`/`feature_gates` list elements, `config` at any depth, and the `sum`/`gauge`/`histogram` descriptor bodies.

Recursing handles all of them, and crucially it must **not** touch sequence order — some upstream lists are order-significant (`histogram.bucket_boundaries`):

```suggestion
        return MetadataParserV1._sorted_deep(merged)

    @staticmethod
    def _sorted_deep(value: Any) -> Any:
        """Recursively key-sort mappings at any depth.

        Sequence *order* is never changed - some upstream lists are
        order-significant (e.g. `histogram.bucket_boundaries`); only the keys of
        mappings inside them are sorted.
        """
        if isinstance(value, dict):
            return {k: MetadataParserV1._sorted_deep(v) for k, v in sorted(value.items())}
        if isinstance(value, list):
            return [MetadataParserV1._sorted_deep(v) for v in value]
        return value
```

One thing this deliberately does *not* do: sort the *element order* of `entities` / `feature_gates`. Two components already ship them unsorted upstream (`k8sclusterreceiver` has 17 entities, `sampleentityreceiver` 2), and that order may be intentional grouping, so I'd rather preserve upstream's choice than impose one.

ecosystem-automation/collector-watcher/src/collector_watcher/metadata_parser.py
```diff
@@ -70,40 +70,59 @@ def get_schema_version(self) -> str:
...
         if "resource_attributes" in raw:
-            parsed["resource_attributes"] = self._parse_attributes(raw["resource_attributes"])
+            normalized["resource_attributes"] = self._parse_attributes(raw["resource_attributes"])
+
+        if isinstance(raw.get("telemetry"), dict):
+            telemetry = raw["telemetry"]
+            telemetry_normalized: dict[str, Any] = {}
+            if "metrics" in telemetry:
+                telemetry_normalized["metrics"] = self._parse_metrics(telemetry["metrics"])
+            normalized["telemetry"] = self._merge_known_fields(telemetry, telemetry_normalized)
```
jaydeluca, line 147: `events` has the same per-entry shape as `metrics` in the upstream schema (`enabled`, `description`, `extended_documentation`, `warnings`, `attributes: [string]`), so it could reuse `_parse_metrics` exactly the way `telemetry.metrics` does on the line above:

```python
if "events" in raw:
    normalized["events"] = self._parse_metrics(raw["events"])
```

ecosystem-automation/collector-watcher/tests/test_metadata_parser.py
```diff
@@ -411,11 +474,248 @@ def test_parser_v1_sorted_attributes():
...
+        "type": "test",
+        "tests": {"config": {"endpoint": "localhost:1234"}},
+        "sem_conv_version": "1.9.0",
+    }
+    result = MetadataParserV1().parse(raw)
+    assert "tests" not in result
+    assert result["sem_conv_version"] == "1.9.0"
+
+
+def test_parser_v1_excluded_fields_is_easy_to_extend():
```
jaydeluca: `test_parser_v1_excluded_fields_is_easy_to_extend` asserts `EXCLUDED_FIELDS == frozenset({"tests"})`, which just restates the constant — it can't fail for any reason other than someone intentionally changing the set, and it has to be edited every time we add an exclusion. The behavior is already covered by `test_parser_v1_excludes_tests_node` directly above. I'd drop this one.

ecosystem-automation/collector-watcher/src/collector_watcher/metadata_parser.py
```diff
@@ -70,40 +70,59 @@ def get_schema_version(self) -> str:
...
+    resource_attributes, telemetry.metrics. Every other field present in the
+    source file — `type`, `display_name`, `sem_conv_version`, `config`,
+    `entities`, `events`, `feature_gates`, or anything the upstream mdatagen
+    schema adds later — is passed through verbatim rather than dropped (see
+    `_merge_known_fields`), so new fields do not require a code change here.
+    The same pass-through-unless-normalized rule applies one level down
+    inside `status`, each `attributes`/`resource_attributes` entry, each
+    `metrics` entry, and `telemetry`: only the specific sub-fields that need
+    sanitization or stable ordering are enumerated, and anything else on
+    those objects is preserved as-is. `EXCLUDED_FIELDS` names top-level
```
jaydeluca: While you're in this docstring: it now contradicts two other places in the file that say V1 is immutable — the module docstring at `:23` ("add a new parser class … and **leave V1 untouched**") and the factory docstring at `:226` ("3. **Do NOT modify MetadataParserV1**").

Modifying V1 here is the right call — this is a read-completeness bug fix, not an upstream input-format change — but a future reader hitting those lines will think the rule was broken. One clause distinguishing the two cases would keep the guidance honest, e.g. *"upstream schema version changes → new parser class; parser bug fixes → change V1 and backfill."*

jaydeluca (commented):
Let's change `Fixes #876` to `Part of #876`. This PR is the correct and necessary first layer, but merging it as `Fixes` would auto-close #876 with the issue still present.

And to answer your question directly, yes, please do take the UI work as a follow-up.

vanshika2720:
@jaydeluca Addressed the latest review feedback:

Added dotted-path exclusions, including status.disable_codecov_badge.
Added recursive deterministic key ordering without reordering lists.
Normalized events using the existing metric parser.
Cleaned up the parser documentation around V1 changes and backfills.
Updated and expanded regression coverage.

Validation: 213 collector-watcher tests passed, Ruff check passed, and formatting is clean. Only the parser and its test file were modified.

jaydeluca (approved):
thanks @vanshika2720 !

Issue #809, opened 2026-07-14
[collector]: promote more community-owned collector distributions


TylerHelmuth:
### Component / area

Collector

### Problem this solves

The Collector SIG manages several [distributions](https://github.com/open-telemetry/opentelemetry-collector-releases/tree/main/distributions) of the collector. The ecosystem-explorer promotes Core and Contrib. We don't recommend Contrib be used in production, so it would be nice to at least add the k8s distro to the ecosystem explorer.

### Proposed solution

Add a sections for all the community-owned distributions

jaydeluca:
This is a great idea. We already have the data, so we just need to make some frontend changes.

Right now we're hardcoding the distribution types based on which repo they've come from, but we also have the list of distributions from the metadata, so we just need to shift the way we're presenting the data a bit.

[`src/v1/lib/list-filters.ts#L41`][list-filters] will need to be changed to compute the set of distributions from the loaded components rather than a hardcoded list.

The distribution facet in [`components/list/facet-panel.tsx`][facet-panel] already maps over `DISTRIBUTIONS`, so it'll pick up new values automatically once that set is data-driven.

The distribution cards in [`features/collector/components/collector-explore-landing.tsx`][explore-landing] will need to be
reworked to become dynamic too. 

We should also incorporate a way to augment these pages per distribution with custom content, for example, a section on the contrib distribution indicating that it's not recommended for production use. That pattern already half-exists: `public/locales/en/collector.json` has a `distributions` block keyed by distribution name with per-key `name`/`desc` and a `generic` fallback, and k8s is already in there.  It's used today on the component detail page. So rather than inventing a new mechanism, we can extend that block (e.g. add a `notice`/`productionWarning` field per distribution) and reuse it on the landing/list cards.

And then we'll need to update the filters for searching and lists etc. This is mostly straightforward, with one thing worth calling out because it's the only real behavioral change: filtering currently asks "did this component come from the selected source repo?" and needs to ask "is the selected distribution in this component's `distributions` list?" A component like `k8sclusterreceiver` will be included in both `contrib` and `k8s`, so it should show up under both filters

[list-filters]: https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/blob/main/ecosystem-explorer/src/v1/lib/list-filters.ts#L41
[...]

PR #898, opened 2026-07-26
feat(collector): extend IndexComponent with distributions property

Pittu-Sharma:
## What changed

Added a `distributions` optional field to the `IndexComponent` interface in `src/types/collector.ts`. Updated `toIndexComponent` in `collector-data.ts` to map `component.status?.distributions` into the index shape, and updated the related test fixtures.

## Why

Part of #809. The `status.distributions` array is already present in the raw component data. Exposing it on `IndexComponent` is the prerequisite for the follow-up PRs that will use it to drive the distribution filter and landing page cards dynamically, instead of relying on a hardcoded list.

## How it was tested

- `bun run typecheck` — passed, no TypeScript errors
- `bun run lint` — passed, no ESLint warnings
- `bun run test` — 1298/1299 tests passed (1 pre-existing flaky timeout unrelated to this change)
- `bun run format:check` — all 3 modified files pass Prettier

## Is this a breaking change?

No. The new field is optional (`distributions?: string[]`), so all existing consumers of `IndexComponent` continue to work without changes.

This is PR 1 of 3 for issue #809. PRs 2 and 3 will build on top of this by using `distributions` to drive the filter panel and explore landing page dynamically.

## Checklist


ecosystem-explorer/src/types/collector.ts
```diff
@@ -154,6 +154,7 @@ export interface IndexComponent {
   id: string;
   name: string;
   distribution: string;
+  distributions?: string[];
```
jaydeluca, line 157: it feels strange to have both `distribution` and `distributions`. I think we are using the existing distribution field to identify which repo it comes from, is that right? if so, maybe we could also re-name that one. we could do it in a followup as i imagine it'll touch a bit more code to update

jaydeluca (approved):

PR #911, opened 2026-07-27
fix(collector): render backtick-wrapped description text as code

### LINKED ISSUE #910: Code formatting in collector component descriptions
In the java agent instrumentation descriptions, we render backticks as code elements:

See: https://explorer.opentelemetry.io/java-agent/instrumentation/elasticsearch-api-client-7.16
Image

We need to do the same for collector components, see: https://deploy-preview-898--otel-ecosystem-explorer.netlify.app/collector/components/contrib/apachesparkreceiver?type=receiver&version=0.156.0

Image

MeloveGupta:
## What changed

collector component descriptions now go through the same `renderWithInlineCode` helper that the java agent side already uses, so stuff wrapped in backticks actually shows up as code instead of just... backticks sitting in a paragraph.

## Why

Fixes #910 

java agent instrumentation descriptions already render backticks as code elements, so on pages like the apache spark receiver, you'd see literal `` `/metrics/json` `` in the description. `collector-detail-page.tsx` was just rendering `component.description` raw, while `instrumentation-detail-page.tsx` was already piping the exact same kind of string through `renderWithInlineCode` from `src/lib/render-inline-code.tsx`. so this just wires collector up to the same utility.

## How it was tested

I have added one new test, for the new test specifically, I actually reverted just the component fix (kept the test) and reran it to make sure it fails without the fix, it did (`getByText` couldn't find `/metrics/json` as its own element since it was buried in one big text block). I put the fix back, reran, it passed.

Screenshot from 2026-07-28 02-03-12

## Is this a breaking change?

No

I found one more spot with the exact same problem: the collector list/card view (`collector-components-page.tsx`) also renders `comp.description` raw, same as the detail page did. The java agent card (`instrumentation-card.tsx`) already uses `renderWithInlineCode` there too, so it's the same asymmetry, just on the card instead of the detail page.
But I can raise a PR for that too... Just let know :)

## Checklist

### DIFF
--- a/ecosystem-explorer/src/features/collector/collector-detail-page.test.tsx
+++ b/ecosystem-explorer/src/features/collector/collector-detail-page.test.tsx
@@ -69,6 +69,11 @@ const mockComponentWithReadme: CollectorComponent = {
   markdown_hash: "abc123def456",
 };
 
+const mockComponentWithBacktickDescription: CollectorComponent = {
+  ...mockComponentWithoutTelemetry,
+  description: "Fetches metrics via the `/metrics/json` endpoint.",
+};
+
 function renderAtRoute(path: string) {
   return render(
     <MemoryRouter initialEntries={[path]}>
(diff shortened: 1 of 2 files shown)


jaydeluca (approved):
thanks!

PR #913, opened 2026-07-28
feat(collector): dynamic distribution filtering in the v1 list view

Pittu-Sharma:
## What changed

Replaces the hardcoded `["core", "contrib"]` distribution allowlist with
an open string type and a URL-safe parser that accepts any value the
index contains. The v1 list page now filters components by checking
`IndexComponent.distributions` (the full list a component belongs to)
instead of the single `distribution` field, and facet counts are
accumulated across all distributions a component appears in.

## Why

Part of #809. The index data model already supports a `distributions`
array per component; the filter layer was discarding that information and
only ever matching against the primary `distribution` field. This PR
wires up the data correctly so new distributions in the registry are
automatically surfaced without a code change.

## How it was tested

- `bun run typecheck` — no errors
- `bun run lint` — no errors
- `bun run test --run` — 1298/1299 tests pass; the single timeout is in
  `java-instrumentation-list-page.test.tsx` and is a pre-existing flake
  that passes when run in isolation (`bun run test
  src/features/java-agent/java-instrumentation-list-page.test.tsx
  --run`). It is unrelated to this change.

## Is this a breaking change?

No.

This PR is the second of three for #809. PR-1 (already merged) added
`distributions?: string[]` to `IndexComponent` and exposed it from
`collector-data.ts`. PR-3 (pending) will update the landing page UI and
locale files to render per-distribution notices dynamically.

The `DISTRIBUTIONS` constant is kept but extended and marked
`@deprecated`; it is not used for filtering anymore.

## Checklist


jaydeluca:
just noting that there's some merge conflicts that need to be fixed up

Pittu-Sharma:
> just noting that there's some merge conflicts that need to be fixed up

Hi @jaydeluca, I have fixed the merge conflict issue. Please review the PR once.

ecosystem-explorer/src/v1/lib/list-filters.ts
```diff
@@ -86,7 +86,11 @@ export const STABILITIES: readonly StabilityFacet[] = [
...
   "unmaintained",
 ];
-export const DISTRIBUTIONS: readonly Distribution[] = ["core", "contrib"];
+export const DISTRIBUTIONS: readonly Distribution[] = ["core", "contrib", "k8s", "otlp"];
+/** @deprecated `Distribution` is now an open string type. This list covers the four
+ * distributions shipped with the official releases and is kept for backwards
+ * compatibility with any callers that render a static list of options. New distributions
+ * are accepted in the URL and will affect filtering/counts when present in the index. */
 const SORTS: readonly SortMode[] = ["name", "updated", "stability"];
 const DENSITIES: readonly DensityMode[] = ["cards", "compact", "table"];
```
jaydeluca, line 91: a couple things:

- This is the V1 list page, but we should really focus on the legacy one first
- This deprecations comment is above the SORTs variable, not the distribution value
- I dont think we want to fully remove the information in the previous distribution field, we proably just want to rename it to something indicating which code base it comes from

Pittu-Sharma: Hi @jaydeluca, I have resolved the above mentioned issues. Please review it once and if any more changes are required please let me know!

jaydeluca (changes requested):
hey @Pittu-Sharma were you able to test this locally? Your PR description makes it sound like this should be updated, but I'm not seeing the K8s distribution end up in the dropdown, am I missing something?

image

Also, i had mentioned that we should focus on the existing pages and leave the V1 pages alone for now ([comment](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/913#discussion_r3703937748)), can you revert the pages under `/V1` for now? the only way to get a deploy preview for changes in that code is to use a custom branch name, so i'd like to keep those changes separate.

before re-requesting my review, please test the code to make sure it does what its supposed to. thanks.

Pittu-Sharma:
Hey @jaydeluca, thanks for the review!

I've pushed an update to address your feedback:

1. Reverted V1 Changes: I've completely reverted all changes under the src/v1/ directory and the associated locale files back to their main branch state to keep them separate for now.

2. Regarding the K8s Distribution in the Dropdown: I investigated this thoroughly and the frontend filtering logic in this PR is actually working correctly! It looks for the distributions array (which contains "k8s") and falls back to the singular distribution string. However, the reason K8s doesn't appear in the dropdown is due to the current JSON bundle data generated by the backend (bundles/[version]-[hash].json). Currently, the bundle files only expose the singular distribution field ("core" or "contrib") and do not include the distributions array. Since the array is only available in the individual component detail JSONs (which aren't loaded for the list view), the frontend has no data to render "k8s" in the dropdown yet.

The dynamic filtering logic in this PR is ready to go, and the K8s option will naturally appear as soon as the data pipeline is updated to include the distributions array in the list bundles.

Pittu-Sharma:
@jaydeluca, Please review the PR once and let me know if any changes are required.

jaydeluca (approved):

PR #925, opened 2026-07-31
fix(collector): improve distribution information on the collector detail page

### LINKED ISSUE #921: [Explorer] Fix up the collector detail page
Right now we have distribution information under the "stability" tab, which isn't intuitive. I think we should move that information into the "Details" Tab.

We also have what looks like contradictory information, where on the details page we show a single disritubution value, but what it really indicates is the code base its in. We should update that.

For example, see https://explorer.opentelemetry.io/collector/components/core/memorylimiterprocessor?type=processor&version=0.156.0

details tab says core:

Image

Stabilty tab shows distribution availability:

Image

This is confusing so we should clean it up and make it more intuitive

Adii-45:
## Summary

This PR improves the Collector component detail page by making distribution information easier to discover and reducing confusion between the **Details** and **Stability** tabs!

### Changes

- Move the **Distribution Availability** section from the **Stability** tab into the **Details** tab
- Rename the ambiguous **Distribution** field to **Source Repository** to better reflect what the value represents
- Add a short helper description explaining that the field indicates the upstream repository the component originates from
- Keep the **Stability** tab focused solely on stability levels and the stability legend
- Update locale strings and tests accordingly

## Why?

Previously, distribution availability was hidden under the **Stability** tab even though it describes where a component is packaged rather than its stability level!

Additionally, the **Distribution** field in the Details tab referred to the component's source repository (e.g. Core or Contrib), while the Stability tab showed actual distribution availability. Since both used the same terminology, it was easy to interpret them as representing the same information!

This change makes the page more intuitive by grouping related information together and using clearer terminology...!

Fixes #921

### DIFF
--- a/ecosystem-explorer/public/locales/en/collector.json
+++ b/ecosystem-explorer/public/locales/en/collector.json
@@ -179,7 +179,8 @@
     "labels": {
       "type": "Type",
       "version": "Version",
-      "distribution": "Distribution"
+      "sourceRepository": "Source Repository",
+      "sourceRepositoryHint": "The Collector repository this component's code lives in, not the distributions that package it."
     },
     "links": {
       "sourceCode": "Source Code",
--- a/ecosystem-explorer/public/locales/es/collector.json
+++ b/ecosystem-explorer/public/locales/es/collector.json
@@ -179,7 +179,8 @@
     "labels": {
       "type": "Tipo",
       "version": "Versión",
-      "distribution": "Distribución"
+      "sourceRepository": "Repositorio de origen",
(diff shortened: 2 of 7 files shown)


Adii-45:
@jaydeluca All Cl checks are now passing and copilot review comments have been addressed - this PR is now ready for review!

Satya900:
@Adii-45 Just a small suggestion: maintainers are automatically notified when a PR is opened, so there's usually no need to tag them unless there's something specific that requires their attention.

jaydeluca (approved):
LGTM, thanks!

PR #941, opened 2026-08-03
feat(collector): add internal telemetry version comparison

### LINKED ISSUE #877: [Collector] Telemetry comparison by version
This is blocked by #876

We want to have the same functionality of the comparison tool in the java agent pages for the collector components.

Image

There aren't `when` conditions in the collector metadata, so we likely don't need to incorporate the "configuration" axis for comparison, but we should double check.

vanshika2720:
Collector metadata.yaml can define a `telemetry` block describing internal self-observability metrics a component emits about its own operation (e.g. `processor_memory_limiter_refused_spans`), distinct from the existing `metrics` field which describes the signal data a component produces about the monitored system.

Add a per-component "Internal Telemetry" tab (shown only when the field is present) with the same Current/Comparison UX already used for Java Agent instrumentation telemetry: a From/To version picker and diff cards for added, removed, and changed metrics. Collector telemetry has no `when`-condition concept, so unlike the Java Agent version there is no configuration-axis selector.

The `telemetry` field is expected to already be present in the registry data consumed by this UI.

## What changed

Added an "Internal Telemetry" tab to Collector component pages, with Current View and Version Comparison modes. The comparison view allows users to compare internal telemetry metrics between Collector versions and displays added, removed, and changed metrics.

The UI-only changes in this PR depend on the Collector telemetry data being available in the registry/database output.

## Why

Fixes #877

## How it was tested

- Added and updated frontend tests covering the Internal Telemetry tab, telemetry comparison UI, telemetry diffing, metric types, and comparison loading behavior.
- `bun run typecheck` passed.
- `bun run lint` passed.
- Collector feature tests passed: 94/94.
- Manually verified the UI using real Collector registry telemetry data.
- Verified the Current View displays internal telemetry metrics.
- Verified version comparison displays added and removed metrics between Collector versions.
- Verified changed telemetry metrics are displayed with the appropriate before/after diff.
- Verified there were no console or page errors during the manual verification.

## Is this a breaking change?

No

This PR contains the frontend/UI portion of the Collector Internal Telemetry feature.

The Collector DB-builder telemetry passthrough is being handled separately so that the database can be rebuilt independently before this UI is reviewed and previewed with the updated production data.

Collector telemetry does not have a `when`-condition/configuration concept, so no configuration-axis selector is included.

## Checklist

### DIFF
--- a/ecosystem-explorer/public/locales/en/collector.json
+++ b/ecosystem-explorer/public/locales/en/collector.json
@@ -167,7 +167,12 @@
       "details": "Details",
       "stability": "Stability",
       "telemetry": "Telemetry",
-      "readme": "Readme"
+      "readme": "Readme",
+      "internalTelemetry": "Internal Telemetry"
+    },
+    "view": {
+      "current": "Current View",
+      "comparison": "Version Comparison"
     },
     "sections": {
       "componentInfo": "Component Info",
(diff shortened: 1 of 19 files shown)


Satya900:
@vanshika2720   
Really solid work here,  at 19 files this might review more efficiently split into a few smaller PRs, e.g. types+schema first, then the diff utility, then the hook, then the UI components/page integration. Each would be easier to review in isolation.

PR #942, opened 2026-08-03
feat(collector): expose deprecated components

### LINKED ISSUE #931: [Explorer] Expose a category for deprecated collector components
We track collector component deprecations in the registry here:
https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/blob/main/ecosystem-registry/collector/deprecations.yaml

We should add a new category to the explorer UI so that someone can still access information for deprecated components, but with it being made clear on the UI that it is deprecated.

We will need to do this in 2 steps:

1. Update the explorer-db-builder to account for these components. If we can get away with having a minimal index with the last available version info and then a pointer to the other component data as to not duplicate the metadata for these components, that would be ideal. 

 
2. Add the UI elements  - we'll need to add a navigation card to the landing page https://explorer.opentelemetry.io/collector
  - Perhaps we could even have a "version" dropdown option for deprecated on the list page and do something like https://explorer.opentelemetry.io/collector/components?version=deprecated
  - When accessing, we should show the details/readme from the version it last appeared in
  - Deprecated component details should specify the version details of when it was deprecated/removed
  - Some kind of other badge or indicator that the component is deprecated to discourage users from choosing it for something new

pratik50:
## What changed

Added support for deprecated Collector components. The database builder now generates a lightweight deprecations index that references existing component metadata. Both Legacy and V1 UIs include deprecated component discovery, filtering, badges, removal versions, details, and preserved README content from the last available version.

## Why

Fixes #931.

## How it was tested

Added backend and frontend tests covering index generation, data loading, deprecated lists, and detail views. Verified with:

- 447 backend tests
- 1,442 frontend tests
- Prettier, ESLint, TypeScript, and Markdownlint
- Production build
- Manual testing in both Legacy and V1 UIs

## Is this a breaking change?

No.

The deprecations index does not duplicate full component metadata. It stores the existing `component_hash` and version information so details and README content can be loaded from the component’s last available release.

## Checklist

Screenshot 2026-08-03 at 8 26 44 PM
Screenshot 2026-08-03 at 8 26 26 PM
Screenshot 2026-08-03 at 8 27 01 PM
</br> </br> 
Screenshot 2026-08-03 at 8 33 22 PM
Screenshot 2026-08-03 at 8 34 40 PM
Screenshot 2026-08-03 at 8 33 53 PM


pratik50:
@jaydeluca pls review this pr!

ecosystem-explorer/src/features/collector/collector-detail-page.tsx
```diff
@@ -201,6 +240,11 @@ export function CollectorDetailPage() {
                   <GlowBadge variant="muted" className="text-xs tracking-wider uppercase">
                     {component.distribution}
                   </GlowBadge>
+                  {deprecatedView && (
+                    <GlowBadge variant="warning" className="text-xs tracking-wider uppercase">
+                      {t("filters.version.deprecated")}
```
jaydeluca: same as the other comment
```suggestion
                      {t("filters.stability.deprecated")}
```

ecosystem-explorer/src/features/collector/collector-components-page.tsx
```diff
@@ -496,14 +513,27 @@ function CollectorComponentsContent({ urlVersion }: { urlVersion?: string }) {
...
+                              deprecatedView
+                                ? "warning"
+                                : comp.stability === "stable"
+                                  ? "success"
+                                  : "info"
+                            }
                             className="px-2 py-0 text-[9px]"
                           >
-                            {comp.stability}
+                            {deprecatedView ? t("filters.version.deprecated") : comp.stability}
```
jaydeluca: my llm picked up that this is key is actually the plural use in the spanish translation.. which im not sure if we should change, but in the meantime if you can use this `status.deprecated` instead, it should be correct

```suggestion
                            {deprecatedView ? t("filters.status.deprecated") : comp.stability}
```

ecosystem-explorer/src/features/collector/components/collector-explore-landing.tsx
```diff
@@ -150,7 +160,7 @@ export function CollectorExploreLanding() {
     );
   }, [collectorIndex, versionData]);
 
-  const error = indexError ?? versionsError;
+  const error = indexError ?? versionsError ?? deprecationsError;
```
jaydeluca: Chaining `deprecationsError` in here means a failed `deprecations-index.json` replaces the entire Collector landing with an error state, for one count. Same for `deprecationsLoading` in the gate just below (line 178): it adds a third blocking request before first paint.

Could the card degrade instead? Render it without the count (or omit it) when `deprecationsError` is
set, and leave `deprecationsLoading` out of the page-level gate.

ecosystem-explorer/src/features/collector/components/collector-explore-landing.tsx
```diff
@@ -225,6 +235,35 @@ export function CollectorExploreLanding() {
                 </article>
               </Link>
             ))}
+            <Link
+              to="/collector/components/deprecated"
```
jaydeluca: hmm, this emits the path form, while the V1 entry (`src/v1/features/ecosystem/collector-config.tsx:134`) emits the query form (`?version=deprecated`).  I think we should standardize on `?version=deprecated`
```suggestion
              to="/collector/components?version=deprecated"
```

ecosystem-explorer/src/v1/components/list/views.tsx
```diff
@@ -40,13 +40,15 @@ export interface ListRow {
...
   href: string;
+  deprecatedInVersion?: string;
 }
 
 export interface CompactRowProps {
   row: ListRow;
 }
 
 export function CompactRow({ row }: CompactRowProps) {
+  const { t } = useTranslation("collector");
```
jaydeluca: `CompactRow` and `CardView` should be ecosystem-agnostic so we should try and avoid pulling in the collector namespace. 

Can you move `deprecated.removedIn` into the `list` namespace?

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/collector_builder.py
```diff
@@ -40,6 +44,76 @@
...
+    entries: list[dict] = []
+
+    for distribution in DISTRIBUTIONS:
+        by_type = deprecations.get(distribution, {})
+        for component_type in COMPONENT_TYPES:
+            for deprecated in by_type.get(component_type, []):
+                name = deprecated.get("name")
+                component_id = f"{distribution}-{name}" if name else f"{distribution}-<missing-name>"
+                if not name:
+                    raise ValueError(f"Deprecated {distribution}/{component_type} component is missing a name")
```
jaydeluca, line 75: The `<missing-name>` fallback is unreachable since the next two lines raise before `component_id` is used for anything

```suggestion
                name = deprecated.get("name")
                if not name:
                    raise ValueError(f"Deprecated {distribution}/{component_type} component is missing a name")
                component_id = f"{distribution}-{name}"
```

jaydeluca (commented):
visually and functionality look good, some small notes

pratik50:
@jaydeluca Thanks for the review! I’ve addressed all the changes. Could you please take another look?

jaydeluca (commented):
a few more notes, and then can you also fix the merge conflicts?

ecosystem-explorer/src/features/collector/collector-detail-page.tsx
```diff
@@ -184,6 +204,25 @@ export function CollectorDetailPage() {
...
       <div className="mt-3 space-y-6">
+        {deprecatedEntry && (
+          <DetailCard className="border-amber-500/50 bg-amber-500/5">
+            <div className="flex gap-4" role="note">
+              <AlertTriangle
+                className="mt-0.5 h-5 w-5 flex-shrink-0 text-amber-700 dark:text-amber-400"
+                aria-hidden="true"
+              />
+              <div>
+                <h2 className="font-semibold">{t("deprecated.title")}</h2>
```
jaydeluca: This banner renders above `<header>`, so this `<h2>` becomes the first heading on the page and the
component name `<h1>` comes second

```suggestion
                <p className="font-semibold">{t("deprecated.title")}</p>
```

ecosystem-explorer/src/v1/features/detail/detail-page.tsx
```diff
@@ -292,6 +305,22 @@ export function CollectorDetailPageV1() {
         </aside>
 
         <main className="td-detail__main">
+          {deprecatedEntry && (
+            <div className="td-detail-header" role="note">
```
jaydeluca: `.td-detail-header` is the block class that styles the component header card
(`src/v1/styles/detail.css:93`, owned by `src/v1/components/detail/detail-header.tsx`). Reusing it
here renders the removal notice as a second, visually identical neutral card stacked directly above
the real header — the `AlertTriangle` glyph ends up being the only thing distinguishing them. The
legacy UI gets an amber-tinted card for the same content, so the two diverge.

V1 already has the tokens for this: `--color-warning-hsl` / `--color-error-hsl`, used by
`.td-pill--warning` / `.td-pill--deprecated` in `src/v1/styles/recent-activity-rail.css:147`. Could
this get its own `.td-detail-notice` block in `detail.css` built on those, instead of a header class
plus inline Tailwind? The v1 sheet is class-driven per `DESIGN_V1.md`, and it'd also drop the
`flex items-start gap-3` / `td-detail-header__description` borrowing.

While you're there — worth picking one semantic: `.td-pill--deprecated` maps deprecated to
**error/red**, but both banners use **warning/amber** for the same state.

ecosystem-explorer/src/features/collector/collector-detail-page.tsx
```diff
@@ -184,6 +204,25 @@ export function CollectorDetailPage() {
       <BackButton />
 
       <div className="mt-3 space-y-6">
+        {deprecatedEntry && (
+          <DetailCard className="border-amber-500/50 bg-amber-500/5">
```
jaydeluca: `border-amber-500/50 bg-amber-500/5` here and `text-amber-700 dark:text-amber-400` on line 211 are
hardcoded Tailwind palette values, while the `GlowBadge variant="warning"` you added at line 244
resolves the same intent through the token layer. `DESIGN.md` asks for token-driven color so light/
dark parity stays honest.

Reusing `GlowBadge`'s warning palette (or adding a warning-surface token if legacy lacks one) would
keep this consistent with the badge sitting a few lines below it.

ecosystem-explorer/src/features/collector/collector-components-page.tsx
```diff
@@ -153,14 +159,17 @@ function CollectorComponentsContent({ urlVersion }: { urlVersion?: string }) {
 
   const currentVersion = useMemo(() => {
     if (urlVersion) return urlVersion;
+    if (versionQuery) return versionQuery;
```
jaydeluca: This accepts *any* value, not just the `deprecated` sentinel, so `/collector/components?version=0.150.0`
and `/collector/components/0.150.0` now both resolve to the same view with no redirect between them —
while `handleVersionChange` only ever emits the path form for real versions. That's a second,
undocumented URL shape for the list route that SEO and the back button will both see.

```suggestion
    if (versionQuery === "deprecated") return versionQuery;
```

ecosystem-explorer/src/v1/components/list/views.tsx
```diff
@@ -63,6 +65,11 @@ export function CompactRow({ row }: CompactRowProps) {
       </div>
       <div className="td-row__meta">
         <span className="td-row__type">{row.type}</span>
+        {row.deprecatedInVersion && (
```
jaydeluca, line 71: Two small things on the removal-version chip:

1. It's rendered inside `td-row__signals` (and `td-card__signals` at line 119), but it isn't a signal
   list. A `td-row__removed` / `td-card__removed` modifier would keep the stylesheet readable.
2. Placement differs between the two views — here it comes **before** the signals, in `CardView` it
   comes **after** the distribution. Same data, two orders. Worth aligning.

jaydeluca (approved):
thanks @pratik50 great to see you around here again!

PR #945, opened 2026-08-04
feat(explorer-db-builder): pass through collector telemetry metadata

vanshika2720:
Collector metadata.yaml can define a `telemetry` block describing internal self-observability metrics a component emits about its own operation (e.g. `processor_memory_limiter_refused_spans`), distinct from the existing `metrics` field which describes the signal data a component produces about the monitored system. The collector transformer only passed through `attributes` and `metrics`, so `telemetry` was silently dropped before reaching the registry database.

Add the missing passthrough in `collector_transformer.py`: `telemetry` is copied onto the component only when present and non-empty, following the same pattern as the existing `metrics` handling.

## What changed

Added the missing `telemetry` passthrough to the Collector transformer and added tests covering telemetry when present, missing, and empty.

## Why

Ensure Collector internal telemetry metadata from `metadata.yaml` is preserved in the generated registry database so it can be consumed by downstream components.

Related to #877

## How it was tested

- `uv run pytest ecosystem-automation/explorer-db-builder/` — 445 tests passed.
- `uv run ruff check` on the changed Python files — passed.
- `uv run ruff format --check` on the changed Python files — passed.
- Added tests covering:
  - telemetry included when present and non-empty
  - telemetry omitted when missing
  - telemetry omitted when present but empty

## Is this a breaking change?

No

This PR contains only the Collector database-builder changes required to preserve the `metadata.telemetry` field.

The frontend/UI portion of the Collector Internal Telemetry feature is handled separately in PR #941.

Once this PR is merged, the Collector database can be rebuilt so the telemetry data is available to the Internal Telemetry UI.

## Checklist

### DIFF
--- a/ecosystem-automation/explorer-db-builder/src/explorer_db_builder/collector_transformer.py
+++ b/ecosystem-automation/explorer-db-builder/src/explorer_db_builder/collector_transformer.py
@@ -124,6 +124,10 @@ def transform_collector_components(
             if metrics:
                 component["metrics"] = metrics
 
+            telemetry = metadata.get("telemetry")
+            if telemetry:
+                component["telemetry"] = telemetry
+
             if name in readme_map:
                 component["markdown_hash"] = readme_map[name]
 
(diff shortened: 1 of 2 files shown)


vanshika2720:
@jaydeluca Once this PR is merged and the Collector database is rebuilt with the updated transformer, the telemetry metadata will be available in the generated registry data. This will allow the Internal Telemetry UI in PR #941 to display the telemetry tab and version comparison data using the real Collector registry data.

Satya900 (approved):
LGTM..... 

One small unrelated observation, not blocking: resource_attributes has a similar gap in the same transformer, might be worth a follow-up issue sometime.

jaydeluca (approved):

Issue #916, opened 2026-07-29
Support the Go compile-time instrumentation ecosystem


amazingakai:
### What ecosystem area?

Go compile-time instrumentation (`otelc`)

### Key questions that will be answered once the resource is created.

- How should Go compile-time instrumentations be represented in the Ecosystem Explorer?
- How should the Ecosystem Explorer express which instrumentation applies to a given Go dependency?
- How should otelc integrate with the Ecosystem Explorer to consume this metadata?
- How can third-party Go compile-time instrumentations participate in the Ecosystem Explorer?

### What should be created?

Research and define a representation for Go compile-time instrumentations in the Ecosystem Explorer.

The resulting resource should allow `otelc` to discover compatible instrumentations without maintaining an embedded registry, while enabling third-party instrumentation authors to publish instrumentations that `otelc` can consume automatically.

As future work, this could also be extended to support declarative configuration, similar to the existing Java agent integration with the Ecosystem Explorer.

### References (issues, articles, PRs, projects, etc.)

- Previous Registry proposal: https://github.com/open-telemetry/opentelemetry.io/issues/11060
- Registry RFC: https://docs.google.com/document/d/1pXCHgbSNQnk23W26xG3xPtfrtLNFy-NBI_zZXukoYOA/edit?usp=sharing
- Declarative configuration tracking: https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/issues/705

amazingakai:
For some additional context, `otelc` already supports using instrumentations distributed as Go modules by declaring them in an `otel.instrumentation.go` file (see [external-configuration](https://github.com/open-telemetry/opentelemetry-go-compile-instrumentation/blob/main/docs/external-configuration.md#format)).

The idea is for `otelc` to inspect an application's Go module dependencies, consult the Ecosystem Explorer to discover compatible instrumentations, and generate the `otel.instrumentation.go` file automatically.

This would decouple instrumentation discovery from the `otelc` binary, allowing support for new and third-party instrumentations without requiring a new `otelc` release.

mikeblum:
[...]

PR #956, opened 2026-08-06
feat(go-watcher): add support for scraping metadata.yaml across multiple repositories

amazingakai:
## What changed

Refactored the Go instrumentation watcher to support scraping metadata across multiple upstream repositories. Moved the target configuration out to `targets.yaml`, added a new metadata parsing strategy to read natively authored `metadata.yaml` files, and updated the registry schema (supporting a `modules` array, an `installation.methods` array, `otelc_min_version`, and a `hidden` flag).
   
## Why

To allow surfacing `opentelemetry-go-compile-instrumentation` instrumentations in the Ecosystem Explorer.

Part of #916 

## How it was tested

Added unit tests to cover the new metadata.yaml file walker, URL-to-directory resolution, and the new auto-inference logic. Verified locally with go test ./... and did a dry run against a custom branch containing compile-time metadata schemas (see https://github.com/amazingakai/opentelemetry-go-compile-instrumentation/tree/work/amazingakai/add-metadata-yaml)

## Is this a breaking change?

No. Changed the instrumentation.yaml schema (`module` -> `modules`) but the file wasn't generated in the repo yet.

## Checklist


ecosystem-automation/golang-instrumentation-watcher/cmd/watcher/main.go
```diff
@@ -29,6 +28,47 @@ import (
...
+)
+
+type Target struct {
+	URL      string
+	Strategy ParserStrategy
+}
+
+var targets = []Target{
+	{
+		URL:      "https://github.com/open-telemetry/opentelemetry-go-contrib.git",
```
kakkoyun: Maybe we should have a dedicated YAML file checked into the repo for targets.

ecosystem-automation/golang-instrumentation-watcher/cmd/watcher/main.go
```diff
@@ -29,6 +28,47 @@ import (
...
+type ParserStrategy string
+
+const (
+	ParserGoMod    ParserStrategy = "gomod"
+	ParserMetadata ParserStrategy = "metadata"
+)
+
+type Target struct {
+	URL      string
+	Strategy ParserStrategy
```
kakkoyun: We need another for version (main vs latest stable)

amazingakai:
cc @jaydeluca @mikeblum this is ready for review! I'll be adding support for extracting the telemetry blocks via weaver codegen in a follow-up PR.

ecosystem-automation/golang-instrumentation-watcher/cmd/watcher/main.go
```diff
@@ -63,50 +98,87 @@ func main() {
...
+		} else if err := syncVersion(log, target, baseDir, mgr, releaseTag, releaseTag, false); err != nil {
+			return err
+		}
+
+		return syncVersion(log, target, baseDir, mgr, branch, snapshotVersion, true)
 	}
 
-	return syncVersion(log, baseDir, mgr, mainBranch, snapshotVersion, true)
+	// No tags found, so just sync the branch as a snapshot
+	return syncVersion(log, target, baseDir, mgr, branch, "v0.0.1-SNAPSHOT", true)
```
mikeblum: can we either const `"v0.0.1-SNAPSHOT"` or better yet append `-SNAPSHOT` to the branch's current tag? As this might be a v1.2.3 tag so we arrive at `v1.2.3-SNAPSHOT`

ecosystem-automation/golang-instrumentation-watcher/cmd/watcher/main.go
```diff
@@ -63,50 +98,87 @@ func main() {
...
+func syncVersion(log *conf.Log, target Target, baseDir string, mgr *inventory.Manager, ref, version string, snapshot bool) error {
+	repoInfo, err := repo.CheckoutAt(target.URL, baseDir, ref)
 	if err != nil {
 		return err
 	}
 
-	result, err := instrumentation.ScanRepo(repoInfo.Path)
+	var result *instrumentation.ScanResult
+
+	switch target.Strategy {
```
mikeblum, line 177: i like this feature flagging with the Strategy enum

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library.go
```diff
@@ -78,3 +83,67 @@ func analyzeLibrary(goModPath string) (*Library, error) {
...
+	metaFiles, err := WalkMetadata(repoPath)
+	if err != nil {
+		return nil, err
+	}
+
+	// Pre-scan all go.mod files to infer GoMinVersion if blank.
+	modVersions := make(map[string]string)
+	if packages, err := Walk(repoPath); err == nil {
+		for _, pkg := range packages {
+			if mod, err := ParseModule(pkg.GoModPath); err == nil && mod.Path != "" {
```
mikeblum: we're swallowing an err here if we can't read a valid go.mod which is something we should track.

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library.go
```diff
@@ -78,3 +83,67 @@ func analyzeLibrary(goModPath string) (*Library, error) {
...
+				modVersions[mod.Path] = mod.GoVersion
```
mikeblum: should this be the fully-qualified path? is there a chance we're overwriting other files if there are multiple go.mods in a repo such as `otel-go-contrib`

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library.go
```diff
@@ -78,3 +83,67 @@ func analyzeLibrary(goModPath string) (*Library, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		if lib == nil {
+			continue
+		}
+
+		// Infer GoMinVersion from go.mod if not specified in metadata.
+		if lib.GoMinVersion == "" {
+			var highest string
```
mikeblum: can we break this out into its own func on the library module?

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library_test.go
```diff
@@ -90,6 +90,85 @@ func TestAnalyzeLibrary(t *testing.T) {
 	}
 }
 
+// writeMetadata writes content as a metadata.yaml under root/relDir, creating the dir.
+func writeMetadata(t *testing.T, root, relDir, content string) {
+	t.Helper()
+	dir := filepath.Join(root, relDir)
+	if err := os.MkdirAll(dir, 0755); err != nil {
```
mikeblum: `0755` feels like overkill. can we const this and use the file perms API to make this more legible? ex `file, err := os.OpenFile("example.txt", os.O_RDWR|os.O_CREATE, 0644)`

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library_test.go
```diff
@@ -90,6 +90,85 @@ func TestAnalyzeLibrary(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(dir, "metadata.yaml"), []byte(content), 0644); err != nil {
```
mikeblum: same idea here with the `os` perms

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library_test.go
```diff
@@ -90,6 +90,85 @@ func TestAnalyzeLibrary(t *testing.T) {
...
+		setup     func(t *testing.T) string
+		wantLibs  int
+		assertLib func(*testing.T, Library)
+	}{
+		{
+			name: "infers GoMinVersion from go.mod",
+			setup: func(t *testing.T) string {
+				root := t.TempDir()
+				// Setup go.mod with 1.25.0
+				writeGoMod(t, root, "instrumentation/dummy", "module go.opentelemetry.io/otelc/dummy\n\ngo 1.25.0\n")
```
mikeblum: can we make the go version param'd so we can make table tests? like you have below?

ecosystem-automation/golang-instrumentation-watcher/instrumentation/version.go
```diff
@@ -22,11 +28,32 @@ func ModuleVersions(tags []string) map[string]string {
...
+// It also infers library stability if currently unspecified.
 func ApplyModuleVersions(libs []Library, versions map[string]string) {
 	for i := range libs {
-		rel := strings.TrimPrefix(libs[i].Module.Path, otelContribPrefix)
-		if v, ok := versions[rel]; ok {
-			libs[i].Module.Version = v
+		for j := range libs[i].Modules {
+			modPath := libs[i].Modules[j].Path
+			for prefix, v := range versions {
+				if strings.HasSuffix(modPath, "/"+prefix) {
```
mikeblum: let's use the `filepath` api here

ecosystem-automation/golang-instrumentation-watcher/repo/repo_test.go
```diff
@@ -33,20 +33,20 @@ func setupGitRepo(t *testing.T, tmpDir string, cmds [][]string) {
...
+func TestCloneDirName(t *testing.T) {
 	tests := []struct {
 		name string
 		url  string
 		want string
 	}{
-		{"github ssh url", "git@github.com:open-telemetry/opentelemetry-go.git", "opentelemetry-go"},
-		{"github https url", "https://github.com/open-telemetry/opentelemetry-go.git", "opentelemetry-go"},
-		{"no .git extension", "git@github.com:open-telemetry/opentelemetry-go", "opentelemetry-go"},
+		{"github ssh url", "git@github.com:open-telemetry/opentelemetry-go.git", "open-telemetry-opentelemetry-go"},
```
mikeblum: seems a bit redundant no? the `open-telemetry-opentelemetry` prefix

mikeblum (commented):
looking good! got some nits here and there.

ecosystem-automation/golang-instrumentation-watcher/cmd/watcher/main.go
amazingakai: I've extracted it out to `const snapshotVersion = "v0.0.1-SNAPSHOT"`

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library.go
amazingakai: Agreed, added a check in https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/956/changes/a8990b832393c795604abea47a63ddadee4ee002

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library.go
amazingakai: Isn't mod.Path already fully-qualified gomod path (like `go.opentelemetry.io/otelc/foo/bar`)?

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library.go
amazingakai: Done in https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/956/changes/a8990b832393c795604abea47a63ddadee4ee002

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library_test.go
amazingakai: Isn't 0755 normal for dirs I think? I've tightened it to 744 and moved both fileperm and dirperm to const

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library_test.go
amazingakai: Done in https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/956/changes/a8990b832393c795604abea47a63ddadee4ee002

ecosystem-automation/golang-instrumentation-watcher/instrumentation/version.go
amazingakai: Added `path.Join` (since `filepath.Join` uses OS-specific separators)

ecosystem-automation/golang-instrumentation-watcher/repo/repo_test.go
amazingakai: Agreed, updated in https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/956/changes/a8990b832393c795604abea47a63ddadee4ee002

ecosystem-automation/golang-instrumentation-watcher/instrumentation/version.go
mikeblum: oh right good call.

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library_test.go
mikeblum: you're spot on. would going a step further here make sense? using the `tmp` folder/file API? ex. `dname, err := os.MkdirTemp("", "sampledir")`

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library.go
mikeblum: oh if its got the module name in the path we're good. i thought they might be the relative path.

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library_test.go
mikeblum: my thought being if the watcher runs twice we make the filesystem hermetic between runs. This is new functionality from how this works today to be fair.

mikeblum:
:heavy_check_mark: from me fwiw. i think let's defer https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/956#discussion_r3789498721 to a follow-up PR.

amazingakai:
Hello @jaydeluca, I believe I've addressed all comments from @mikeblum. I'd appreciate a review whenever you have a chance.

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library.go
```diff
@@ -78,3 +83,78 @@ func analyzeLibrary(goModPath string) (*Library, error) {
...
+}
+
+func (l *Library) inferGoMinVersion(modVersions map[string]string) {
+	if l.GoMinVersion != "" {
+		return
+	}
+	var highest string
+	for _, m := range l.Modules {
+		if gv, ok := modVersions[m.Path]; ok && gv != "" {
+			if highest == "" || version.Compare(gv, highest) > 0 {
```
jaydeluca: `go/version.Compare` requires the `go` prefix (`go1.25.0`, not `1.25.0`). `ParseModule` returns
`f.Go.Version`, which is bare — so both arguments are invalid versions, and `go/version` documents
invalid versions as comparing *equal to each other*. `Compare` returns `0` unconditionally, the
`> 0` branch is unreachable, and `highest` keeps whatever the first module contributed.

```go
version.Compare("1.25.0", "1.20.0")  // 0
version.IsValid("1.25.0")            // false
```

Against the real function:

```go
l.Modules = []metadata.Module{{Path: "a"}, {Path: "b"}}
l.inferGoMinVersion(map[string]string{"a": "1.20.0", "b": "1.25.0"})
// got "1.20.0", want "1.25.0"
```

```suggestion
			if highest == "" || version.Compare("go"+gv, "go"+highest) > 0 {
```

`TestScanMetadataRepo` only builds single-module libraries, so it passes for the wrong reason —
worth a two-module case with differing Go versions, since `modules[]` is the point of this PR.

ecosystem-automation/golang-instrumentation-watcher/cmd/watcher/main.go
```diff
@@ -63,50 +82,88 @@ func main() {
...
+		log.Warn("No targets found in configuration", "file", *targetsFile)
 	}
-	snapshotVersion, err := inventory.NextSnapshot(releaseTag)
-	if err != nil {
+
+	for _, target := range config.Targets {
+		inventoryDir := filepath.Join(root, "ecosystem-registry", "go", repo.CloneDirName(target.URL))
+
+		if err := syncRepo(log, target, *baseDir, inventoryDir); err != nil {
+			log.WithErrorMsg(err, "sync failed for target", "url", target.URL)
```
jaydeluca, line 126: The error is logged and the loop continues, then `main` falls off the end and exits 0 — a clone
failure, a `git ls-remote` blip, an unparseable `go.mod`, or an unknown strategy all produce a green
run. This was `os.Exit(1)` before.

It also undercuts `ScanRepo`'s own rationale (`library.go:44-47`): "A scan error is returned rather
than swallowed: a missing subtree or an unreadable go.mod would otherwise produce a smaller
inventory that commits as a spurious 'libraries removed' diff." Propagating it carefully and then
dropping it at the top is the same outcome by another route.

Track a failure across the loop so one bad target still lets the others run, but the run goes red:

```go
failed := false
for _, target := range config.Targets {
	inventoryDir := filepath.Join(root, "ecosystem-registry", "go", repo.CloneDirName(target.URL))
	if err := syncRepo(log, target, *baseDir, inventoryDir); err != nil {
		log.WithErrorMsg(err, "sync failed for target", "url", target.URL)
		failed = true
	}
}
if failed {
	os.Exit(1)
}
```

ecosystem-automation/golang-instrumentation-watcher/cmd/watcher/main.go
```diff
@@ -63,50 +82,88 @@ func main() {
...
+}
+
+func syncRepo(log *conf.Log, target Target, baseDir, inventoryDir string) error {
+	branch := target.Branch
+	if branch == "" {
+		branch = mainBranch
+	}
+
+	releaseTag, err := repo.LatestReleaseTag(target.URL)
+	if err != nil && !strings.Contains(err.Error(), "no release tag found") {
```
jaydeluca: Control flow keyed on an error message string. The producing site is `repo/repo.go:180`:

```go
return "", fmt.Errorf("no release tag found for %s", url)
```

Rewording that message — a one-line change with no obvious blast radius — silently converts
"repository has no tags yet" (the normal state of go-compile-instrumentation, and the reason this
branch exists) into a hard failure for that target. A sentinel makes the coupling explicit:

```go
// repo/repo.go
var ErrNoReleaseTag = errors.New("no release tag found")
return "", fmt.Errorf("%w for %s", ErrNoReleaseTag, url)
```

```suggestion
	if err != nil && !errors.Is(err, repo.ErrNoReleaseTag) {
```

ecosystem-automation/golang-instrumentation-watcher/cmd/watcher/main.go
```diff
@@ -29,6 +30,24 @@ import (
...
+	URL      string         `yaml:"url"`
+	Strategy ParserStrategy `yaml:"strategy"`
+	Branch   string         `yaml:"branch,omitempty"`
+}
+
+type TargetsConfig struct {
+	Targets []Target `yaml:"targets"`
+}
+
+
```
jaydeluca: `gofmt -l .` flags this file — stray double blank line. Didn't realize we don't have go fmt hooked up to pre-commit yet, but we should.

```suggestion
```

ecosystem-automation/golang-instrumentation-watcher/metadata/metadata.go
```diff
@@ -77,11 +77,12 @@ func (t *InstallType) UnmarshalYAML(value *yaml.Node) error {
 type Stability int
 
 const (
-	StabilityExperimental Stability = iota // feature-complete but not yet production-ready
-	StabilityStable                        // stable API, production-ready
+	StabilityUnknown Stability = iota // unspecified, should be inferred
```
jaydeluca: `gofmt -l .` flags this file — the const block lost its alignment when `StabilityUnknown` (a longer
identifier) was prepended.

```suggestion
	StabilityUnknown      Stability = iota // unspecified, should be inferred
	StabilityExperimental                  // feature-complete but not yet production-ready
	StabilityStable                        // stable API, production-ready
```

Also: the doc comment two lines above still says the YAML tokens are `"experimental"` and
`"stable"` — `"unknown"` is now a third.

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library.go
```diff
@@ -78,3 +83,78 @@ func analyzeLibrary(goModPath string) (*Library, error) {
...
+				return nil, fmt.Errorf("failed to parse module path: %s", pkg.GoModPath)
+			}
+
+			modVersions[mod.Path] = mod.GoVersion
+		}
+	}
+
+	var libraries []Library
+	for _, mf := range metaFiles {
+		lib, err := analyzeMetadataLibrary(mf.MetadataPath)
```
jaydeluca, line 122: `WalkMetadata` computes `MetadataFile.Path` — the repo-relative directory, which is exactly what
`source_path` means — and this loop then uses only `mf.MetadataPath` and discards it.
`MetadataFile.Path` is currently dead outside its own test.

`SourcePath` has no `omitempty`, so any `metadata.yaml` that omits `source_path` writes
`source_path: ""` into the registry. The data's already in hand:

```go
if lib.SourcePath == "" {
	lib.SourcePath = mf.Path
}
```

jaydeluca (commented):
LGTM! Four things I'd like resolved before merge — details inline

Also, we'll want to hook this up to a github action to run nightly. let's put it in its own workflow since it's 
different from the other watchers. Feel free to do that in a follow-up PR if you'd like.

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library.go
```diff
@@ -78,3 +83,78 @@ func analyzeLibrary(goModPath string) (*Library, error) {
...
+// it into a [ScanResult].
+func ScanMetadataRepo(repoPath string) (*ScanResult, error) {
+	metaFiles, err := WalkMetadata(repoPath)
+	if err != nil {
+		return nil, err
+	}
+
+	// Pre-scan all go.mod files to infer GoMinVersion if blank.
+	modVersions := make(map[string]string)
+	if packages, err := Walk(repoPath); err == nil {
```
jaydeluca, line 100: A `Walk` failure silently disables all Go-version inference, while a single `ParseModule` failure
four lines down aborts the entire target. Both feed the same optional map, so they should be handled
the same way — given it's only a fallback, warn-and-continue on both reads better than the split.

Separately: `ScanRepo` deliberately scopes to `instrumentation/` and `bridges/`, but this calls
`Walk(repoPath)` at the repo root, so it parses every `go.mod` in the repository — tool modules,
fixture modules, anything. Combined with the hard-fail below, one intentionally-malformed `go.mod`
anywhere in go-compile-instrumentation takes the whole target down.

ecosystem-automation/golang-instrumentation-watcher/instrumentation/version.go
```diff
@@ -22,11 +29,38 @@ func ModuleVersions(tags []string) map[string]string {
...
 func ApplyModuleVersions(libs []Library, versions map[string]string) {
 	for i := range libs {
-		rel := strings.TrimPrefix(libs[i].Module.Path, otelContribPrefix)
-		if v, ok := versions[rel]; ok {
-			libs[i].Module.Version = v
+		for j := range libs[i].Modules {
+			modPath := libs[i].Modules[j].Path
+			var bestPrefix, bestVersion string
+			for prefix, v := range versions {
+				if strings.HasSuffix(modPath, path.Join("/", prefix)) {
```
jaydeluca: This traded exact matching for suffix matching. Before: `TrimPrefix(modPath, otelContribPrefix)` then
an exact map lookup. Now a partial trailing path binds:

```go
versions := map[string]string{"http/otelhttp": "v9.9.9"}
// module go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp → v9.9.9
```

The longest-prefix tiebreak means a correct longer tag wins when one exists, so this probably doesn't
bite go-contrib today — but the guarantee is now "longest segment-aligned trailing match" rather than
"the module's own tag." `Library.SourcePath` already holds the repo-relative path that
`ModuleVersions` keys on; anchoring there restores exactness.

The doc comment at line 29 also needs updating: it still says "sets each [Library]'s Module.Version
… keyed by the module's repo-relative path", which is now neither the field name nor the mechanism.

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library.go
```diff
@@ -78,3 +83,78 @@ func analyzeLibrary(goModPath string) (*Library, error) {
...
+}
+
+func analyzeMetadataLibrary(metadataPath string) (*Library, error) {
+	data, err := os.ReadFile(metadataPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read %s: %w", metadataPath, err)
+	}
+
+	var lib Library
+	if err := yaml.Unmarshal(data, &lib); err != nil {
```
jaydeluca: `yaml.Unmarshal` ignores unknown keys, so a typo in an upstream `metadata.yaml` (`module:` instead of
`modules:`, `instalation:`) yields a silently empty record instead of an error. These files are
authored in a *different repository* by people who won't see this watcher's output until it's
already committed — worth failing loudly:

```go
dec := yaml.NewDecoder(bytes.NewReader(data))
dec.KnownFields(true)
if err := dec.Decode(&lib); err != nil {
	return nil, fmt.Errorf("failed to unmarshal %s: %w", metadataPath, err)
}
```

Related: nothing validates the parsed record. An empty `name` is accepted (and sorts first — `name`
is the identity key downstream), and duplicate names across files are written unchecked. Worth a
hard fail, or skip-with-warning, on both.

ecosystem-automation/golang-instrumentation-watcher/cmd/watcher/main.go
```diff
@@ -63,50 +82,88 @@ func main() {
...
+	if err != nil {
+		log.WithErrorMsg(err, "failed to read targets file", "file", *targetsFile)
 		os.Exit(1)
 	}
-}
 
-func run(log *conf.Log, baseDir, inventoryDir string) error {
-	log.Info("🔭OTel Ecosystem Explorer: Golang 🔭")
+	var config TargetsConfig
+	if err := yaml.Unmarshal(data, &config); err != nil {
```
jaydeluca: Same strictness point as `metadata.yaml`: a mistyped key here is silently ignored. And a mistyped
`strategy` only surfaces inside `syncVersion`, after the clone and checkout have already run —
worth validating the strategy set once, up front, before any network work. We have a workflow to raise these things with other ecosystems so we can then go upstream and fix them. Could add something like that for this if you want, but not a blocker for this PR.

ecosystem-automation/golang-instrumentation-watcher/repo/repo.go
```diff
@@ -47,19 +40,28 @@ func (r RepoInfo) LogValue() slog.Value {
...
-func name(url string) string {
-	name := filepath.Base(url)
-	return strings.TrimSuffix(name, ".git")
+func CloneDirName(url string) string {
+	s := strings.TrimPrefix(url, "https://github.com/")
+	s = strings.TrimPrefix(s, "git@github.com:")
+	s = strings.TrimSuffix(s, ".git")
+
+	if name, ok := strings.CutPrefix(s, "open-telemetry/"); ok {
+		if name, ok = strings.CutPrefix(name, "opentelemetry-go-"); ok {
```
jaydeluca: Both branches return the same value — `CutPrefix` returns the input unchanged when the prefix is
absent, so the inner `if` decides nothing.

```suggestion
		name, _ = strings.CutPrefix(name, "opentelemetry-go-")
		return name
```

ecosystem-automation/golang-instrumentation-watcher/instrumentation/walker_test.go
```diff
@@ -54,3 +54,52 @@ func TestWalk(t *testing.T) {
...
+			name:      "empty directory",
+			wantCount: 0,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tmpDir := t.TempDir()
+			for _, dir := range tt.dirs {
+				dirPath := filepath.Join(tmpDir, dir)
+				if err := os.MkdirAll(dirPath, 0755); err != nil {
```
jaydeluca: `library_test.go` now uses the `dirPerm`/`filePerm` consts per @mikeblum's nit, but this file — new
code in the same package, added in this PR — still uses the literals. Lines 86 and 89.

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library_test.go
```diff
@@ -90,6 +95,76 @@ func TestAnalyzeLibrary(t *testing.T) {
...
+	dir := filepath.Join(root, relDir)
+	if err := os.MkdirAll(dir, dirPerm); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(dir, "metadata.yaml"), []byte(content), filePerm); err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestScanMetadataRepo(t *testing.T) {
```
jaydeluca, line 110: Both cases are about `GoMinVersion` and both are single-module, which is what lets the
`version.Compare` bug through. Worth adding:

- multi-module with differing Go versions
- an authored `source_path` (currently never populated)
- round-trip of the new fields: `hidden`, `otelc_min_version`, `installation.methods`
- a malformed / unknown-field `metadata.yaml`

(`TestApplyModuleVersions_StabilityInference` is good — covering the explicit-value case is the part
usually forgotten.)

ecosystem-automation/golang-instrumentation-watcher/README.md
```diff
@@ -41,15 +42,20 @@ libraries:
...
     target_module: github.com/aws/aws-lambda-go
     go_min_version: 1.25.0
+    library_link: https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/github.com/aws/aws-lambda-go/otellambda
     instrumentation_type: wrapper
     installation:
-      type: wrapper
+      methods:
+        - wrapper
     stability: experimental
+    hidden: false
```
jaydeluca: The field is `yaml:"hidden,omitempty"`, so `false` is never emitted — this sample output can't occur.

```suggestion
```

ecosystem-automation/golang-instrumentation-watcher/cmd/watcher/main.go
```diff
@@ -1,25 +1,26 @@
...
+// instrumentation inventory into the ecosystem registry.
 //
 // It inventories two versions per run: the latest bare release tag and a
 // snapshot of the main branch. The release is skipped when it has already been
 // inventoried.
 //
 // Usage:
 //
-//	watcher [-base-dir dir] [-inventory-dir dir]
+//	watcher [-base-dir dir]
```
jaydeluca: The `-targets` flag is missing from the usage line (the README documents it):

```suggestion
//	watcher [-base-dir dir] [-targets file]
```

And line 4 above — "It inventories two versions per run" — is no longer true: tagless repos get one
snapshot only, and this now loops over N targets.

ecosystem-automation/golang-instrumentation-watcher/metadata/metadata.go
```diff
@@ -105,47 +106,76 @@ func (s *Stability) UnmarshalYAML(value *yaml.Node) error {
 // embeds [Scope], [Module], and [Installation], and carries an
```
jaydeluca:
```suggestion
// embeds [Scope], [Modules], and [Installation], and carries an
```

ecosystem-automation/golang-instrumentation-watcher/instrumentation/version.go
amazingakai: While we can't do a direct `versions[lib.SourcePath]` lookup because a library might have multiple modules with their own tags (e.g. `instrumentation/net/http/client` and `instrumentation/net/http/server` under `SourcePath` `instrumentation/net/http`), we *can* use `SourcePath` as an anchor. I've updated the code to locate `SourcePath` within the module path and extract the exact tag key.

amazingakai:
Thanks @jaydeluca, addressed all comments in https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/956/commits/275782845c77356ab0d9af83f96acb3b6a2dc215 (please lmk if I missed something).

> Also, we'll want to hook this up to a github action to run nightly. let's put it in its own workflow since it's
different from the other watchers. Feel free to do that in a follow-up PR if you'd like.

Let's do this in a follow-up PR, we are currently moving our instrumentations to a contrib repo. We'd like to do this once that is done.

ecosystem-automation/golang-instrumentation-watcher/instrumentation/library_test.go
amazingakai: Re-looking at this, the tests actually are hermetic between runs! The base directory passed into these helpers is instantiated with `t.TempDir()` at the top of every test case, and we just use os.MkdirAll to build the internal tree structure (like instrumentation/foo/bar) inside that temporary directory.

jaydeluca (approved):
lgtm!

Issue #883, opened 2026-07-22
Deduplicate readmes in registry


jaydeluca:
### Component / area

`collector-watcher` and `java-instrumentation-watcher` for code changes, will result in changes to `ecosystem-registry`

### Problem this solves

Readme's are verbose and rarely change, and we are duplicating the content for every release. We should store one file per version of it in the `ecosystem-registry`, and add references to it.

## Collector

We already are creating files with the hash included in the name, but right now we are creating a readme directory per release (ecosystem-registry/collector/contrib/v0.157.1-SNAPSHOT/component_readmes), so update to store them in the root of the distro instead: `ecosystem-registry/collector/contrib/readmes` and add a new `readme` reference in each component's metadata that has the file name

## Java agent

Instead of storing them in ecosystem-registry/java/javaagent/v2.30.0/library_readmes, put them in ecosystem-registry/java/javaagent/library_readmes and similarly, add a readme link in the metadata

### Acceptance criteria

We rebuild the java and collector registries and it deletes all the excess readme's and adds references to all components

The explorer-db-builder will need to be updated to handle this change as well to map the content back to the components

Pittu-Sharma:
Hi @jaydeluca, I would like to work on this issue. Please assign it to me!

Pittu-Sharma:
Hello @jaydeluca , I've gone deep down into the codebase to understand the problem and made an Implementation plan for this issue. I have decided to make 4 PRs for this issue so the it can be reviewed easily. Please go through the plan once and let me know if you like this plan.

### What's the problem?

Right now, every time a new version is released, the README files for each component 
get copied into that version's directory:
- Java: `v{version}/library_readmes/`
- Collector: `v{version}/component_readmes/`

The issue is that READMEs almost never change between releases, so we're storing a ton 
of duplicate content in the registry for no reason.

### What I'm thinking

The simplest fix is to store each README just once, in a shared folder per distribution, 
and name the file based on a hash of its content — so if the same README appears across 
10 versions, it's only stored once. Then each component just references the filename in 
its metadata.

I'd break this into 4 PRs so each one is easier to review:

---

**PR 1 — Move Java agent READMEs to a global folder**

Instead of saving to `v{version}/library_readmes/`, save to a single 
`library_readmes/` folder at the javaagent root. File names stay the same pattern: 
`{safe_name}-{hash}.md`.

The tricky part here is updating `load_library_readme_map` — right now it scans the 
[...]

Issue #242, opened 2026-04-14
Add Library README Support to explorer-db-builder


jaydeluca:
Extend the `explorer-db-builder` to process library README files from the registry and
integrate them into the content-addressed database. This enables the frontend to display
standalone library documentation in instrumentation detail pages.

## Background

The `java-instrumentation-watcher` will extract library README files to the registry after #241 is completed.

The db-builder then needs to:

1. Read README files from the registry
2. Add `markdown_hash` field to instrumentation objects
3. Write markdown files to the public data directory in a content-addressed manner
4. Enable lazy loading in the frontend

### POC Implementation Reference

The POC demonstrates the end-to-end flow:

1. **Data Model**: Library objects have `markdown_hash` field
    - See: [instrumentation-explorer/frontend/src/types.ts](https://github.com/jaydeluca/instrumentation-explorer/blob/main/frontend/src/types.ts)

2. **Data Loading**: Frontend loads markdown lazily
    - Fetches from `${BASE_PATH}/markdown/{id}-{hash}.md`
    - See: [instrumentation-explorer/frontend/src/utils/dataLoader.ts](https://github.com/jaydeluca/instrumentation-explorer/blob/main/frontend/src/utils/dataLoader.ts)

3. **Rendering**: ReactMarkdown component in "Standalone Library" tab
    - See: instrumentation-explorer/frontend/src/LibraryDetail.tsx

### 1. Discover README Files from Registry

For each version being processed, scan the `library_readmes/` directory to discover
available README files:

```python
# Scan: ecosystem-registry/java/javaagent/{version}/library_readmes/
# Files found:
#   apache-httpclient-4.3-a3bae406cfcf.md
#   armeria-1.3-977075160b7e.md
#   ...

# Parse filename to extract library name and hash:
# Format: {library-name}-{hash}.md
# Example: "apache-httpclient-4.3-a3bae406cfcf.md"
#   -> library_name: "apache-httpclient-4.3"
#   -> hash: "a3bae406cfcf"
```

### 2. Augment Instrumentation Objects

When processing libraries, add `markdown_hash` field if a README exists:

```python
# Before
{
  "name": "apache-httpclient-4.3",
  "display_name": "Apache HttpClient",
  "description": "...",
  ...
}

# After
{
  "name": "apache-httpclient-4.3",
  "display_name": "Apache HttpClient",
  "description": "...",
  "markdown_hash": "a3bae406cfcf",  # NEW FIELD
[...]

PR #960, opened 2026-08-08
feat(java-watcher): deduplicate READMEs by storing globally in registry (#883)

Pittu-Sharma:
## What changed

This PR implements PR-1 for issue #883, moving Java agent README files from per-version directories into a single global `library_readmes/` folder in the registry. Each library entry in `instrumentation.yaml` now contains a `readme: <filename>` content-addressed reference. `JavaagentInventoryManager` and `InstrumentationSync` were updated to handle global README persistence, YAML-based lookup, and orphan README cleanup.

## Why

PartOf #883

## How it was tested

- Ran `uv run pytest ecosystem-automation/watcher-common ecosystem-automation/java-instrumentation-watcher` (**98 passed**)
- Ran `uv run pytest ecosystem-automation/explorer-db-builder` (**442 passed**)
- Executed local integration verification testing content-addressing, idempotency, snapshot cleanup, and global directory reads/writes.
- Verified code formatting and linting via `uv run ruff check` and `uv run ruff format --check`.

## Is this a breaking change?

No. Adding the `readme` key to `instrumentation.yaml` is additive and backward compatible.

This is PR 1 of 4 as discussed in #883. Subsequent PRs will address Collector watcher README extraction, DB builder publishing, and frontend UI updates.

## Checklist


ecosystem-automation/watcher-common/src/watcher_common/inventory_manager.py
```diff
@@ -231,93 +232,64 @@ def save_library_readmes(
...
-                    current = selected_readmes.get(library_name)
-                    if current is None:
-                        selected_readmes[library_name] = (markdown_hash, mtime_ns, item.name)
-                    else:
-                        _, current_mtime_ns, current_name = current
-                        if mtime_ns > current_mtime_ns or (mtime_ns == current_mtime_ns and item.name > current_name):
-                            selected_readmes[library_name] = (markdown_hash, mtime_ns, item.name)
-                else:
-                    logger.warning("Malformed README filename in %s: %s", version, item.name)
+        data = self.load_versioned_inventory(version)
```
jaydeluca, line 331: **Blocking.** No registry file has a `readme:` key yet, so this returns `{}` for every existing version. On the next nightly run we'll lose all READMEs.

Let's keep the read path backward compatible, and then remove the fallback after we do the db migration.

```python
def load_library_readme_map(self, version: Version) -> dict[str, str]:
    readme_map = self._readme_map_from_yaml(version)
    if readme_map:
        return readme_map
    # Pre-#883 versions have no `readme:` refs. Remove with the leftover
    # v*/library_readmes/ dirs once every version is migrated.
    return self._legacy_readme_map(version)
```

Same for `load_library_readme_content`: try the global path, then `v{version}/library_readmes/`.

Also drop `st_mtime_ns` when lifting the legacy scan across; tiebreak on sorted filename.

ecosystem-automation/watcher-common/src/watcher_common/inventory_manager.py
```diff
@@ -336,3 +308,40 @@ def _parse_readme_filename(self, filename: str) -> tuple[str, str] | None:
         if match:
             return match.group(1), match.group(2)
         return None
+
+    def prune_orphan_readmes(self) -> int:
```
jaydeluca, line 445: **Blocking. `prune_orphan_readmes` is never called.** Only its definition and one test reference it.

`cleanup_snapshots()` used to delete per-version README dirs; it no longer does, and `update_snapshot()` refetches nightly. Every upstream edit writes a file nothing collects — the registry grows monotonically, defeating the dedup goal.

Wire it into `sync()` and return 0 when `referenced_files` is empty (otherwise it wipes the whole dir), and note it only walks `libraries`, not `custom`.

ecosystem-automation/java-instrumentation-watcher/src/java_instrumentation_watcher/instrumentation_sync.py
```diff
@@ -194,7 +199,16 @@ def _sync_library_readmes(
...
-        written = self.inventory_manager.save_library_readmes(version, fetched)
-        logger.info(f"  Stored {written} library README(s) for v{version}")
+        written_map = self.inventory_manager.save_library_readmes(version, fetched)
+
+        for lib in libraries:
+            if lib.get("name") in written_map:
+                lib["readme"] = written_map[lib["name"]]
+
+        if fetch_errors == 0:
+            instrumentations["readmes_synced"] = True
```
jaydeluca: `readmes_synced` is pipeline state in the registry data artifact — a fifth top-level key, in every diff and content hash, let's get rid of that.

ecosystem-automation/java-instrumentation-watcher/src/java_instrumentation_watcher/instrumentation_sync.py
```diff
@@ -194,7 +199,16 @@ def _sync_library_readmes(
...
 
-        written = self.inventory_manager.save_library_readmes(version, fetched)
-        logger.info(f"  Stored {written} library README(s) for v{version}")
+        written_map = self.inventory_manager.save_library_readmes(version, fetched)
+
+        for lib in libraries:
+            if lib.get("name") in written_map:
+                lib["readme"] = written_map[lib["name"]]
+
+        if fetch_errors == 0:
```
jaydeluca: `readmes_synced` is set only when `fetch_errors == 0`, so a *persistent* failure (404 on a moved blob, repeated 403) refetches all ~60 READMEs every night forever. AGENTS.md flags rate limits here. Track failed names so retries are scoped, or add a max-attempts counter.

ecosystem-automation/watcher-common/src/watcher_common/inventory_manager.py
```diff
@@ -231,93 +232,64 @@ def save_library_readmes(
...
-                )
-
+        for lib in libraries:
+            readme_file = lib.get("readme")
+            if readme_file:
+                parsed = self._parse_readme_filename(readme_file)
+                if parsed:
+                    _, markdown_hash = parsed
+                    safe_name = self._sanitize_name(lib["name"])
+                    readme_map[safe_name] = markdown_hash
```
jaydeluca: Keyed by sanitized name, but `explorer_db_builder/main.py:221` looks up the raw `item["name"]` — works only because Java names sanitize to themselves. This is the #242 mismatch. The old code had to sanitize (it parsed filenames); this reads the real name from YAML, and `load_library_readme_content` sanitizes internally anyway.

```suggestion
                    readme_map[lib["name"]] = markdown_hash
```

ecosystem-automation/watcher-common/src/watcher_common/inventory_manager.py
```diff
@@ -231,93 +232,64 @@ def save_library_readmes(
...
-                    version,
-                    selected_name,
-                    markdown_hash,
-                    sorted(hashes),
-                )
-
+        for lib in libraries:
+            readme_file = lib.get("readme")
+            if readme_file:
+                parsed = self._parse_readme_filename(readme_file)
```
jaydeluca, line 342: A `readme:` value that fails `_parse_readme_filename` is now dropped silently; the old code logged `"Malformed README filename"`. Add a `logger.warning` with the version and value.

ecosystem-automation/watcher-common/src/watcher_common/inventory_manager.py
```diff
@@ -231,93 +232,64 @@ def save_library_readmes(
...
+                    _, markdown_hash = parsed
+                    safe_name = self._sanitize_name(lib["name"])
+                    readme_map[safe_name] = markdown_hash
         return readme_map
 
     def load_library_readme_content(self, version: Version, library_name: str, markdown_hash: str) -> str | None:
         """
         Load the content of a specific library README.
 
         Args:
```
jaydeluca, line 413: `version: ... (ignored, readmes are global)` — with the fallback suggested on `load_library_readme_map` this becomes wrong, so the docstring should describe the fallback. `save_library_readmes`'s `version` (line 233) is unused either way.

ecosystem-automation/java-instrumentation-watcher/tests/test_instrumentation_sync.py
```diff
@@ -287,16 +311,25 @@ def test_process_latest_release_backfills_missing_readmes(self, mock_client, inv
...
         assert result is None
-        readme_dir = inventory_manager.get_version_dir(version) / "library_readmes"
-        assert readme_dir.exists()
+        # READMEs should now be in the global dir after backfill
+        global_readme_dir = inventory_manager.inventory_dir / "library_readmes"
+        assert global_readme_dir.exists()
+        # instrumentation.yaml should have been updated with readme fields
+        loaded = inventory_manager.load_versioned_inventory(version)
+        libs = loaded.get("libraries", [])
+        assert any(lib.get("readme") for lib in libs)
```
jaydeluca: `any(...)` passes when only one of two libraries got a README — partial-fetch is exactly what this PR changed.

```suggestion
        assert all(lib.get("readme") for lib in libs)
        assert inventory_manager.readmes_synced(version)
```

Same at line 268.

ecosystem-automation/java-instrumentation-watcher/tests/test_inventory_manager.py
```diff
@@ -342,77 +359,76 @@ def test_parse_readme_filename(self, inventory_manager):
...
-
+        inventory_manager.save_versioned_inventory(
+            version=version,
+            instrumentations={
+                "file_format": 0.1,
+                "libraries": [{"name": "mylib"}, {"name": "other-lib"}],
+            },
+        )
         readme_map = inventory_manager.load_library_readme_map(version)
+        assert readme_map == {}
```
jaydeluca, line 465: `assert readme_map == {}` for libraries with no `readme:` field — the shape of every current registry file. This locks in the behavior flagged on `load_library_readme_map` as correct; it needs to invert with the fallback.

Broader gap: no test proves READMEs reach the database. The db-builder tests mock `inventory_manager` (`test_main.py:271`), so they stay green while the real pipeline returns `{}`. A fixture in production registry shape would have caught this (same as #644/#673).

ecosystem-automation/java-instrumentation-watcher/tests/test_instrumentation_sync.py
```diff
@@ -337,8 +385,9 @@ def test_resolve_ref_failure_does_not_abort_sync(self, mock_client, inventory_ma
 
         assert version == Version("2.10.0")
         assert inventory_manager.version_exists(version)
-        readme_dir = inventory_manager.get_version_dir(version) / "library_readmes"
-        assert not readme_dir.exists()
+        # Global readme dir should NOT exist when readme fetch fails entirely
+        global_readme_dir = inventory_manager.inventory_dir / "library_readmes"
+        assert not global_readme_dir.exists()
```
jaydeluca: `assert not global_readme_dir.exists()` holds only because the fixture is a fresh `tmp_path`. Now that the dir is shared, assert the specific filenames are absent.

ecosystem-automation/java-instrumentation-watcher/tests/test_inventory_manager.py
```diff
@@ -342,77 +359,76 @@ def test_parse_readme_filename(self, inventory_manager):
...
+        # Write the file in the GLOBAL readmes dir (not version-specific)
+        global_readme_dir = inventory_manager.inventory_dir / "library_readmes"
+        global_readme_dir.mkdir(parents=True, exist_ok=True)
+        (global_readme_dir / f"{sanitized_name}-{markdown_hash}.md").write_text("safe content")
+
+        version = Version("1.0.0")
         content = inventory_manager.load_library_readme_content(version, library_name, markdown_hash)
         assert content == "safe content"
+
+    def test_prune_orphan_readmes(self, inventory_manager):
```
jaydeluca: Missing the destructive case: no version references any README (today's registry state) currently deletes every file in the global dir.

Pittu-Sharma:
Hi @jaydeluca, I've addressed all your review comments. Here's a summary of what was done:

### Blocking issues fixed:
1. **Backward-compatible `load_library_readme_map`** – Added `_readme_map_from_yaml` and `_legacy_readme_map` helpers. The method now tries the new YAML `readme:` path first, and falls back to scanning the per-version `library_readmes/` directory for pre-#883 versions. This prevents losing READMEs on the next nightly run.
2. **`prune_orphan_readmes` wired into `sync()`** – It's now called at the end of `InstrumentationSync.sync()`. Also added a guard: returns `0` without touching anything when `referenced_files` is empty (i.e., today's pre-migration registry state), so it won't destructively wipe the entire global dir.
3. **`readmes_synced` removed from the instrumentation data dict** – Added a `mark_readmes_synced(version)` method to `InventoryManager` that writes the flag as a separate post-save YAML patch. `_sync_library_readmes` now returns a `bool`, and callers invoke `mark_readmes_synced` after `save_versioned_inventory` to avoid ordering issues.

### Non-blocking issues fixed:
4. **Raw name as key** – Changed `readme_map[safe_name]` → `readme_map[lib["name"]]` in `_readme_map_from_yaml` to fix the #242 mismatch with `explorer_db_builder`.
5. **Warning log for malformed `readme:` values** – Added `logger.warning` with version and value when `_parse_readme_filename` returns `None`.
6. **`load_library_readme_content` backward-compat + docstring** – Now tries the global path first, then falls back to the per-version legacy path. Docstring updated to describe the fallback.
7. **Tests – `any()` → `all()`** – Fixed at both occurrences (lines 268 and 320), with `assert inventory_manager.readmes_synced(version)` added.
8. **Tests – `test_resolve_ref_failure` assertion** – Replaced `assert not global_readme_dir.exists()` with a filename-specific absence check.
9. **Tests – destructive case for `prune_orphan_readmes`** – Added `test_prune_orphan_readmes_no_refs_returns_zero`: verifies that when no version references any README, the method returns `0` and leaves all files intact.

ecosystem-automation/watcher-common/src/watcher_common/inventory_manager.py
```diff
@@ -336,3 +369,62 @@ def _parse_readme_filename(self, filename: str) -> tuple[str, str] | None:
...
+    def mark_readmes_synced(self, version: Version) -> None:
+        """
+        Persist a readmes_synced flag for this version inside its instrumentation.yaml.
+
+        Keeping the flag in the YAML (rather than the in-memory instrumentations dict
+        passed through _sync_library_readmes) avoids polluting every instrumentation
+        diff with a fifth top-level key that is only pipeline bookkeeping.
+        """
+        data = self.load_versioned_inventory(version)
+        data["readmes_synced"] = True
```
jaydeluca: **Blocking.** This still writes the fifth top-level key — moving it out of the in-memory dict into a
second `load`/`save` puts it in the same file:

```yaml
file_format: 0.1
libraries:
- name: mylib
  readme: mylib-abc123def456.md
readmes_synced: true
```

Registry top-level keys today are `custom` / `definitions` / `file_format` / `libraries`, and the
first nightly stamps this onto the already-published `v2.30.0/instrumentation.yaml`. The docstring
above claims the opposite ("Keeping the flag in the YAML ... avoids polluting every instrumentation
diff with a fifth top-level key") — that comment goes too.

Move the state to `ecosystem-registry/java/javaagent/readme-sync-state.json`, git-tracked (it has
to survive between CI runs) and never read by the db-builder. Not a dotfile — dotfiles get missed in
review and skipped by globs.

```python
SYNC_STATE_FILE = "readme-sync-state.json"
MAX_README_FETCH_ATTEMPTS = 3

def _load_readme_sync_state(self) -> dict[str, Any]:
    path = self.inventory_dir / self.SYNC_STATE_FILE
    if not path.exists():
        return {}
    with open(path, encoding="utf-8") as f:
        data = json.load(f)
    return data if isinstance(data, dict) else {}

def _save_readme_sync_state(self, state: dict[str, Any]) -> None:
    path = self.inventory_dir / self.SYNC_STATE_FILE
    path.parent.mkdir(parents=True, exist_ok=True)
    with open(path, "w", encoding="utf-8") as f:
        # sort_keys + trailing newline: registry output must be deterministic (AGENTS.md).
        json.dump(state, f, indent=2, sort_keys=True)
        f.write("\n")

def readmes_synced(self, version: Version) -> bool:
    return bool(self._load_readme_sync_state().get(str(version), {}).get("synced", False))

def record_readme_sync(self, version: Version, failed: dict[str, int]) -> None:
    """failed: library name -> consecutive failed fetch attempts."""
    state = self._load_readme_sync_state()
    state[str(version)] = {"synced": not failed, "failed": failed}
    self._save_readme_sync_state(state)
```

Delete `mark_readmes_synced` and the YAML-backed `readmes_synced` at line 222.

This also settles the retry point I raised last round, which is still open — so please do both here
rather than leaving the file format to change twice. In `_sync_library_readmes`, carry the previous
counts forward: skip any library whose count has reached `MAX_README_FETCH_ATTEMPTS`, increment on
failure, drop the entry on success, then call `record_readme_sync(version, failed)`. That bounds a
permanently-404ing blob to 3 attempts instead of re-fetching all ~60 every night, which
`ecosystem-automation/AGENTS.md` calls out as a live rate-limit concern.

Deriving the flag instead of storing it does not work, in case it's tempting: not every library with
a `source_path` has a README upstream, so an `all(lib.get("readme"))` check is never true and would
refetch forever.

ecosystem-automation/java-instrumentation-watcher/tests/test_inventory_manager.py
```diff
@@ -342,77 +359,110 @@ def test_parse_readme_filename(self, inventory_manager):
...
+        version = Version("1.0.0")
+        inventory_manager.save_versioned_inventory(
+            version=version,
+            instrumentations={
+                "file_format": 0.1,
+                "libraries": [{"name": "mylib"}, {"name": "other-lib"}],
+            },
+        )
         readme_map = inventory_manager.load_library_readme_map(version)
+        # No libraries have a 'readme' field (pre-migration state), and no legacy
```
jaydeluca: **Blocking.** The fallback is the fix for last round's blocking finding and nothing exercises it —
this test still asserts `{}` because the fixture has no legacy directory. Keep this test, but it
isn't coverage of the new path.

Four tests required before merge:

1. `test_load_library_readme_map_falls_back_to_legacy_dir` — no `readme:` fields, **plus** a
   populated `v{version}/library_readmes/` → asserts the legacy map is returned with the expected
   name→hash pairs.
2. `test_load_library_readme_content_falls_back_to_version_dir` — file absent from the global dir,
   present under `v{version}/library_readmes/` → returns the content.
3. `test_sync_prunes_orphan_readmes` — asserts `sync()` actually calls `prune_orphan_readmes`.
   `test_sync_full_workflow` passes with or without the wiring, so it isn't covering it.
4. `test_load_library_readme_map_merges_legacy_and_yaml` — the partially-migrated case from my
   comment on line 321.

Also required: one db-builder test that does **not** mock `inventory_manager`, using a fixture in
production registry shape (`libraries` with no `readme:` + a populated per-version
`library_readmes/`), asserting `markdown_hash` lands on the output. `test_main.py:271` mocks the
manager, which is why that whole suite stayed green through the original regression and would stay
green if the fallback were deleted tomorrow. Same rule as #644/#673.

ecosystem-automation/java-instrumentation-watcher/src/java_instrumentation_watcher/instrumentation_sync.py
```diff
@@ -78,6 +79,10 @@ def sync(self) -> dict[str, Any]:
         summary["snapshot_updated"] = str(snapshot_version)
         logger.info(f"✓ Updated snapshot: {snapshot_version}")
 
+        pruned = self.inventory_manager.prune_orphan_readmes()
```
jaydeluca: **Non-blocking, but fix it in this PR** — it's ~8 lines and the failure mode is a mass delete.

This runs even when `_sync_library_readmes` returned `False` this run. `update_snapshot` re-parses
`instrumentations` fresh every night; if `discover_library_readmes` raises, `_sync_library_readmes`
returns at line 193 before writing any `readme:` keys, so the snapshot YAML is saved without its
refs — and prune then deletes every global file only the snapshot referenced (~60 post-migration).
It self-heals the next night, but the diff is a mass delete followed by a mass re-add, which is the
nightly churn `ecosystem-automation/AGENTS.md` warns about.

Gate it on an instance flag rather than changing the public return types of
`process_latest_release` / `update_snapshot` — those are both tested and used directly:

```python
def __init__(self, ...):
    ...
    self._readme_sync_complete = True

def sync(self) -> dict[str, Any]:
    self._readme_sync_complete = True  # reset per run
    ...
    if self._readme_sync_complete:
        pruned = self.inventory_manager.prune_orphan_readmes()
        if pruned > 0:
            logger.info(f"  Pruned {pruned} orphaned README file(s) from global library_readmes/")
    else:
        logger.warning("  Skipping orphan README prune: README sync was incomplete this run")
```

and set `self._readme_sync_complete = False` on both failure paths in `_sync_library_readmes` (the
`GithubAPIError` return at line 193, and when `fetch_errors` is non-zero at line 226).

The release path is already safe — `process_latest_release` re-saves the loaded inventory, so refs
survive a discovery failure. This is snapshot-only.

ecosystem-automation/watcher-common/src/watcher_common/inventory_manager.py
```diff
@@ -231,93 +232,125 @@ def save_library_readmes(
...
+        pre-#883 versions that have no 'readme' fields yet.  Remove the fallback
+        after every version has been migrated.
+
+        Args:
+            version: Version to scan
+
+        Returns:
+            Dictionary mapping library names to their markdown content hashes
+        """
+        readme_map = self._readme_map_from_yaml(version)
```
jaydeluca: **Non-blocking, but fix it in this PR.** Once *any* library has a `readme:` ref, the legacy scan is
skipped for the whole version. A partial first run (say 58 of 61 fetched) writes 58 refs,
`readmes_synced` stays false, and that day the db-builder sees 58 — the 3 legacy files sitting on
disk are ignored. v2.30.0 is the newest release so `metadata_backfiller` has no forward source, and
those 3 README tabs vanish until the next successful run.

Merge the two maps instead of choosing between them:

```suggestion
        readme_map = self._legacy_readme_map(version)
        readme_map.update(self._readme_map_from_yaml(version))
        return readme_map
```

I considered the argument for keeping either/or — merging means a README deleted upstream, for a
library that still exists, keeps resurrecting from the legacy dir until those dirs are removed. I'm
choosing merge anyway: the failure it prevents (a transient GitHub 403/404 silently dropping
READMEs) is common and user-visible, the failure it introduces is rare and self-clears when the
legacy dirs are deleted in the migration follow-up. Legacy entries for libraries no longer in
`libraries` are harmless — the db-builder only looks up names present in the inventory.

jaydeluca (commented):
Fallback works — I ran the new lookup against the real registry and all 303 READMEs resolve across
all 6 versions (was 0). Prune wiring and the test fixes look right.

Two things block: `readmes_synced` isn't actually out of the registry artifact yet, and the fallback
has no tests. Three more are non-blocking.

ecosystem-automation/watcher-common/src/watcher_common/inventory_manager.py
```diff
@@ -231,93 +232,125 @@ def save_library_readmes(
...
 
+        readme_map: dict[str, str] = {}
+        # Sort filenames for a deterministic tie-break order.
+        for item in sorted(version_readme_dir.iterdir()):
+            if not (item.is_file() and item.suffix == ".md"):
+                continue
+            parsed = self._parse_readme_filename(item.name)
+            if parsed:
+                name, markdown_hash = parsed
+                readme_map[name] = markdown_hash
```
jaydeluca, line 381: **Non-blocking — add the comment, don't change the code.** These keys are **sanitized** names
(recovered from filenames written through `_sanitize_name`), while `_readme_map_from_yaml` now keys
by the raw `lib["name"]` — so one public method returns two different key spaces depending on which
branch ran.

I checked all 261 distinct library names in the registry: none change under `_sanitize_name`, so
this is latent and not worth restructuring for. It does need to be written down, because the next
reader will assume both paths agree. Add exactly this above the loop:

```python
# Keys here are *sanitized* names recovered from filenames, unlike _readme_map_from_yaml
# which keys by the raw name. Equivalent for all 261 current Java library names; would
# diverge for any name containing a character outside [a-zA-Z0-9._-]. Goes away with the
# legacy dirs.
```

ecosystem-automation/watcher-common/src/watcher_common/inventory_manager.py
```diff
@@ -231,93 +232,125 @@ def save_library_readmes(
         self,
         version: Version,
         readmes: Iterable[tuple[str, str]],  # (library_name, content)
-    ) -> int:
-        """Write each README content-addressed. Returns count newly written."""
-        target_dir = self.get_version_dir(version) / self.README_DIR
+    ) -> dict[str, str]:
+        """Write each README content-addressed to global folder. Returns mapping of name to filename."""
```
jaydeluca: **Non-blocking.** `version` (line 233) is unused. Unlike `load_library_readme_content`, the fallback
didn't make it load-bearing again — drop the parameter and update the two call sites in
`instrumentation_sync.py`.

ecosystem-automation/watcher-common/src/watcher_common/inventory_manager.py
```diff
@@ -231,93 +232,125 @@ def save_library_readmes(
...
+
+        Returns:
+            Dictionary mapping library names to their markdown content hashes
+        """
+        readme_map = self._readme_map_from_yaml(version)
+        if readme_map:
+            return readme_map
+        # Pre-#883 versions have no `readme:` refs.  Remove this branch once
+        # every version directory has been migrated.
+        return self._legacy_readme_map(version)
```
jaydeluca: **Non-blocking, no code change — this one's on me, not you.** Nothing migrates v2.27.0–v2.29.0
(there's no `--backfill` mode here like the collector and configuration watchers have), so this
branch and the 364 files under `v*/library_readmes/` stay indefinitely unless something schedules
the cleanup. I'll file the follow-up on #883 rather than expand this PR's scope. For the record it
covers: backfill `readme:` refs into v2.27.0–v2.29.0, delete the per-version `library_readmes/`
dirs, then remove `_legacy_readme_map`, the legacy branch in `load_library_readme_content`, and the
merge in `load_library_readme_map`.

ecosystem-automation/java-instrumentation-watcher/src/java_instrumentation_watcher/instrumentation_sync.py
```diff
@@ -185,16 +205,47 @@ def _sync_library_readmes(
...
+
+        for lib in libraries:
+            if lib.get("name") in written_map:
+                lib["readme"] = written_map[lib["name"]]
+
+        logger.info(f"  Stored {len(written_map)} library README(s) for v{version}")
+
+        self.inventory_manager.record_readme_sync(version, current_failures)
+
+        if current_failures:
```
jaydeluca: The retry cap and the prune gate compose into a permanent stall. Once a library
reaches `_MAX_README_FETCH_ATTEMPTS`, line 219 re-records it in `current_failures` every run, so
this branch is taken forever: `record_readme_sync` never marks the version synced, and
`_readme_sync_complete` never goes back to true.

Both of this PR's goals then stop working, permanently, from one bad blob:

- `prune_orphan_readmes` never runs again — the global dir grows monotonically, which is the dedup
  win #883 is for.
- `readmes_synced(version)` stays false, so `process_latest_release` re-syncs that release every
  night, re-fetching every *healthy* README indefinitely. The cap bounds the failing fetch and
  nothing else — the ~60/night that `ecosystem-automation/AGENTS.md` flags is unchanged.

I ran this branch over five simulated nights, one library permanently 404ing:

```text
night 1: fetch calls=4  synced=False  failures={'akka': 1}  orphan_still_there=True
night 3: fetch calls=4  synced=False  failures={'akka': 3}  orphan_still_there=True
night 4: fetch calls=2  synced=False  failures={'akka': 3}  orphan_still_there=True
night 5: fetch calls=2  synced=False  failures={'akka': 3}  orphan_still_there=True
```

Night 4 onward is the steady state: the two remaining calls are the healthy README re-fetched for
the release and the snapshot, every night, and the orphan never gets collected.

A library that has exhausted its retries is *given up on*, not *pending* — it shouldn't hold either
flag open. Move the constant onto the inventory manager so `record_readme_sync` can apply it, and
gate on retryable failures only:

```python
# watcher_common/inventory_manager.py
MAX_README_FETCH_ATTEMPTS = 3

def record_readme_sync(self, version: Version, failures: dict[str, int]) -> None:
    """Record sync results.

    A version counts as synced once every remaining failure has exhausted its
    retries — those libraries are permanently given up on, so re-running the
    sync buys nothing and only spends rate limit.
    """
    synced = not any(count < self.MAX_README_FETCH_ATTEMPTS for count in failures.values())
```

```python
# instrumentation_sync.py — replace this block
        if any(count < _MAX_README_FETCH_ATTEMPTS for count in current_failures.values()):
            self._readme_sync_complete = False
            return False

        return True
```

Pruning is correct in the given-up case: those libraries got no `readme:` ref written this run, so
their old file genuinely is unreferenced and the db-builder won't look for it.

Please add a test that runs `sync()` past the cap and asserts an orphan *is* collected — the
current `test_readme_sync_retry_limit` stops at "fetch not called", which is exactly the assertion
that passes under both behaviours.

ecosystem-automation/explorer-db-builder/tests/test_main.py
```diff
@@ -502,6 +504,63 @@ def test_writes_ecosystem_stats(self, tmp_path):
...
+        """
+        import tempfile
+
+        with tempfile.TemporaryDirectory() as inv_dir:
+            inventory_manager = InventoryManager(inventory_dir=inv_dir)
+            version = Version("2.10.0")
+
+            readme_content = "# Akka Actor\n\nReal readme content."
+            readme_hash = compute_content_hash(readme_content)
+            readme_filename = f"akka-actor-2.3-{readme_hash}.md"
```
jaydeluca: This seeds the *post*-migration shape — a `readme:` ref plus the global dir. That's the shape no
registry file has yet. The test I asked for covers the pre-migration shape, because that's the
regression it exists to catch: delete `_legacy_readme_map` tomorrow and this test still passes.

Drop the `readme:` key and put the file under `v{version}/library_readmes/` instead:

```python
            legacy_dir = inventory_manager.get_version_dir(version) / "library_readmes"
            legacy_dir.mkdir(parents=True, exist_ok=True)
            (legacy_dir / readme_filename).write_text(readme_content, encoding="utf-8")

            inventory_manager.save_versioned_inventory(
                version=version,
                instrumentations={
                    "file_format": 0.1,
                    # No `readme:` ref — the shape of every registry file today.
                    "libraries": [{"name": "akka-actor-2.3"}],
                },
            )
```

Keeping a second case for the migrated shape is fine, but the legacy one is the one that has to
exist. Same rule as #644/#673.

ecosystem-automation/explorer-db-builder/tests/test_main.py
```diff
@@ -502,6 +504,63 @@ def test_writes_ecosystem_stats(self, tmp_path):
...
         assert data["library_count"] == 3
 
+    def test_run_javaagent_builder_with_real_inventory_manager(self, tmp_path):
+        """Integration test: real InventoryManager + real DatabaseWriter.
+
+        Seeds the inventory with a production-shaped registry (instrumentation.yaml
+        has a 'readme' field, global library_readmes/ contains the file), then
+        asserts that markdown_hash ends up in the DB-writer output — no mocks.
+        """
+        import tempfile
```
jaydeluca: `tmp_path` is already injected — the nested `TemporaryDirectory` isn't buying anything, and
`AGENTS.md` asks for imports at the top of the file. `inventory_dir=str(tmp_path / "registry")`
drops both the import and the `with` block, and un-indents the body.

jaydeluca (commented):
One new blocking issue: the retry cap you added disables orphan pruning permanently, and doesn't
actually stop the nightly re-fetch it was meant to bound. I ran this branch through five simulated
nights with one README permanently 404ing:

```text
night 1: fetch calls=4  synced=False  failures={'akka': 1}  orphan_still_there=True
night 3: fetch calls=4  synced=False  failures={'akka': 3}  orphan_still_there=True
night 4: fetch calls=2  synced=False  failures={'akka': 3}  orphan_still_there=True
night 5: fetch calls=2  synced=False  failures={'akka': 3}  orphan_still_there=True
```

Details on `instrumentation_sync.py:247`. It's a small fix.

Two more that can't take an inline comment:

1. `ecosystem-automation/java-instrumentation-watcher/README.md:21` and `:26` still describe
   READMEs as living in "the version's `library_readmes/` directory". After this PR they go in the
   shared `ecosystem-registry/java/javaagent/library_readmes/`, each library entry carries a
   `readme:` ref, and there's a new `readme-sync-state.json` next to the version dirs. Both
   paragraphs need updating — main just touched the same lines for the JMX index, so rebase first.

2. The branch conflicts with #978 (JMX weaver model scraping), which landed on main on Aug 18 and
   touches the same three files. One trap worth knowing before you resolve it: main's
   `process_latest_release` calls `save_versioned_inventory` and *then* `_sync_library_readmes`,
   while this branch needs the reverse order, because `_sync_library_readmes` now writes the
   `readme:` refs into `instrumentations` in place. Taking main's ordering resolves clean and
   silently saves every new release with no `readme:` keys at all. Main's
   `_sync_jmx_models` calls go after the save on both paths. You'll also need to update main's new
   JMX tests for the `save_library_readmes(fetched)` signature and for `record_readme_sync`
   replacing the old readme-dir seeding.

ecosystem-automation/watcher-common/src/watcher_common/inventory_manager.py
```diff
@@ -219,105 +220,189 @@ def load_versioned_inventory(self, version: Version) -> dict[str, Any]:
...
+
+        if not failures:
+            synced_versions.add(version_str)
+            if "failures" in state and version_str in state["failures"]:
+                del state["failures"][version_str]
+        else:
+            synced_versions.discard(version_str)
+            state.setdefault("failures", {})[version_str] = failures
+
+        state["synced_versions"] = sorted(list(synced_versions))
```
jaydeluca: `synced_versions` accumulates one entry per snapshot version forever. `cleanup_snapshots()` deletes
`v2.30.1-SNAPSHOT/` on the next release bump but leaves `"2.30.1-SNAPSHOT"` in here, so this
git-tracked file grows a stale line every release. Drop entries whose version directory no longer
exists:

```python
        live = {str(v) for v in self.list_versions()}
        state["synced_versions"] = sorted(v for v in synced_versions if v in live)
```

Also worth pruning `state["failures"]` the same way, so a retired snapshot doesn't carry its counts
forward.

(`sorted()` here is string order, so `"2.10.0"` sorts before `"2.9.0"` — deterministic, which is
what `AGENTS.md` requires, just noting it's not version order in case that reads as a bug later.)

PR #966, opened 2026-08-10
Add prHourlyLimit to Renovate configuration

thompson-tomo:
## What changed

Set the hourly pr limit to 10 which is the same as the default concurrent limit.

Note, i would normally suggest to introduce grouping of atleast patches but i see that was previously removed.

## Why

Reduces occurrences of renovate being rate limited from raising pr’s and needing to wait another week before it can raise again.

IMG_0347

## How it was tested

Used in ruby sig repo’s

## Is this a breaking change?

No

## Checklist

- [ ] Tests added or updated for new behavior

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -5,6 +5,7 @@
   internalChecksFilter: "strict",
   labels: ["dependencies"],
   schedule: ["before 8am on Monday"],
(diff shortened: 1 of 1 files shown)


jaydeluca (approved):
thanks!

Issue #241, opened 2026-04-14
Update java-instrumentation-watcher to pull in Library readmes


jaydeluca:
## Summary

Extend the `java-instrumentation-watcher` to extract library README files from the
`instrumentation/*/library/README.md` paths and store them in the `ecosystem-registry`
alongside the instrumentation metadata. This enables displaying standalone library
documentation in the explorer UI.

## Background

The POC (instrumentation-explorer) demonstrates that many Java Agent instrumentations
include `library/README.md` files that document the standalone library usage. These
README files are valuable content that should be:

1. Version-specific (extracted from the same tag/commit as metadata)
2. Content-addressed (to deduplicate across versions)
3. Available in the registry for consumption by explorer-db-builder

### POC Implementation Reference

The POC's [scripts/update-library-readmes.py](https://github.com/jaydeluca/instrumentation-explorer/blob/main/scripts/update-library-readmes.py) demonstrates the extraction pattern:

- Discovers README files at `instrumentation/*/library/README.md` paths
- Uses GitHub API to fetch content at specific commit SHA
- Computes content hash (SHA-256, first 12 chars)
- Stores as `{library-name}-{hash}.md` in shared directory

### 1. README Extraction During Sync

The `InstrumentationSync` class should extract library README files during both release
and snapshot processing:

- Extract from the same ref/commit as the instrumentation.yaml
- Store in `ecosystem-registry/java/javaagent/{version}/library_readmes/`
- Use content-addressed naming: `{library-name}-{content-hash}.md`

### 2. Discovery Logic

Discover README files by:

1. Using GitHub API tree structure (efficient, single API call)
2. Looking for paths matching `instrumentation/*/library/README.md`
3. Extracting library name from path (the directory before `/library/README.md`)

### 3. Content Hashing

- Use SHA-256 hash of the README content
- Take first 12 characters of hex digest (matches POC pattern)
- Hash function should be reusable (e.g., in common module or content_hashing.py)

### 4. Storage Structure

```
ecosystem-registry/java/javaagent/
├── v2.26.0/
│   ├── instrumentation.yaml
│   └── library_readmes/
│       ├── apache-httpclient-4.3-a3bae406cfcf.md
│       ├── armeria-1.3-977075160b7e.md
│       └── ... (more README files)
└── v2.26.1/
    ├── instrumentation.yaml
    └── library_readmes/
        └── ... (version-specific README files)
```

### Code Organization

[...]

PR #978, opened 2026-08-14
feat(java-watcher): scrape JMX weaver models

### LINKED ISSUE #975: Scrape jmx-metrics weaver models into the registry
The JMX instrumentation in the java agent now has weaver models with information about the telemetry emitted:

https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/jmx-metrics/model

We should update the https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/tree/main/ecosystem-automation/java-instrumentation-watcher to scrape these, and then we can figure out how we want to use/display it.

The weaver files are newish, so we need to make sure the watcher accounts for when they don't exist and just ignores them and doesn't fail in the event we run a backfill.

We can extend `InstrumentationSync` to fetch the `instrumentation/jmx-metrics/model/*.yaml` files at the same ref it already uses for `instrumentation-list.yaml`, and persist them content-addressed.

Model files change rarely and are identical across most releases, so store them **once** in a
shared directory at the ecosystem root, with a small per-version index pointing into it:

```
ecosystem-registry/java/javaagent/
├── jmx/                                      # shared, content-addressed model store
│   ├── activemq-6c55bb962678.yaml
│   ├── camel-e74b2ed35478.yaml
│   ├── jetty-a3bae406cfcf.yaml
│   ├── jvm-45b20f5f3f0e.yaml
│   ├── manifest-977075160b7e.yaml
│   ├── tomcat-5846db79cc39.yaml
│   └── wildfly-6d2c8fd10301.yaml
├── v2.29.0/
│   └── instrumentation.yaml                  # no jmx-models.yaml — models don't exist yet
└── v2.30.0/
    ├── instrumentation.yaml
    ├── jmx-models.yaml                       # index for this version
    └── library_readmes/
```

`v2.30.0/jmx-models.yaml`:

```yaml
models:
  activemq: activemq-6c55bb962678.yaml
  camel: camel-e74b2ed35478.yaml
  jetty: jetty-a3bae406cfcf.yaml
  jvm: jvm-45b20f5f3f0e.yaml
  tomcat: tomcat-5846db79cc39.yaml
  wildfly: wildfly-6d2c8fd10301.yaml
manifest: manifest-977075160b7e.yaml
```

Notes on this shape:

- **The keys are JMX target systems, not library names.** Each model basename matches a value accepted by the `otel.jmx.target.system` config option, which corresponds to a rule file in  `instrumentation/jmx-metrics/library/src/main/resources/jmx/rules/`. There is exactly one
  *instrumentation* here (`jmx-metrics`); the models describe its target systems.
- **`manifest.yaml` is not a model** — it is the weaver registry manifest, and it carries the pinned semconv version (`semantic-conventions@v1.43.0`) that the next phase of this project will need to resolve  attribute `ref`s. Store it, but under a separate top-level key so it is never iterated as a target system.
- **The shared-store layout matches where #883 is taking `library_readmes/`.** Please land it this way rather than nesting per-version, so the two converge. If #883 lands first, follow whatever helper it introduces.

pratik50:
## What changed

Scrape JMX weaver model YAML files from Java instrumentation releases and snapshots. Store model content once in a shared content-addressed `jmx/` directory and write a per-version `jmx-models.yaml` index, keeping the manifest separate from target-system models.

  ## Why

  Fixes #975.

  ## How it was tested

  - `uv run pytest ecosystem-automation/java-instrumentation-watcher/` - 102 passed
  - `uv run ruff check ecosystem-automation/`
  - `uv run ruff format --check ecosystem-automation/`
  - Pre-commit hooks passed

  ## Is this a breaking change?

  No.

Missing model directories are ignored for older releases. Incomplete GitHub fetches do not write a partial index, allowing later runs to retry.

  PR #960 overlaps with Java watcher inventory handling. This branch may need rebasing if #960 merges first.

  ## Checklist

  - [ ] Screenshots attached for any UI changes (not applicable; no UI changes)

### DIFF
--- a/ecosystem-automation/java-instrumentation-watcher/src/java_instrumentation_watcher/instrumentation_sync.py
+++ b/ecosystem-automation/java-instrumentation-watcher/src/java_instrumentation_watcher/instrumentation_sync.py
@@ -23,6 +23,7 @@ from semantic_version import Version
 from .instrumentation_parser import parse_instrumentation_yaml
 from .inventory_manager import InventoryManager
 from .java_instrumentation_client import GithubAPIError, JavaInstrumentationClient
+from .jmx_model_extractor import JmxModelExtractor
 from .readme_extractor import ReadmeExtractor
 
 logger = logging.getLogger(__name__)
@@ -38,16 +39,19 @@ class InstrumentationSync:
         client: JavaInstrumentationClient,
         inventory_manager: InventoryManager,
         readme_extractor: ReadmeExtractor | None = None,
+        jmx_model_extractor: JmxModelExtractor | None = None,
     ):
         """
         Args:
             client: GitHub API client for fetching data
             inventory_manager: Inventory manager for storing data
             readme_extractor: README extractor (defaults to ReadmeExtractor(client))
+            jmx_model_extractor: JMX model extractor (defaults to JmxModelExtractor(client))
         """
         self.client = client
         self.inventory_manager = inventory_manager
         self.readme_extractor = readme_extractor or ReadmeExtractor(client)
+        self.jmx_model_extractor = jmx_model_extractor or JmxModelExtractor(client)
 
     def sync(self) -> dict[str, Any]:
         """
@@ -96,6 +100,8 @@ class InstrumentationSync:
             if not self.inventory_manager.readme_dir_exists(version):
                 instrumentations = self.inventory_manager.load_versioned_inventory(version)
                 self._sync_library_readmes(version, tag_string, instrumentations)
+            if not self.inventory_manager.jmx_models_index_exists(version):
+                self._sync_jmx_models(version, tag_string)
             return None
 
         logger.info(f"  Fetching instrumentation list for {tag_string}...")
@@ -107,6 +113,7 @@ class InstrumentationSync:
             instrumentations=instrumentations,
         )
         self._sync_library_readmes(version, tag_string, instrumentations)
+        self._sync_jmx_models(version, tag_string)
 
         return version
 
@@ -153,6 +160,7 @@ class InstrumentationSync:
             instrumentations=instrumentations,
         )
         self._sync_library_readmes(snapshot_version, main_ref, instrumentations)
+        self._sync_jmx_models(snapshot_version, main_ref)
 
         return snapshot_version
 
@@ -198,3 +206,50 @@ class InstrumentationSync:
 
         written = self.inventory_manager.save_library_readmes(version, fetched)
         logger.info(f"  Stored {written} library README(s) for v{version}")
+
+    def _sync_jmx_models(self, version: Version, ref: str) -> None:
+        """Best-effort: fetch JMX weaver model files and write version index."""
+        try:
+            sha = ref if _SHA_RE.match(ref) else self.client.resolve_ref_to_sha(ref)
+            discovered_models, manifest_path = self.jmx_model_extractor.discover_jmx_model_paths(sha)
+        except GithubAPIError as e:
+            logger.warning(f"  JMX model discovery failed for {ref}: {e}")
+            return
+
+        if not discovered_models and manifest_path is None:
+            logger.info(f"  No JMX weaver model files found at {ref}")
+            return
+
+        models: dict[str, str] = {}
+        fetch_failed = False
+        for target_system, path in sorted(discovered_models.items()):
+            try:
+                models[target_system] = self.jmx_model_extractor.fetch_model(path, sha)
+            except GithubAPIError as e:
+                logger.warning(f"  JMX model fetch failed for {target_system}: {e}")
+                fetch_failed = True
+
+        manifest_content = None
+        if manifest_path is not None:
+            try:
+                manifest_content = self.jmx_model_extractor.fetch_model(manifest_path, sha)
+            except GithubAPIError as e:
+                logger.warning(f"  JMX manifest fetch failed: {e}")
+                fetch_failed = True
+
+        if fetch_failed:
+            logger.warning(f"  JMX model index not written for v{version}; fetch incomplete")
+            return
+
+        written_models, manifest_file = self.inventory_manager.save_jmx_models_index(
+            version=version,
+            models=models,
+            manifest=manifest_content,
+        )
+
+        if not written_models and manifest_file is None:
+            logger.info(f"  No JMX model index written for v{version}")
+            return
+
+        suffix = " and manifest" if manifest_file else ""
+        logger.info(f"  Stored {len(written_models)} JMX model file(s){suffix} for v{version}")
--- a/ecosystem-automation/java-instrumentation-watcher/src/java_instrumentation_watcher/instrumentation_sync.py
+++ b/ecosystem-automation/java-instrumentation-watcher/src/java_instrumentation_watcher/instrumentation_sync.py
@@ -160,7 +160,6 @@ class InstrumentationSync:
             instrumentations=instrumentations,
         )
         self._sync_library_readmes(snapshot_version, main_ref, instrumentations)
-        self._sync_jmx_models(snapshot_version, main_ref)
 
         return snapshot_version
 
@@ -246,10 +245,6 @@ class InstrumentationSync:
             models=models,
             manifest=manifest_content,
         )
-
-        if not written_models and manifest_file is None:
-            logger.info(f"  No JMX model index written for v{version}")
-            return
 
         suffix = " and manifest" if manifest_file else ""
         logger.info(f"  Stored {len(written_models)} JMX model file(s){suffix} for v{version}")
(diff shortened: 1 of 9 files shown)


pratik50:
@jaydeluca PTAL

ecosystem-automation/java-instrumentation-watcher/src/java_instrumentation_watcher/instrumentation_sync.py
```diff
@@ -153,6 +160,7 @@ def update_snapshot(self) -> Version:
             instrumentations=instrumentations,
         )
         self._sync_library_readmes(snapshot_version, main_ref, instrumentations)
+        self._sync_jmx_models(snapshot_version, main_ref)
```
jaydeluca: Drop this call — JMX models should be release-only.

The shared store has no GC, so each nightly snapshot orphans a file permanently: this index is the only reference, and `cleanup_snapshots()` removes it without removing the file. Release-only keeps every file in `jmx/` referenced forever.

Two test follow-ons:
- `test_snapshot_sync_writes_readmes` — `resolve_ref_to_sha.call_count` goes back to `2`, and the comment above it needs reverting.
- Add a case asserting `update_snapshot()` writes no `jmx-models.yaml`.

ecosystem-automation/java-instrumentation-watcher/src/java_instrumentation_watcher/instrumentation_sync.py
```diff
@@ -198,3 +206,50 @@ def _sync_library_readmes(
 
         written = self.inventory_manager.save_library_readmes(version, fetched)
         logger.info(f"  Stored {written} library README(s) for v{version}")
+
+    def _sync_jmx_models(self, version: Version, ref: str) -> None:
+        """Best-effort: fetch JMX weaver model files and write version index."""
+        try:
+            sha = ref if _SHA_RE.match(ref) else self.client.resolve_ref_to_sha(ref)
```
jaydeluca, line 212: `_sync_library_readmes` (line 179) already resolved this ref and fetched this tree — a multi-MB, ~10k-entry response. Resolve the SHA once in the caller and pass it into both `_sync_*` helpers, or memoize `fetch_tree` by SHA on the client.

ecosystem-automation/java-instrumentation-watcher/src/java_instrumentation_watcher/inventory_manager.py
```diff
@@ -14,10 +14,75 @@
...
+        manifest: str | None,
+    ) -> tuple[dict[str, str], str | None]:
+        """Persist shared JMX model files and write per-version index.
+
+        Returns:
+            Tuple of (models index map, manifest filename). If nothing was
+            persisted, returns ({}, None) and does not write an index file.
+        """
+        jmx_dir = self.get_jmx_store_dir()
+        jmx_dir.mkdir(parents=True, exist_ok=True)
```
jaydeluca: Create the dir inside `_save_jmx_content_addressed_file` on first write instead — with `models={}, manifest=None` this leaves an empty `jmx/` behind.

Harmless today, but it's the `save_library_readmes` empty-dir bug from #241/#242, which already recurred once.

ecosystem-automation/java-instrumentation-watcher/src/java_instrumentation_watcher/jmx_model_extractor.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self, client: JavaInstrumentationClient):
+        self.client = client
+
+    def discover_jmx_model_paths(self, sha: str) -> tuple[dict[str, str], str | None]:
+        """Return ({target_system: blob_path}, manifest_blob_path_or_none)."""
+        tree = self.client.fetch_tree(sha)
+        models: dict[str, str] = {}
+        manifest_path: str | None = None
+
+        for entry in sorted(tree, key=lambda item: str(item.get("path", ""))):
```
jaydeluca: Iterate `tree` directly — both consumers re-sort (`sorted(discovered_models.items())` in `_sync_jmx_models`, `sorted(models)` in `save_jmx_models_index`) and `manifest_path` is a single value, so sorting ~10k entries here buys nothing.

ecosystem-automation/java-instrumentation-watcher/src/java_instrumentation_watcher/instrumentation_sync.py
```diff
@@ -198,3 +206,50 @@ def _sync_library_readmes(
...
+            logger.warning(f"  JMX model index not written for v{version}; fetch incomplete")
+            return
+
+        written_models, manifest_file = self.inventory_manager.save_jmx_models_index(
+            version=version,
+            models=models,
+            manifest=manifest_content,
+        )
+
+        if not written_models and manifest_file is None:
```
jaydeluca: Unreachable — the early returns at lines 219 and 240 mean `save_jmx_models_index` can't be reached with both empty. Drop it, or keep with a comment marking it defensive.

ecosystem-automation/java-instrumentation-watcher/tests/test_instrumentation_sync.py
```diff
@@ -354,8 +371,82 @@ def test_library_without_source_path_skipped(self, mock_client, inventory_manage
...
+            "jvm-model",
+            "tomcat-model",
+            "manifest-model",
+        ]
+
+        version = sync.process_latest_release()
+
+        assert version == Version("2.30.0")
+        index_path = sync.inventory_manager.get_version_dir(version) / "jmx-models.yaml"
+        assert index_path.exists()
```
jaydeluca, line 405: Assert the index contents — as written this passes even if the manifest leaks into `models`, which is the one shape #975 explicitly rules out.

```python
index = yaml.safe_load(index_path.read_text(encoding="utf-8"))
assert set(index["models"]) == {"jvm", "tomcat"}
assert "manifest" not in index["models"]
assert index["manifest"]
```

ecosystem-automation/java-instrumentation-watcher/tests/test_inventory_manager.py
```diff
@@ -416,3 +416,65 @@ def test_load_library_readme_content_sanitization(self, inventory_manager, tmp_p
...
+        assert first_models["jvm"] == second_models["jvm"]
+        assert len(list(inventory_manager.get_jmx_store_dir().glob("jvm-*.yaml"))) == 1
+
+    def test_save_jmx_models_index_skips_empty_payload(self, inventory_manager):
+        version = Version("2.29.0")
+        models, manifest = inventory_manager.save_jmx_models_index(version=version, models={}, manifest=None)
+
+        assert models == {}
+        assert manifest is None
+        assert not inventory_manager.jmx_models_index_exists(version)
```
jaydeluca, line 480: Also assert the store dir wasn't created — it currently **is** (see the comment on `inventory_manager.py:53`), so add this alongside that fix:

```python
assert not inventory_manager.get_jmx_store_dir().exists()
```

ecosystem-automation/java-instrumentation-watcher/tests/test_instrumentation_sync.py
```diff
@@ -354,8 +371,82 @@ def test_library_without_source_path_skipped(self, mock_client, inventory_manage
...
+        )
+        sync.jmx_model_extractor.fetch_model.return_value = "jvm-model"
+
+        result = sync.process_latest_release()
+
+        assert result is None
+        assert inventory_manager.jmx_models_index_exists(version)
+
+    @pytest.mark.parametrize("failed_file", ["model", "manifest"])
+    def test_release_sync_does_not_write_partial_jmx_index(self, sync, mock_client, inventory_manager, failed_file):
```
jaydeluca, line 434: Add a case for `discover_jmx_model_paths` raising `GithubAPIError` — `fetch_tree` failing or truncating leaves `instrumentation_sync.py:215-217` untested. `test_resolve_ref_failure_does_not_abort_sync` only covers `resolve_ref_to_sha`.

jaydeluca (commented):
One change before merge: **don't sync JMX models for snapshots — releases only.**

The shared store has no GC. Every nightly run writes any changed model into `jmx/`, referenced only by that snapshot's index, and `cleanup_snapshots()` then deletes the index but not the file — one permanent orphan in Git per model change. `library_readmes/` avoids this only because it lives inside the version dir and dies with the `rmtree`.

Restricting JMX to releases removes the problem outright: release indexes are never deleted, so every file in `jmx/` stays referenced and no pruning is needed. Models rarely move between releases anyway. Drop the `_sync_jmx_models` call in `update_snapshot()` and adjust the two tests noted inline.

---

One item that can't take an inline comment because the file isn't in the diff:

`java_instrumentation_client.py:97` — the truncation error still reads `f"Tree at {sha} was truncated; README discovery aborted"`. JMX discovery is now a second caller, so a truncated tree during JMX sync reports a README failure. Drop the caller-specific clause and let each `_sync_*` wrapper supply its own context.

Everything else is optional polish. Fetch-all-then-persist in `_sync_jmx_models` is the right ordering, and the parametrized partial-failure test asserting observable outcomes rather than internal flags is a good test.

pratik50:
@jaydeluca PTAL!

jaydeluca (approved):
thanks!

PR #989, opened 2026-08-18
Fix Release Comparison dropdown and badge legibility

yugeshweb:
## Summary
Two small styling fixes in the Java Agent Release Comparison view.

Version dropdowns were unreadable when open. The "From" and "To" selects used a near-transparent bg-primary/5 tint, but browsers render the native <option> popup on their own default surface regardless, so the open list showed as plain white with poor contrast. The control now uses an opaque bg-card, and each <option> sets bg-card text-foreground to follow the active theme.

The metric/span/config badges sized to their content, so "1 Metric" and "12 Metrics" rendered at different widths. They now share a common min-w with centered, tabular-figure text, min-w rather than a fixed w so an unusually large count (104 Spans)  grows instead of clipping.

## Screenshots
Before

After

Before

After

## Checklist

### DIFF
--- a/ecosystem-explorer/src/features/java-agent/components/release-comparison/instrumentation-diff-card.tsx
+++ b/ecosystem-explorer/src/features/java-agent/components/release-comparison/instrumentation-diff-card.tsx
@@ -45,6 +45,13 @@ const STATUS_CONFIG: Record<
   },
 };
 
+/*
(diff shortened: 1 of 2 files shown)


jaydeluca (approved):
thanks @yugeshweb !

PR #992, opened 2026-08-19
Add go fmt to pre-commit, add CI jobs to run tests and build

jaydeluca:


maryliag (approved):
lgtm, but make sure any suggestion from copilot worth doing it are added

PR #993, opened 2026-08-19
fix: filter non-component packages by status class

### LINKED ISSUE #991: Collector scanner catalogs internal interface packages as real components
## Problem

`collector-watcher`'s `ComponentScanner` currently discovers the following internal interface packages from `opentelemetry-collector` as if they were real, deployable Collector components:

- `receiver/xreceiver`
- `exporter/xexporter`
- `processor/xprocessor`
- `connector/xconnector`
- `extension/xextension`

These packages declare `status.class: pkg` in their upstream metadata, which is distinct from the actual component classes (`receiver`, `exporter`, `processor`, `connector`, and `extension`). They are interface packages used internally by the Collector and are not intended to be built or presented as standalone user-facing components!

The scanner currently determines whether a directory is a component based primarily on its path/name and does not cross-check the upstream `status.class` metadata before registering it...

## Observed behavior

All five packages are currently present in the generated registry/database and appear in the Explorer's corresponding component catalogs with empty metadata such as `display_name: null` and `description: null`..!

This is also present across the tracked core registry versions, rather than being limited to a single generated artifact!

## Expected behavior

Internal/interface packages should not be registered or exposed as user-facing Collector components in the Explorer..

## Additional context

There are already similar exclusions for other internal `pkg` packages in `component_scanner.py`, but these five packages are currently not covered!

The existing `collector_display_name_audit.py` explicitly excludes these IDs from its audit, but that only suppresses the resulting audit warning; the packages are still scanned, registered, and exposed through the generated registry..

The appropriate filtering mechanism (explicit exclusions vs. using the upstream `status.class` metadata) can be determined as part of the fix..

Adii-45:
## What changed

Updates `ComponentScanner` to use the upstream `metadata.status.class` when determining whether a scanned package is a valid Collector component...

Packages whose declared `status.class` does not match the component type being scanned are now excluded, while packages with missing metadata or missing `status.class` continue to fail open and retain the existing behavior!

This prevents internal interface packages such as `xreceiver`, `xexporter`, `xprocessor`, `xconnector`, and `xextension` from being cataloged as user-facing Collector components..!

## Why

Closes #991

The scanner previously determined component inclusion before parsing the upstream metadata, so these internal interface packages could pass the directory-based checks and be registered as real components...

This change follows the metadata-driven approach discussed in the issue rather than adding another hardcoded package exclusion list.

## How it was tested

- Added regression coverage for:
  - `status.class` mismatch → excluded
  - matching `status.class` → included
  - missing `status.class` → included
  - missing `status` → included
  - nested components and their parent component types
  - all five affected interface package types
- `pytest ecosystem-automation/collector-watcher/tests/test_component_scanner.py -q`
- Full `ecosystem-automation` Python test suite: **898 passed**
- Ruff lint and format checks pass

## Scope

- `component_scanner.py`
- `test_component_scanner.py`

Registry backfill/generated data is intentionally not included in this PR!

### DIFF
--- a/ecosystem-automation/collector-watcher/src/collector_watcher/component_scanner.py
+++ b/ecosystem-automation/collector-watcher/src/collector_watcher/component_scanner.py
@@ -89,7 +89,8 @@ class ComponentScanner:
                     components.extend(nested_components)
                 elif self._is_component_directory(item):
                     component_info = self._extract_component_info(item, component_type)
-                    components.append(component_info)
+                    if component_info is not None:
+                        components.append(component_info)
 
         return components
 
@@ -109,7 +110,8 @@ class ComponentScanner:
         for item in sorted(nested_dir.iterdir()):
             if item.is_dir() and self._is_nested_component_directory(item):
                 component_info = self._extract_component_info(item, component_type, subtype=subtype)
-                components.append(component_info)
+                if component_info is not None:
+                    components.append(component_info)
         return components
 
     def _is_valid_component_name(self, path: Path) -> bool:
@@ -191,7 +193,7 @@ class ComponentScanner:
 
     def _extract_component_info(
         self, component_path: Path, component_type: str, subtype: str | None = None
-    ) -> dict[str, Any]:
+    ) -> dict[str, Any] | None:
         """
         Extract information about a component.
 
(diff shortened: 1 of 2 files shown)


jaydeluca (approved):
thanks!

PR #1020, opened 2026-08-25
feat: split nightly registry update by ecosystem

### LINKED ISSUE #997: Split nightly registry update by ecosystem
Now that we have multiple ecosystems being sync'd each night, its likely that we will encounter situations where we might want to merge some and not others. Therefore I think we should split each ecosystem into it's own workflow and generate independent PRs for each.

It might make sense to create a shared worfklow that we can re-use and plug the new ecosystems in (similar to reusable-workflow-notification.yml)

Current workflow: https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/blob/main/.github/workflows/nightly-registry-update.yml

Adii-45:
## What changed

This PR completes the split of the nightly registry update workflow into independent per-ecosystem workflows!

- Adds dedicated workflows for Collector, Java, .NET, JavaScript, and Configuration
- Migrates each ecosystem to use the reusable registry-sync workflow introduced in PR #1018
- Preserves the existing schedules, dispatch triggers, concurrency, permissions, branch/PR configuration, and notification behavior
- Keeps Collector-specific schema-drift detection within the Collector workflow
- Removes the previous monolithic `nightly-registry-update.yml` workflow now that its responsibilities have been migrated

## Why

This completes the workflow refactoring so each ecosystem can run and be updated independently while sharing the common synchronization logic...

## How was this tested?

- Validated all new workflow YAML files successfully
- Ran Prettier formatting checks
- Verified the old workflow has no remaining references
- Verified the Collector schema-drift workflow and per-ecosystem configuration

## Is this a breaking change?

No

Closes #997


lucacavenaghi97 (approved):

.github/workflows/nightly-registry-update-collector.yml
```diff
@@ -0,0 +1,180 @@
...
+          if [ ! -f "$FRESH" ]; then
+            echo "Fresh schema not found in cloned repo, skipping drift detection"
+            echo "schema_changed=false" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+
+          # Schemas are stored in CAS at meta/schemas/{hash}.yaml. If a file
+          # already exists for the fresh content's hash, we've stored it
+          # before — no drift.
+          FRESH_HASH=$(sha256sum "$FRESH" | cut -c1-12)
```
jaydeluca: This hash comparison can never match, so `schema_changed=true` on every run.

`sha256sum` hashes the raw file bytes, but CAS filenames are hashes of the *normalized* YAML —
`CollectorSchemaCopier.compute_schema_hash` runs `safe_load` → `safe_dump(sort_keys=True, ...)` before
hashing (`collector-watcher/src/collector_watcher/schema_copier.py:84-119`). Against the current
collector `main` schema:

| hash                   | value          | exists in `meta/schemas/`? |
| ---------------------- | -------------- | -------------------------- |
| raw `sha256sum` (used) | `f47bd674eac7` | no                         |
| normalized (correct)   | `a0135ca939d9` | **yes**                    |

Same for all five stored schemas — none is named after its own raw SHA-256 prefix. So there's no real
drift today, but this will open `[schema-drift] collector metadata-schema.yaml changed` on the first
nightly run and comment on it every night after, with a comment/formatting-only diff. The
`Close schema drift issue if schema is stable` step below becomes unreachable.

It was harmless before this PR only because `$FRESH` never existed (see the checkout comment above).

Fix: don't re-derive the hash in shell. Add the uv + `setup-python` steps to this job and call the
watcher — e.g. a `collector-watcher schema-hash <path>` entry point — so the workflow and the CAS
share one implementation, and it gets unit coverage instead of living in an untested heredoc.

.github/workflows/nightly-registry-update-collector.yml
```diff
@@ -0,0 +1,180 @@
...
+    steps:
+      - name: Checkout opentelemetry-ecosystem-explorer
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Checkout opentelemetry-collector
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          repository: open-telemetry/opentelemetry-collector
```
jaydeluca: This checkout is what makes the drift check live for the first time — the watcher never populated
`tmp_repos/opentelemetry-collector` (it uses `-core` / `-contrib` suffixes), so the step always
short-circuited on the missing file. Good fix in itself; see the hash comment below for what it
exposes.

Two follow-ons while you're here:

- No `ref` means this reads `main` HEAD, whereas the registry only stores schemas read at release tags
  (`VersionDetector.read_file_at_ref`). Checking "have we ever seen this content?" against the CAS is
  a reasonable early warning, so keep it — but a comment saying so would help, since the old step read
  whatever ref the watcher happened to leave behind.
- The job needs one 23 KB file. `sparse-checkout: cmd/mdatagen/metadata-schema.yaml` (or a raw `curl`)
  avoids cloning the repo nightly.

.github/workflows/nightly-registry-update-collector.yml
```diff
@@ -0,0 +1,180 @@
...
+      pr-title: "[automated] Update collector registry inventory"
+    secrets:
+      OTELBOT_PRIVATE_KEY: ${{ secrets.OTELBOT_PRIVATE_KEY }}
+
+  detect-schema-drift:
+    name: Detect collector schema drift
+    if:
+      ${{ always() && !cancelled() && needs.synchronize-inventory.result != 'skipped' &&
+      (github.event_name != 'schedule' || github.repository ==
+      'open-telemetry/opentelemetry-ecosystem-explorer') }}
```
jaydeluca: `always() && !cancelled()` evaluates identically to `!cancelled()` — please restore the pre-fixup
form. GitHub's docs also advise against `always()` specifically because it suppresses cancellation.

Two other things on this condition:

- `github.event_name != 'schedule' || github.repository == '...'` is already implied by
  `needs.synchronize-inventory.result != 'skipped'` — if that guard skips the sync job, `result` is
  `skipped`. Dead clause.

```suggestion
    if: ${{ !cancelled() && needs.synchronize-inventory.result != 'skipped' }}
```

.github/workflows/nightly-registry-update-collector.yml
```diff
@@ -0,0 +1,180 @@
...
+            echo "Closed schema-drift issue #$EXISTING"
+          fi
+
+  notify:
+    name: Notify on collector-watcher failure
+    permissions:
+      contents: read # Checkout reusable workflow source
+      issues: write # Open or comment on the collector-watcher failure-tracking issue
+    needs: [synchronize-inventory]
+    if: ${{ always() && !cancelled() && needs.synchronize-inventory.result != 'skipped' }}
```
jaydeluca: Same `always()` revert here. Also, this job only `needs: [synchronize-inventory]`, so a
`detect-schema-drift` failure fails the workflow with no notification at all:

```yaml
needs: [synchronize-inventory, detect-schema-drift]
if: ${{ !cancelled() && needs.synchronize-inventory.result != 'skipped' }}
uses: ./.github/workflows/reusable-workflow-notification.yml
with:
  success:
    ${{ needs.synchronize-inventory.result == 'success' && needs.detect-schema-drift.result ==
    'success' }}
```

(Pre-existing gap — the monolith reported the watcher *step*'s outcome, not the drift step's — but
it's cheap to close now that drift is its own job.)

.github/workflows/nightly-registry-update-java.yml
```diff
@@ -0,0 +1,44 @@
...
+    secrets:
+      OTELBOT_PRIVATE_KEY: ${{ secrets.OTELBOT_PRIVATE_KEY }}
+
+  notify:
+    name: Notify on java-instrumentation-watcher failure
+    permissions:
+      contents: read # Checkout reusable workflow source
+      issues: write # Open or comment on the java-instrumentation-watcher failure-tracking issue
+    needs: [synchronize-inventory]
+    if: ${{ always() && !cancelled() && needs.synchronize-inventory.result != 'skipped' }}
```
jaydeluca:
```suggestion
    if: ${{ !cancelled() && needs.synchronize-inventory.result != 'skipped' }}
```

`always()` adds nothing here (see review body); this matches the monolith's `notify-java`.

.github/workflows/nightly-registry-update-dotnet.yml
```diff
@@ -0,0 +1,44 @@
...
+    secrets:
+      OTELBOT_PRIVATE_KEY: ${{ secrets.OTELBOT_PRIVATE_KEY }}
+
+  notify:
+    name: Notify on dotnet-instrumentation-watcher failure
+    permissions:
+      contents: read # Checkout reusable workflow source
+      issues: write # Open or comment on the dotnet-instrumentation-watcher failure-tracking issue
+    needs: [synchronize-inventory]
+    if: ${{ always() && !cancelled() && needs.synchronize-inventory.result != 'skipped' }}
```
jaydeluca:
```suggestion
    if: ${{ !cancelled() && needs.synchronize-inventory.result != 'skipped' }}
```

.github/workflows/nightly-registry-update-js.yml
```diff
@@ -0,0 +1,44 @@
...
+    secrets:
+      OTELBOT_PRIVATE_KEY: ${{ secrets.OTELBOT_PRIVATE_KEY }}
+
+  notify:
+    name: Notify on js-instrumentation-watcher failure
+    permissions:
+      contents: read # Checkout reusable workflow source
+      issues: write # Open or comment on the js-instrumentation-watcher failure-tracking issue
+    needs: [synchronize-inventory]
+    if: ${{ always() && !cancelled() && needs.synchronize-inventory.result != 'skipped' }}
```
jaydeluca:
```suggestion
    if: ${{ !cancelled() && needs.synchronize-inventory.result != 'skipped' }}
```

.github/workflows/nightly-registry-update-configuration.yml
```diff
@@ -0,0 +1,44 @@
...
+    secrets:
+      OTELBOT_PRIVATE_KEY: ${{ secrets.OTELBOT_PRIVATE_KEY }}
+
+  notify:
+    name: Notify on configuration-watcher failure
+    permissions:
+      contents: read # Checkout reusable workflow source
+      issues: write # Open or comment on the configuration-watcher failure-tracking issue
+    needs: [synchronize-inventory]
+    if: ${{ always() && !cancelled() && needs.synchronize-inventory.result != 'skipped' }}
```
jaydeluca:
```suggestion
    if: ${{ !cancelled() && needs.synchronize-inventory.result != 'skipped' }}
```

jaydeluca (commented):
A few things before we merge:

**1. The collector schema-drift check goes from dead to always-firing.** It was a permanent no-op
before this PR: it read `tmp_repos/opentelemetry-collector/...`, but the watcher clones to
`tmp_repos/opentelemetry-collector-{core,contrib}`
(`collector-watcher/src/collector_watcher/repository_manager.py:93`), so it always hit the
`"Fresh schema not found in cloned repo, skipping"` branch — zero `schema-drift` issues have ever
been opened in this repo. This PR checks the collector out to exactly that path, which makes the
check live, and the hash comparison underneath it is wrong. Details inline.

**2. `always() && !cancelled()` is a no-op.** Copilot's five inline comments on the first push were
all the same false positive — a job whose `if` doesn't call `success()` already runs when a `needs`
job fails (the implicit `success()` applies only when there is no `if` at all). That's how the
monolith's `notify-*` jobs worked. Please revert those five `if` lines to `!cancelled()`.

**3. Out of diff — `tmp_repos` cache key is now shared by five concurrent workflows.**
`reusable-registry-sync.yml:84-90` uses a static `key: tmp-repos`. `actions/cache` never overwrites
an existing key, so the entry is frozen at whatever content was saved first; five callers now fire in
the same minute with different clone content, and if the entry is ever evicted (repo cache size
limit) one workflow wins permanently and the other four run cold clones every night. Separately,
`restore-keys: tmp-repos-` can never match `key: tmp-repos`, so the fallback is dead. Suggest adding a
`cache-key` input to the reusable workflow and passing a per-ecosystem value — here or in a
fast-follow.

Smaller out-of-diff note: `reusable-workflow-notification.yml:44` searches
`--search "in:title $ISSUE_TITLE"` unquoted, so it's a relevance match rather than an exact one. I
probed it against the new dashed workflow names and the right ecosystem's issue still ranks first, so
it works today — but five workflows can now each hold an open failure issue, and `--limit 1` on a
fuzzy match is how you end up closing someone else's. Quoting it (as the drift step already does)
removes the risk.

.github/workflows/nightly-registry-update-js.yml
```diff
@@ -0,0 +1,44 @@
+name: Nightly Registry Update - JavaScript
+
+on:
+  schedule:
+    # Run daily at 2 AM UTC
+    - cron: "0 2 * * *"
```
jaydeluca, line 6: Non-blocking: all five workflows are on `0 2 * * *`. Staggering them (`0 2`, `15 2`, `30 2`, …) cuts
concurrent runner, GitHub API and cache contention, and scheduled runs queue less under load.

.github/workflows/nightly-registry-update-js.yml
```diff
@@ -0,0 +1,44 @@
...
+      'open-telemetry/opentelemetry-ecosystem-explorer'
+    permissions:
+      contents: write # Push registry updates to the ecosystem branch
+      pull-requests: write # Create or update the automated inventory-update PR
+    uses: ./.github/workflows/reusable-registry-sync.yml
+    with:
+      watcher-command: "uv run js-instrumentation-watcher"
+      registry-path: "ecosystem-registry/javascript/"
+      branch-name: "otelbot/automated-inventory-update-js"
+      pr-title: "[automated] Update js registry inventory"
```
jaydeluca, line 30: Non-blocking naming nit: `[automated] Update js registry inventory` and the `.NET` one's
`... dotnet ...` read oddly next to the workflow display names "JavaScript" / ".NET". Consider the
display names in `pr-title` and the job `name`, leaving `branch-name` and `registry-path` as-is.

.github/workflows/nightly-registry-update-java.yml
```diff
@@ -0,0 +1,44 @@
...
+    uses: ./.github/workflows/reusable-workflow-notification.yml
+    with:
+      success: ${{ needs.synchronize-inventory.result == 'success' }}
```
jaydeluca, line 44: Non-blocking: `success:` is now driven by the whole sync job, so checkout / `uv sync` / push /
`gh pr create` failures all report as `Workflow failed: ... (java-instrumentation-watcher)`. The
monolith used the per-step outcome, so the title was accurate. Fine trade for independence — passing
`failure-details` would keep the issue honest about which step died.

Adii-45:
@jaydeluca Thanks for the detailed review! I've addressed all the requested blocking changes from your review, including:

- corrected the schema hashing to match the collector's normalization logic
- replaced the full collector checkout with the latest-release schema fetch
- removed the unnecessary `always()` conditions
- updated the notification dependencies/conditions to account for both relevant jobs
- isolated the reusable workflow cache keys using per-ecosystem suffixes
- fixed the issue-title quoting in the failure-tracking workflow

PTAL whenever you get a chance...

jaydeluca (approved):
🚀

PR #1024, opened 2026-08-27
fix: respect selected version in source code link

### LINKED ISSUE #996: Collector detail page “View Source Code” link ignores the displayed component version
## Problem

On the Collector component detail page, the **"View Source Code"** link points to the `main` branch even when the page is displaying a specific historical component version via the `?version=` query parameter!

This makes the source link inconsistent with the version currently being viewed...

## Steps to reproduce

1. Open a collector component with an explicit historical version, for example:
   `/collector/components/core/otlpreeiver?version=0.150.0`
2. Confirm that the page displays metadata for version `0.150.0`
3. Under **Links & Resources**, click **View Source Code**
4. Observe the GitHub URL that is opened

## Actual behavior

The link points to:

`https://github.com/open-telemetry/opentelemetry-collector/tree/main/...`

even though the page is displaying version `0.150.0`...!

## Expected behavior

When a specific component version is being displayed, the **View Source Code** link should point to the corresponding version tag, for example:

`https://github.com/open-telemetry/opentelemetry-collector/tree/v0.150.0/...`

This would keep the source link consistent with the version selected in the Explorer..

## Additional context

The `deprecatedView` path already uses the selected version when constructing the source URL, while the general versioned path falls back to `main`!

The existing tests verify that explicit versions are correctly resolved and rendered, but do not currently assert the generated `View Source Code` URL..

A focused fix should therefore be possible by updating the source-link construction and adding regression coverage for the versioned case..

Adii-45:
## What changed

Updated the Collector detail page's **View Source Code** link to respect the selected component version!

When a specific version is selected, the link now points to the corresponding version-tagged path instead of always falling back to `main`!

Also added regression coverage for both the default and explicitly versioned cases...!

## Why

Previously, the source code link always pointed to `main` when a historical version was selected, making the link inconsistent with the version currently displayed in the Explorer..

## How it was tested

- Added regression tests for:
  - Default view → source link points to `main`
  - Explicit version (`v0.150.0`) → source link points to the corresponding version tag
- `npm run test -- ecosystem-explorer/src/features/collector/collector-detail-page.test.tsx`
- `npx tsc --noEmit`
- `npm run lint`

All tests passed successfully

## Special notes for reviewers

This is intentionally a minimal fix that preserves the existing `deprecated` behavior while ensuring explicitly selected versions use the correct versioned source link!

Closes #996

### DIFF
--- a/ecosystem-explorer/src/features/collector/collector-detail-page.test.tsx
+++ b/ecosystem-explorer/src/features/collector/collector-detail-page.test.tsx
@@ -459,4 +459,46 @@ describe("CollectorDetailPage", () => {
     expect(screen.getByText("Stability Levels")).toBeInTheDocument();
     expect(screen.queryByText("Distribution Availability")).not.toBeInTheDocument();
   });
+
+  it("View Source Code link points to the main branch when no version is explicitly selected", () => {
+    vi.mocked(useCollectorVersions).mockReturnValue({
+      data: { versions: [{ version: "0.150.0", is_latest: true }] },
+      loading: false,
+      error: null,
+    });
+    vi.mocked(useCollectorComponent).mockReturnValue({
+      data: mockComponentWithoutTelemetry,
+      loading: false,
+      error: null,
+    });
+
+    renderAtRoute("/collector/components/core/otlpreceiver");
+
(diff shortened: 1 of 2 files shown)


vitorvasc (approved):
Thank you!

PR #1083, opened 2026-09-07
fix(v1): stabilize navigation, theme and drawer behavior

vitorvasc:
## What changed

Collector navigation keeps the selected release through historical routes, detail links, source links, and tab history. The mobile filter drawer retains focus while filters change and releases modal behavior at desktop widths. Theme utilities follow the selected Light, Dark, or Auto mode. Screenshot acceptance checks verify the requested app and fail when required tabs, panels, or the drawer are missing.

The unused theme palette mirror is removed in a separate commit. Five commits cover C01, C03, C09, and the C10 acceptance foundation from #1082.

## Why

These are the first implementation steps from the v1 architecture review for #84. They fix observed navigation, focus, and theme inconsistencies and make the acceptance runner reliable for subsequent changes.

## How it was tested

- Unit suite: 1,577 passed; 11 opt-in browser tests skipped in that run. The acceptance suite passed all 30 tests with browser checks enabled.
- Integration suite: 60 passed. Type checking, production build, formatting, and lint passed; lint retains three existing warnings.
- Chromium checks against the built v1 app covered historical release navigation, tab history, drawer focus and resizing, and all four OS/selected-theme combinations. Legacy theme rendering also passed.
- Screenshot capture produced 80 new images across three viewports and both themes. The 28 accessibility reports contain 22 remaining violation instances: contrast, heading order, and landmarks.
- React Doctor remains 63/100 with the same two existing errors and one fewer warning (119). No new diagnostic locations.

## Is this a breaking change?

No.

This is the first implementation batch; the architecture plan and visual report remain in the separate documentation draft #1082. Review the release helper, responsive focus lifecycle, and theme selector changes first. The final commit makes app selection explicit in acceptance tooling; screenshot pixel differences remain informational.

The remaining accessibility findings and later architecture candidates are follow-up work. No performance improvement is claimed without measurements. Screenshot publication is requested through the repository's `add-screenshots` workflow.

## Checklist


didiViking (approved):

ecosystem-explorer/scripts/acceptance.test.mjs
```diff
@@ -0,0 +1,254 @@
...
+import {
+  acceptanceConfig,
+  optionalScenario,
+  waitForReady,
+  clickTab,
+  openCollectorFilters,
+} from "./acceptance.mjs";
+
+const workflow = load(
+  readFileSync(resolve("../.github/workflows/screenshots-capture.yml"), "utf8")
```
jaydeluca, line 31: This reads a file at module load with a path resolved against `process.cwd()` rather than the test file's own location, so it throws during import if vitest is invoked from anywhere but `ecosystem-explorer/`. From the repo root:

```
$ node -e "readFileSync(resolve('../.github/workflows/screenshots-capture.yml'))"
THROWS: ENOENT /.github/workflows/screenshots-capture.yml
```

A module-load throw surfaces as "Failed Suites 1, 0 tests" with no test names, which is painful to diagnose. `theme-rendering.test.tsx:29` in this same PR already has the pattern that avoids both problems - `import.meta.dirname` plus the read inside `beforeAll`:

```js
let workflow;
beforeAll(async () => {
  workflow = load(
    await readFile(
      resolve(import.meta.dirname, "../../.github/workflows/screenshots-capture.yml"),
      "utf8"
    )
  );
});
```

The same cwd assumption applies to `resolve("scripts", script)` on line 68, though that one at least fails with a test name attached.

.github/workflows/nightly-declarative-config-test.yml
```diff
@@ -48,6 +48,8 @@ jobs:
           bunx playwright install --with-deps chromium
 
       - name: Generate config file
+        env:
+          ACCEPTANCE_APP_MODE: legacy
```
jaydeluca, line 52: Pinning the acceptance mode here is right, but `Build explorer` on line 43 is still a bare `bun run build` with no `VITE_FEATURE_FLAG_V1_REDESIGN`. `waitForReady` now hard-fails on a mode mismatch, so this workflow's correctness rests on the flag's compiled default staying `false` - and it breaks with `Expected legacy app, rendered v1` the moment V1_REDESIGN defaults on, which is the point of #84.

`screenshots-baseline.yml` and `screenshots-capture.yml` both pin their build flag; adding `VITE_FEATURE_FLAG_V1_REDESIGN: "false"` to the build step here would make all three consistent. (Suggested diff in the review body - the build step itself is outside this PR's diff, so I can't anchor there.)

ecosystem-explorer/src/v1/features/list/list-page.test.tsx
```diff
@@ -229,6 +251,7 @@ describe("CollectorListPageV1", () => {
     await user.keyboard("{Escape}");
     expect(screen.queryByRole("dialog")).not.toBeInTheDocument();
     expect(toggle).toHaveFocus();
+    matchMedia.mockRestore();
```
jaydeluca, line 254: Restoring inline at the end of the test body means the spy leaks if any assertion above throws - and this spy returns `matches: false` for *every* query, including `prefers-color-scheme`, so one failure here cascades into unrelated failures for the rest of the file. There's no `afterEach` in this file and no `restoreMocks` in `vitest.config.ts` to catch it.

`facet-panel.test.tsx:90` in this same PR gets it right:

```js
afterEach(() => {
  cleanup();
  document.body.style.overflow = "";
  vi.restoreAllMocks();
  vi.useRealTimers();
});
```

Moving the restore into an `afterEach` here would match that and the repo convention.

ecosystem-explorer/scripts/acceptance.mjs
```diff
@@ -0,0 +1,108 @@
...
+          `VITE_FEATURE_FLAG_V1_REDESIGN=${appMode === "v1"} and set ACCEPTANCE_APP_MODE=${appMode}`
+      );
+    }
+    await page.waitForLoadState("networkidle", { timeout });
+    // Exact route failure copy avoids rejecting legitimate headings such as
+    // "Error handling" in documentation or the component showcase.
+    const errorHeading = page
+      .locator("main")
+      .first()
+      .getByRole("heading", {
```
jaydeluca, line 75: Suggestion. I checked this allowlist against every error string in `public/locales/en/*.json` and it does cover all the current *heading* strings, so it's accurate as written. The gap is that v1's own error and empty states aren't headings - they render as `<p class="td-empty__title">` inside a `role="alert"` container:

| String | Source |
|---|---|
| `Component not found` | `detail.json`, rendered at `detail-page.tsx:204` |
| `Couldn't load the diff` | `detail.json`, `diff-page.tsx:190` |
| `Couldn't load components` | hardcoded, `list-page.tsx:296` |

Detection for those rests entirely on the per-scenario `ready` selectors, which do happen to cover them today. But the four `visit()` calls that fall back to the default `"main h1"` selector (home, collector-landing, java-agent-landing, dev-components) would screenshot an alert-shaped error state without noticing.

Adding a check for `main [role="alert"]` alongside the heading check would cover both shapes and be locale-independent, rather than a regex that has to be kept in sync with two locale files by hand.

ecosystem-explorer/src/v1/lib/collector-release.ts
```diff
@@ -0,0 +1,65 @@
...
+    detailHref(component: ComponentAddress, version = selectedVersion): string {
+      const suffix = version ? `?version=${encodeURIComponent(version)}` : "";
+      return `/collector/components/${component.distribution}/${component.name}${suffix}`;
+    },
+    sourceHref(component: ComponentSource, lastVersion?: string): string | null {
+      if (!component.repository) return null;
+      const version = deprecated ? lastVersion : selectedVersion;
+      if (deprecated && !version) return null;
+      // An implicit latest view follows main; an explicit release is pinned to its tag.
+      const ref = version ? `v${version}` : "main";
```
jaydeluca, line 59: Suggestion, low severity. This now trusts the `?version=` value verbatim when building a tag ref, so `?version=v0.160.0` produces `/tree/vv0.160.0/`. It was unreachable before, because non-deprecated views always used `main`.

Real data has no prefix (`versions-index.json` -> `"0.160.0"`) and the data fetch would 404 into "Component not found" before a Source link renders, so practical impact is low. Flagging it mainly because two places model the *other* convention and could mislead someone later: `list-filters.ts:26` documents the query as `&version=v0.150.0`, and `facet-panel.test.tsx:424` uses `versions: ["v0.150.0"]`. Stripping a leading `v` here would close it.

ecosystem-explorer/src/v1/lib/collector-release.ts
```diff
@@ -0,0 +1,65 @@
...
+  pathVersion,
+  versions,
+}: {
+  searchParams: URLSearchParams;
+  pathVersion?: string;
+  versions?: VersionsIndex | null;
+}) {
+  // The query string is the canonical carrier; a version in the path is the historical
+  // route form and yields to it when both are present.
+  const selectedVersion = searchParams.get("version") || pathVersion || undefined;
```
jaydeluca, line 34: Minor: `parseFilters` trims this same param (`list-filters.ts:134` - `params.get("version")?.trim() || null`) and this doesn't, so on the list page `filters.version` and the release context can disagree for a whitespace-padded value. Cosmetic, but they're now two readers of one URL param and it'd be nice if they normalized identically.

ecosystem-explorer/src/v1/components/list/facet-panel.tsx
```diff
@@ -34,14 +34,42 @@
...
+  return window.matchMedia(DESKTOP_QUERY).matches;
+}
+
+function subscribeViewport(onChange: () => void) {
+  const media = window.matchMedia(DESKTOP_QUERY);
+  media.addEventListener("change", onChange);
+  return () => media.removeEventListener("change", onChange);
+}
+
+function isVisible(element: HTMLElement): boolean {
```
jaydeluca, line 58: Suggestion. This walks the full ancestor chain calling `getComputedStyle` on each element, and `focusables()` calls it for every candidate inside the `Tab` branch of the keydown handler (line 164). That's roughly 30 facet controls x ancestor depth worth of forced style recalc on every Tab keypress while the drawer is open.

`Element.checkVisibility()` would be a single call each, but jsdom doesn't implement it, so that'd cost you the test coverage. Cheaper options that keep jsdom working: short-circuit on `element.offsetParent === null` before walking, or hoist the visibility filter out of the Tab path entirely - Tab only needs the first and last focusable, not the whole filtered list.

Worth noting the filter itself is a real improvement: the previous selector-only version would trap focus on hidden and disabled controls.

ecosystem-explorer/scripts/generate-test-config.mjs
```diff
@@ -98,34 +103,42 @@ async function generateConfig() {
       await dialog.accept();
     });
     await page.goto(`${BASE_URL}/java-agent/configuration/builder`, {
-      waitUntil: "networkidle",
+      waitUntil: "domcontentloaded",
       timeout: 20000,
     });
+    await waitForReady(page, {
```
jaydeluca, line 109: Minor: `goto` above dropped from `waitUntil: "networkidle", timeout: 20000` to `domcontentloaded`, and this `waitForReady` call passes no timeout, so network settling now gets the 10s default instead of 20s. The heavy instrumentation load happens after the tab click (which still has 10s/30s), so this is probably fine - but it's a nightly-only flake surface with no PR signal, so passing an explicit `20000` here would cost nothing.

ecosystem-explorer/src/v1/styles/facet-panel.css
```diff
@@ -41,7 +41,8 @@
   gap: 1rem;
 }
 
-@media (max-width: 991px) {
+/* Complement the desktop query exactly, including fractional viewport widths. */
+@media (width < 992px) {
```
jaydeluca, line 45: Nit, no action needed. The range syntax is correct and the comment explains why it beats `max-width: 991px`. Just noting the codebase now has two idioms for the same complement - this and `language-toggle.css:53`'s `@media (max-width: 991.98px)` - in case you want to settle on one.

jaydeluca (commented):
some inline comments. also `src/test/setup.ts:71` (not in the diff) — the shared `matchMedia` stub now answers the new viewport query. It returns `matches: true` for *any* query, and its comment documents that only as
"defaults to 'system prefers dark'". `facet-panel.tsx` now asks `(min-width: 992px)`, so every test
that doesn't override the stub gets `isDesktop === true` → `isDrawer === false`. That's exactly why
this PR had to add a stub to the existing drawer test in `list-page.test.tsx` — the hazard has
already bitten once. Nothing is wrong today (jsdom's 1024px default makes the answer coincidentally
correct), but the next drawer or focus-trap test written without a viewport stub will silently
exercise the rail path and pass vacuously. Either narrow the stub to `prefers-color-scheme`, or
extend the comment to say it also forces the desktop breakpoint.

jaydeluca (commented):
**`.github/workflows/build-and-test.yml` (`test-ecosystem-explorer`, lines 284-322) — the opt-in
Chromium tests have nowhere to run.** Expanding on the inline comment at
`scripts/acceptance.test.mjs:169`: that job runs `bun run test`, `bun run test:integration`, `bun run
build` and `bun run test:edge`, but installs no Playwright browsers, so
`ACCEPTANCE_BROWSER_TESTS=true` can't be set there as-is. Adding a step would look like:

```yaml
      - name: Install Playwright browsers
        run: bunx playwright install --with-deps chromium

      - name: Run tests
        env:
          ACCEPTANCE_BROWSER_TESTS: "true"
        run: bun run test
```

ecosystem-explorer/src/v1/lib/collector-release.ts
```diff
@@ -0,0 +1,71 @@
...
+}) {
+  // The query string is the canonical carrier; a version in the path is the historical
+  // route form and yields to it when both are present.
+  const selectedVersion =
+    searchParams
+      .get("version")
+      ?.trim()
+      .replace(/^v(?=\d)/, "") ||
+    pathVersion?.trim().replace(/^v(?=\d)/, "") ||
+    undefined;
```
jaydeluca, line 40: Stripping the `v` here is the right answer to the earlier review, but `parseFilters` is the other reader of this same param and it only trims (`list-filters.ts:134` — `params.get("version")?.trim() || null`). The list page uses both:

```tsx
const currentVersion = release.dataVersion();            // "0.149.0"
<FacetPanel filters={filters} versions={allVersions} />  // filters.version === "v0.149.0"
```

`allVersions` comes from `versionsData`, so it holds bare strings. `v0.149.0` matches no `<option>` and the controlled `<select>` falls back to the empty "Latest" entry — the user is looking at 0.149.0 data with the release selector reading "Latest", and `ActiveFilterChips` shows a `v0.149.0` chip that corresponds to nothing. Same failure shape as the `<select value={i18n.language}>` region-tag bug in #651/#652: a controlled select whose value isn't in its option list renders blank rather than wrong, so the desync is silent.

Confirmed by adding one row to this PR's own consistency test (see my comment on `list-page.test.tsx:177`):

```
 FAIL  uses the same normalized version for the facet, data and links: v0.149.0
 Error: expect(element).toHaveValue(0.149.0)
 Expected the element to have value:  0.149.0
 Received:
```

`useCollectorComponents` and the row `href` both pass — only the facet diverges.

Worth normalizing in one place: either give `parseFilters` the same `.trim().replace(/^v(?=\d)/, "")`, or export a `normalizeRelease()` from this file and call it from both. A side benefit is that `sourceHref`'s second `.replace(/^v/, "")` on line 65 becomes redundant and can go — right now the two spellings differ (`/^v(?=\d)/` vs `/^v/`), which reads like one of them is guarding a case the other doesn't.

ecosystem-explorer/src/v1/features/list/list-page.test.tsx
```diff
@@ -156,23 +174,42 @@ describe("CollectorListPageV1", () => {
...
 
-  it("renders the deprecated catalog without changing the list layout", () => {
-    vi.mocked(useCollectorDeprecations).mockReturnValue({
-      data: { ecosystem: "collector", components: [deprecatedComponent] },
-      loading: false,
-      error: null,
-    });
+  it.each([
+    ["+0.149.0+", "0.149.0", "?version=0.149.0", "0.149.0"],
+    ["+%09+", "0.150.0", "", ""],
```
jaydeluca, line 179: This test is the right shape for the facet/data/link divergence in my comment on `collector-release.ts:34` — it just needs the `v`-prefix row, which is the one normalization the release context applies and `parseFilters` doesn't:

```diff
   it.each([
     ["+0.149.0+", "0.149.0", "?version=0.149.0", "0.149.0"],
     ["+%09+", "0.150.0", "", ""],
+    ["v0.149.0", "0.149.0", "?version=0.149.0", "0.149.0"],
   ])(
```

That row fails today on the `toHaveValue` assertion and passes once the two readers agree.

ecosystem-explorer/src/v1/features/detail/detail-page.test.tsx
```diff
@@ -403,6 +424,109 @@ describe("CollectorDetailPageV1", () => {
...
+      expect(screen.getByRole("tab", { name: "Configuration" })).toHaveAttribute(
+        "aria-selected",
+        "true"
+      );
+      expect(screen.getByTestId("location").textContent).toBe(path);
+    }
+  );
+});
+
+describe("Collector detail release changes", () => {
```
jaydeluca, line 484: This block has no `beforeEach`, unlike the two sibling blocks in the file (`:155` and `:432`, both `vi.clearAllMocks()`), and `vitest.config.ts` sets neither `clearMocks` nor `restoreMocks`. `mockHooks()` only assigns return values — it never touches call history. So `expect(useCollectorComponent).toHaveBeenCalledWith(...)` on line 492 matches calls made by earlier tests in the file, and both of this block's parameters are versions earlier tests already requested: `"0.149.0"` from the tab-routing block's `?version=0.149.0`, `"0.150.0"` from the default-route tests. Both rows are free passes right now.

Mutation probe — expected data version changed to the wrong value, whole file run:

```
$ bunx vitest run <probe>.test.tsx     # 0.149.0 expected where 0.150.0 is correct
  Test Files  1 passed (1)
  Tests  22 passed (22)                <- vacuous pass

$ # same probe, with beforeEach(() => vi.clearAllMocks()) added to this block
  Tests  1 failed | 21 passed (22)
  -> expect(useCollectorComponent).toHaveBeenCalledWith("core","otlpreceiver","0.149.0")
     1st vi.fn() call: ["core","otlpreceiver","0.150.0"]
```

The `Source` / sibling / breadcrumb `href` assertions in the same test are real, so the block isn't worthless — but the data-version half of "keeps data, source and navigation consistent" proves nothing. Either add the `beforeEach`, or set `clearMocks: true` in `vitest.config.ts` and drop the three hand-rolled calls across the file.

ecosystem-explorer/scripts/acceptance.test.mjs
```diff
@@ -0,0 +1,331 @@
...
+      /collector-list.*collector\/components.*required ready v1.*network did not settle/
+    );
+  });
+});
+
+// Opt-in real Chromium checks; the regular unit suite needs no browser install.
+// ACCEPTANCE_BROWSER_TESTS=true NODE_OPTIONS=--no-experimental-webstorage \
+//   bun run test scripts/acceptance.test.mjs --maxWorkers=2
+// Uses intercepted fixture URLs, so it opens no dev-server port.
+describe.runIf(process.env.ACCEPTANCE_BROWSER_TESTS === "true")(
```
jaydeluca, line 169: `ACCEPTANCE_BROWSER_TESTS` appears nowhere under `.github/` — only in the comment two lines above, as a copy-pasteable local command. So these 15 tests never run in CI.

`test-ecosystem-explorer` in `build-and-test.yml` installs no Playwright browsers, so it can't just be flipped on.

ecosystem-explorer/src/theme-rendering.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+beforeAll(async () => {
+  const from = resolve("src/styles/index.css");
+  // Compile the real entry point with the installed Tailwind plugin, including
+  // its imports and selector flattening. jsdom cannot compute layered CSS.
+  // Tailwind bundles another PostCSS 8 type tree; the runtime plugin API is shared.
+  const plugin = tailwind({ optimize: true }) as AcceptedPlugin;
+  const result = await postcss([plugin]).process(await readFile(from, "utf8"), { from });
+  compiled = result.root;
+  const html = new DOMParser().parseFromString(
+    await readFile(resolve("index.html"), "utf8"),
```
jaydeluca, line 38: Both reads resolve against `process.cwd()`, which is the same assumption `acceptance.test.mjs` just shed in favour of `import.meta.dirname` — and this file was the one I pointed at as the example to copy for the `beforeAll` half of that fix.

Consistency only: these are inside `beforeAll`, so a wrong cwd fails with a test name attached rather than as a nameless failed suite, and CI is fine because `test-ecosystem-explorer` sets `working-directory: ecosystem-explorer`. `resolve(import.meta.dirname, "styles/index.css")` and `resolve(import.meta.dirname, "../index.html")` would make it invocation-independent.

ecosystem-explorer/src/v1/styles/navbar.css
```diff
@@ -170,7 +170,7 @@
  * upstream's mobile min-height (6.5rem ≈ 104px) and collapse the nav row
  * into an overlay panel toggled by the hamburger.
  */
-@media (max-width: 767.98px) {
+@media (width < 768px) {
```
jaydeluca, line 173: Stale comment just above, at line 134 (outside the diff hunk, so I can't anchor on it):

```css
 * Hamburger toggler — hidden on desktop, revealed below `md` (767.98px).
```

`767.98px` is gone from the tree after this PR — this comment is the only place that still names it. Worth updating while the file is open.

The unification itself is complete, for what it's worth: every remaining `@media` in `src/v1/styles/` is either `min-width` or the new range form, with no mixed complement pairs left.

ecosystem-explorer/src/v1/lib/collector-release.ts
vitorvasc: Centralized normalization in the Collector release module and reused it for filter parsing, serialization, and links. The selector, chips, and data request now agree on `0.149.0` for `?version=v0.149.0`. Removed the separate source-tag normalization too.

Fixed in [c0a7602e](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/commit/c0a7602ea3a38a4b1fb430e3dccca09916f7b4e3).

ecosystem-explorer/src/v1/features/list/list-page.test.tsx
vitorvasc: Added the `v0.149.0` row and a whitespace-padded variant. Both reproduced the selector mismatch before the fix and pass now. The test also checks the active chip, preserves the URL on arrival, and confirms that changing list density writes the canonical release to the URL.

Fixed in [c0a7602e](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/commit/c0a7602ea3a38a4b1fb430e3dccca09916f7b4e3).

ecosystem-explorer/src/v1/features/detail/detail-page.test.tsx
vitorvasc: Moved mock clearing and the hash reset into a file-level `beforeEach`. The release checks now assert the latest hook call, and I verified them with the whole file running.

Fixed in [c0a7602e](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/commit/c0a7602ea3a38a4b1fb430e3dccca09916f7b4e3).

ecosystem-explorer/scripts/acceptance.test.mjs
vitorvasc: Added Chromium installation to the frontend CI job and enabled `ACCEPTANCE_BROWSER_TESTS` on the test step. All 15 browser checks now run and pass in [CI](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/actions/runs/35567872785/job/106233203440); the acceptance file passed all 37 tests.

Fixed in [c0a7602e](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/commit/c0a7602ea3a38a4b1fb430e3dccca09916f7b4e3).

ecosystem-explorer/src/theme-rendering.test.tsx
vitorvasc: Changed both reads to resolve from `import.meta.dirname`. Verified the theme tests from the frontend directory and the repo root.

Fixed in [c0a7602e](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/commit/c0a7602ea3a38a4b1fb430e3dccca09916f7b4e3).

ecosystem-explorer/src/v1/styles/navbar.css
vitorvasc: Updated the comment to match `width < 768px`.

Fixed in [c0a7602e](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/commit/c0a7602ea3a38a4b1fb430e3dccca09916f7b4e3).

vitorvasc:
Addressed the two review summaries too:

- [The `matchMedia` note](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/1083#pullrequestreview-5140995026): [caf1b5be](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/commit/caf1b5bef75e3a5b688d63e5dd209c536457d62f) documents that the shared stub forces dark-mode preferences and desktop breakpoints, and that drawer tests need an explicit viewport mock.
- [The Chromium CI note](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/pull/1083#pullrequestreview-5221359225): [c0a7602e](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/commit/c0a7602ea3a38a4b1fb430e3dccca09916f7b4e3) installs Chromium and enables `ACCEPTANCE_BROWSER_TESTS` in the frontend job. [CI passed all 37 acceptance tests, including the 15 browser checks](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/actions/runs/35567872785/job/106233203440).

The six inline threads have replies with fix links and are resolved. Build and Test passed with 1,765 tests, 60 integration tests, and 8 edge tests.

jaydeluca (approved):

Issue #636, opened 2026-06-02
Make the Explorer information accessible to agents


jaydeluca:
This is going to be a sticky issue to keep a continual discussion going and will stay open, so please do not ask to be assigned. For ideas that we decide to move forward on, we can create subtask issues and assign those out.

The point of this issue is to brainstorm and discuss ways of improving the explorer so that the information is easily accessible and usable to AI workflows and agents.

It builds upon the work that was started via #322 

We are looking for ideas of both improving the site, but also ideas of ways to measure our progress.

One site we have used is: https://buildwithfern.com/agent-score/company/opentelemetry-ecosystem-explorer

As of today (June 2, 2026) we have a "B" score (87).

Let's get that to 100 and then find other benchmarks to conquer.

Pittu-Sharma:
I would like to work on this issue. Please assign it to me!

jaydeluca:
hey @Pittu-Sharma, please read the issue:

> This is going to be a sticky issue to keep a continual discussion going and will stay open, **so please do not ask to be assigned**. For ideas that we decide to move forward on, we can create subtask issues and assign those out.
> The point of this issue is to brainstorm and discuss ways of improving the explorer so that the information is easily accessible and usable to AI workflows and agents.


If you have a particular idea, we can discuss and create a subtask

Pittu-Sharma:
Hi @jaydeluca,

I've finished implementing the solution for the telemetry reconciliation bug and have broken it down into the two PRs as you suggested!

Here is a quick summary of what I've done:

1. PR 1: when-aware reconciliation (Frontend)

- I've updated the compareReleases logic in release-diff.ts so that it intercepts spans and metrics across all when conditions. When a signal is matched across conditions, it now merges the removed/added pair into a single changed entry.
- I've extended MetricChanges and SpanChanges with a new whenChanged interface property to track the "before" and "after" states.
- I've updated the frontend UI (metric-diff-card.tsx and span-diff-card.tsx) to explicitly render a "CONDITION MOVED" message showing the old condition struck out and the new condition highlighted.
- Tested against a genuine add, remove, real content change, and a when condition move to ensure all diff scenarios render perfectly.
2. PR 2: telemetry corrections overlay (db-builder)

- I've created the telemetry_when_corrections.yaml seed file under explorer-db-builder/src/explorer_db_builder/ and populated it with all the ignore_conditions and corrections data you provided.
[...]

PR #1085, opened 2026-09-08
Add reverse indexes for llm navigation

jaydeluca:
Generates a reverse-lookup and facet artifacts for both ecosystems. One pass over the latest release writes `by-metric.json` and `by-attribute.json` (name to the component ids that reference it) plus `facets.jsonl` (one row per component: id, signals, description, page) under `/data/{ecosystem}/`. Both section indexes gain a `Signals` column and a section pointing at the new artifacts, and `llms.txt` documents the lookup pattern.

Also fixes a bug with content negotiation when agents try and access json files

Part of #636

### DIFF
--- a/ecosystem-explorer/netlify/edge-functions/__tests__/agent-negotiation.test.ts
+++ b/ecosystem-explorer/netlify/edge-functions/__tests__/agent-negotiation.test.ts
@@ -130,6 +130,43 @@ describe("agent-negotiation edge function", () => {
     expect(res?.headers.get("content-type")).toBe("application/json");
   });
 
+  // Agents that send `Accept: text/markdown` on every request must still be able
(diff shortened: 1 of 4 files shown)


lucacavenaghi97 (changes requested):
The two Copilot filed under "previously missed" are right, and I'd call them blocking. They're in a collapsed block under a "Findings: None" header, which is probably why they've sat there.

Numbers on 0.161.0: 41 of 1713 metric definitions have a `prefix`, and 126 of 663 attributes have a `name_override` that isn't the map key. `by-attribute.json` lists 2 components under `type` when 11 emit it, and nothing in `by-metric.json` starts with `otelcol.`.

Rest of it looked fine to me, including the 304 handling on `/data` from #632.

lucacavenaghi97 (approved):

PR #908, opened 2026-07-27
feat(collector): add internal telemetry version comparison

vanshika2720:
Collector metadata.yaml can define a `telemetry` block describing the internal self-observability metrics a component emits about its own operation (e.g. `processor_memory_limiter_refused_spans`), distinct from the existing `metrics` field which describes the signal data a component produces about the monitored system. Once collector-watcher preserves this field, nothing surfaced it: explorer-db-builder's collector transformer only passed through `attributes` and `metrics`, so `telemetry` would still be silently dropped before reaching the frontend.

Add the missing passthrough in collector_transformer.py, model `telemetry` on CollectorComponent, and add a per-component "Internal Telemetry" tab (shown only when the field is present) with the same Current/Comparison UX already used for Java Agent instrumentation telemetry: a From/To version picker and diff cards for added, removed, and changed metrics. Collector telemetry has no `when`-condition concept, so unlike the Java Agent version there is no configuration-axis selector.

## What changed

Added `telemetry` passthrough in the collector transformer, modeled `telemetry` on `CollectorComponent`, and added an "Internal Telemetry" tab with Current/Comparison views, From/To version selection, and diff cards for added, removed, and changed metrics.

## Why

Fixes #877

## How it was tested

- Added tests for telemetry diffing, metric types, comparison loading, UI rendering, and collector transformer passthrough.
- `1306/1306` frontend tests passed.
- `375/375` Python tests passed.
- Typecheck, lint, and formatting checks passed.
- Manually verified added, removed, and changed telemetry metrics using local version fixtures.
- Verified the existing Collector Telemetry tab behavior remains unchanged.

## Is this a breaking change?

No

## Special notes for your reviewer

This PR depends on #876. The feature can be verified against production telemetry data after #876 is merged and the collector-watcher backfill and explorer database rebuild are completed.

Collector telemetry does not have a `when`-condition/configuration concept, so no configuration-axis selector is included.

## Checklist

- [x] Tests added or updated for new behavior
- [ ] Screenshots attached for any UI changes
- [x] I wrote this PR description myself (AI tools must not check this box on behalf of the author)

PR #1092, opened 2026-09-10
feat(collector): add internal telemetry comparison

### LINKED ISSUE #877: [Collector] Telemetry comparison by version

Adii-45:
## What changed

Adds end-to-end support for viewing and comparing Collector internal telemetry in the OpenTelemetry Ecosystem Explorer!

This brings Collector internal telemetry into the Collector component detail page through a dedicated `Internal Telemetry` tab, with both current-version viewing and version-to-version comparison!

### Changes

- Extend Collector types to represent internal telemetry metrics and comparison results
- Add shared metric-type detection for telemetry metrics
- Add telemetry diffing for:
  - Added metrics
  - Removed metrics
  - Changed metrics
  - Unchanged metrics
  - Metric attribute differences
- Add the `Internal Telemetry` tab to Collector component pages when telemetry data is available
- Add `Current View` and `Version Comparison` modes
- Add `From` / `To` Collector version selectors for comparisons
- Add loading, error, same-version, unavailable-version, and no-differences states
- Add metric comparison cards showing the relevant metric details and before/after changes
- Reuse the existing Collector telemetry UI where applicable
- Add and update English and Spanish translations for the new UI
- Add frontend tests covering the diff logic, metric types, comparison hook, loading/error states, tab rendering, and view switching
- Redesign the `Expand All` / `Collapse All` controls so they are integrated into the Metrics section header instead of appearing as a standalone centered control

## Related work

This PR incorporates and continues the work from:

- #908 - Initial Collector internal telemetry version comparison implementation
- #941 - Collector internal telemetry frontend/UI implementation
- #945 - Collector telemetry metadata passthrough in the Explorer DB builder

## Why

Collector components can expose internal self-observability metrics describing their own operation.. Previously, this telemetry was not surfaced in the Ecosystem Explorer!

This change makes that information discoverable directly from Collector component pages and allows users to compare internal telemetry across Collector releases

Fixes #877 

## Testing

- Typecheck passed
- Lint passed
- Frontend tests passed
- Collector feature tests passed
- Telemetry diff, metric-type, comparison hook, and UI test coverage added
- Manually verified the Collector Internal Telemetry UI locally across current and version-comparison views
- Verified existing Collector telemetry behavior remains intact

## Breaking change

No

### Internal Telemetry - Current View

image

### Internal Telemetry - Version Comparison

image

### DIFF
--- a/ecosystem-explorer/public/locales/es/collector.json
+++ b/ecosystem-explorer/public/locales/es/collector.json
@@ -180,8 +180,13 @@
       "details": "Detalles",
       "stability": "Estabilidad",
       "telemetry": "Telemetría",
+      "internalTelemetry": "Internal Telemetry",
       "featureGates": "Feature Gates",
       "readme": "Readme"
+    },
+    "view": {
+      "current": "Current View",
+      "comparison": "Version Comparison"
     },
     "sections": {
       "componentInfo": "Información del componente",
@@ -278,6 +283,68 @@
       "stableIn": "Estable en {{version}}",
       "deprecatedIn": "Obsoleto en {{version}}",
       "reference": "Referencia"
+    }
+  },
+  "telemetryComparison": {
+    "loading": "Loading comparison...",
+    "error": {
+      "title": "Failed to load comparison"
+    },
+    "warnings": {
+      "availability": {
+        "title": "Version availability",
+        "fromNotFound": "Could not load internal telemetry for version {{fromVersion}}.",
+        "toNotFound": "Could not load internal telemetry for version {{toVersion}}."
+      },
+      "sameVersion": {
+        "title": "Same version selected",
+        "message": "Choose two different versions to see a comparison."
+      }
+    }
+  },
+  "versionSelectorPanel": {
+    "banner": "Compare internal telemetry metrics between two Collector releases.",
+    "from": "From",
+    "to": "To",
+    "latestSuffix": "(latest)"
+  },
+  "diffResults": {
+    "addedRemovedHeader": "Added & Removed",
+    "changedHeader": "Changed",
+    "empty": {
+      "title": "No differences found",
+      "description": "Internal telemetry metrics are identical between these two versions."
+    }
+  },
+  "diffCard": {
+    "status": {
+      "added": "Added",
+      "removed": "Removed",
+      "changed": "Changed"
+    },
+    "unit": "Unit",
+    "descriptionChanged": "Description changed",
+    "enabledChanged": "Enabled state changed",
+    "enabledValue": {
+      "enabled": "Enabled",
+      "disabled": "Disabled"
+    },
+    "stabilityChanged": "Stability changed",
+    "metricTypeChanged": "Metric type changed",
+    "attributeChanges": "Attribute Changes",
+    "metricRemoved": "This metric was removed in the selected \"to\" version."
+  },
+  "diffAttributeTable": {
+    "ariaLabel": "Attribute changes",
+    "columns": {
+      "status": "Status",
+      "key": "Key",
+      "type": "Type"
+    },
+    "status": {
+      "added": "Added",
+      "removed": "Removed",
+      "modified": "Modified"
     }
   },
   "explore": {
--- a/ecosystem-explorer/src/features/collector/collector-detail-page.tsx
+++ b/ecosystem-explorer/src/features/collector/collector-detail-page.tsx
@@ -23,6 +23,7 @@ import {
   AlertTriangle,
   Check,
   Activity,
+  Gauge,
   BookOpen,
   Flag,
 } from "lucide-react";
@@ -46,6 +47,7 @@ import {
 import { CollectorTelemetryTab } from "./components/collector-telemetry-tab";
 import { CollectorFeatureGatesTab } from "./components/collector-feature-gates-tab";
 import { CollectorReadmeTab } from "./components/collector-readme-tab";
+import { TelemetryComparisonSection } from "./components/telemetry-comparison/telemetry-comparison-section";
 
 const getBadgeVariant = (level: string): "success" | "info" | "warning" | "muted" => {
   const lower = level.toLowerCase();
@@ -91,6 +93,7 @@ export function CollectorDetailPage() {
   } = useCollectorComponent(distribution ?? "", name ?? "", version);
   const error = componentError ?? (deprecatedView ? deprecationsQuery.error : null);
   const [activeTab, setActiveTab] = useState("details");
+  const [showInternalTelemetryComparison, setShowInternalTelemetryComparison] = useState(false);
 
   const getStabilityLabel = (level: string) =>
     t(`detail.stabilityLabels.${level.toLowerCase()}`, { defaultValue: level });
@@ -291,6 +294,16 @@ export function CollectorDetailPage() {
                           value: "telemetry",
                           label: t("detail.tabs.telemetry"),
                           icon: <Activity className="h-4 w-4" aria-hidden="true" />,
+                        },
+                      ]
+                    : []),
+                  ...(component.telemetry?.metrics &&
+                  Object.keys(component.telemetry.metrics).length > 0
+                    ? [
+                        {
+                          value: "internal-telemetry",
+                          label: t("detail.tabs.internalTelemetry"),
+                          icon: <Gauge className="h-4 w-4" aria-hidden="true" />,
                         },
                       ]
                     : []),
@@ -537,6 +550,62 @@ export function CollectorDetailPage() {
                 />
               </TabsContent>
             )}
+
+            {component.telemetry?.metrics &&
+              Object.keys(component.telemetry.metrics).length > 0 && (
+                <TabsContent value="internal-telemetry" className="mt-0 p-6">
+                  <div className="space-y-8">
+                    <div className="flex justify-center">
+                      <div
+                        className="border-border inline-flex w-full rounded-lg border bg-transparent p-1 sm:w-auto"
+                        role="group"
+                      >
+                        <button
+                          type="button"
+                          onClick={() => setShowInternalTelemetryComparison(false)}
+                          aria-pressed={!showInternalTelemetryComparison}
+                          className={`flex-1 rounded-lg px-4 py-2 text-sm font-medium transition-all duration-300 ease-in-out sm:flex-none ${
+                            !showInternalTelemetryComparison
+                              ? "bg-card text-foreground shadow-sm"
+                              : "text-muted-foreground hover:text-foreground"
+                          }`}
+                        >
+                          {t("detail.view.current")}
+                        </button>
+                        <button
+                          type="button"
+                          onClick={() => setShowInternalTelemetryComparison(true)}
+                          aria-pressed={showInternalTelemetryComparison}
+                          className={`flex-1 rounded-lg px-4 py-2 text-sm font-medium transition-all duration-300 ease-in-out sm:flex-none ${
+                            showInternalTelemetryComparison
+                              ? "bg-card text-foreground shadow-sm"
+                              : "text-muted-foreground hover:text-foreground"
+                          }`}
+                        >
+                          {t("detail.view.comparison")}
+                        </button>
+                      </div>
+                    </div>
+
+                    {!showInternalTelemetryComparison ? (
+                      <CollectorTelemetryTab
+                        metrics={component.telemetry.metrics}
+                        attributes={component.attributes}
+                        resourceAttributes={component.resource_attributes}
+                      />
+                    ) : (
+                      versionData && (
+                        <TelemetryComparisonSection
+                          distribution={component.distribution}
+                          name={component.name}
+                          versions={versionData.versions}
+                          currentVersion={version}
+                        />
+                      )
+                    )}
+                  </div>
+                </TabsContent>
+              )}
 
             {component.feature_gates && component.feature_gates.length > 0 && (
               <TabsContent value="feature-gates" className="mt-0 p-6">
--- a/ecosystem-explorer/src/features/collector/components/telemetry-comparison/attribute-diff-list.tsx
+++ b/ecosystem-explorer/src/features/collector/components/telemetry-comparison/attribute-diff-list.tsx
@@ -1,0 +1,139 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Plus, Minus } from "lucide-react";
+import { useTranslation } from "react-i18next";
+import type { CollectorAttributeChanges } from "@/types/collector";
+
+interface AttributeDiffListProps {
+  changes: CollectorAttributeChanges;
+}
+
+export function AttributeDiffList({ changes }: AttributeDiffListProps) {
+  const { t } = useTranslation("collector");
+  const hasChanges =
+    changes.added.length > 0 || changes.removed.length > 0 || changes.changed.length > 0;
+
+  if (!hasChanges) {
+    return null;
+  }
+
+  return (
+    <div className="border-border/30 overflow-hidden rounded-lg border">
+      <table aria-label={t("diffAttributeTable.ariaLabel")} className="w-full border-collapse">
+        <thead>
+          <tr className="bg-muted/30">
+            <th
+              scope="col"
+              className="text-muted-foreground p-3 text-left text-[10px] font-bold tracking-widest uppercase"
+            >
+              {t("diffAttributeTable.columns.status")}
+            </th>
+            <th
+              scope="col"
+              className="text-muted-foreground p-3 text-left text-[10px] font-bold tracking-widest uppercase"
+            >
+              {t("diffAttributeTable.columns.key")}
+            </th>
+            <th
+              scope="col"
+              className="text-muted-foreground p-3 text-left text-[10px] font-bold tracking-widest uppercase"
+            >
+              {t("diffAttributeTable.columns.type")}
+            </th>
+          </tr>
+        </thead>
+        <tbody>
+          {changes.added.map((attr, index) => (
+            <tr key={`added-${attr.key}`} className={index % 2 === 1 ? "bg-muted/20" : ""}>
+              <td className="p-4">
+                <div className="flex items-center gap-2">
+                  <Plus className="h-3 w-3 text-green-600 dark:text-green-400" aria-hidden="true" />
+                  <span className="text-xs font-medium text-green-700 dark:text-green-400">
+                    {t("diffAttributeTable.status.added")}
+                  </span>
+                </div>
+              </td>
+              <td className="p-4 font-mono text-sm md:text-[12px]">
+                {attr.definition?.name_override ?? attr.key}
+              </td>
+              <td className="p-4">
+                <span className="bg-muted/50 text-foreground/70 inline-block w-fit rounded px-2 py-1 text-xs font-bold uppercase">
+                  {attr.definition?.type ?? "—"}
+                </span>
+              </td>
+            </tr>
+          ))}
+
+          {changes.removed.map((attr, index) => (
+            <tr
+              key={`removed-${attr.key}`}
+              className={(index + changes.added.length) % 2 === 1 ? "bg-muted/20" : ""}
+            >
+              <td className="p-4">
+                <div className="flex items-center gap-2">
+                  <Minus className="h-3 w-3 text-red-600 dark:text-red-400" aria-hidden="true" />
+                  <span className="text-xs font-medium text-red-700 dark:text-red-400">
+                    {t("diffAttributeTable.status.removed")}
+                  </span>
+                </div>
+              </td>
+              <td className="p-4 font-mono text-sm line-through opacity-60 md:text-[12px]">
+                {attr.definition?.name_override ?? attr.key}
+              </td>
+              <td className="p-4">
+                <span className="bg-muted/50 text-foreground/70 inline-block w-fit rounded px-2 py-1 text-xs font-bold uppercase opacity-60">
+                  {attr.definition?.type ?? "—"}
+                </span>
+              </td>
+            </tr>
+          ))}
+
+          {changes.changed.map((change, index) => (
+            <tr
+              key={`changed-${change.key}`}
+              className={
+                (index + changes.added.length + changes.removed.length) % 2 === 1
+                  ? "bg-muted/20"
+                  : ""
+              }
+            >
+              <td className="p-4">
+                <span className="text-xs font-medium text-orange-700 dark:text-orange-400">
+                  {t("diffAttributeTable.status.modified")}
+                </span>
+              </td>
+              <td className="p-4 font-mono text-sm md:text-[12px]">
+                {change.after?.name_override ?? change.before?.name_override ?? change.key}
+              </td>
+              <td className="p-4">
+                <div className="flex items-center gap-2">
+                  <span className="bg-muted/50 text-foreground/70 inline-block w-fit rounded px-2 py-1 text-xs font-bold uppercase line-through opacity-60">
+                    {change.before?.type ?? "—"}
+                  </span>
+                  <span className="text-muted-foreground text-xs">→</span>
+                  <span className="bg-muted/50 text-foreground/70 inline-block w-fit rounded px-2 py-1 text-xs font-bold uppercase">
+                    {change.after?.type ?? "—"}
+                  </span>
+                </div>
+              </td>
+            </tr>
+          ))}
+        </tbody>
+      </table>
+    </div>
+  );
+}
--- a/ecosystem-explorer/src/features/collector/components/telemetry-comparison/diff-results-section.tsx
+++ b/ecosystem-explorer/src/features/collector/components/telemetry-comparison/diff-results-section.tsx
@@ -1,0 +1,84 @@
+/*
+ * Copyright The OpenTelemetry Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { ReactNode } from "react";
+import { useTranslation } from "react-i18next";
+import type { CollectorTelemetryDiffResult } from "@/types/collector";
+import { MetricDiffCard } from "./metric-diff-card";
+
+interface DiffResultsSectionProps {
+  diffResult: CollectorTelemetryDiffResult;
+}
+
+function SectionDivider({ children }: { children: ReactNode }) {
+  return (
+    <div className="flex items-center justify-center gap-4">
+      <div className="to-border h-px w-16 bg-gradient-to-r from-transparent" />
+      <span className="text-muted-foreground text-sm font-medium tracking-wider uppercase">
+        {children}
+      </span>
+      <div className="to-border h-px w-16 bg-gradient-to-l from-transparent" />
+    </div>
+  );
+}
+
+export function DiffResultsSection({ diffResult }: DiffResultsSectionProps) {
+  const { t } = useTranslation("collector");
+  const { metrics } = diffResult;
+
+  const addedOrRemoved = metrics.filter((m) => m.status === "added" || m.status === "removed");
+  const changed = metrics.filter((m) => m.status === "changed");
+  const hasAnyChanges = metrics.some((m) => m.status !== "unchanged");
+
+  if (!hasAnyChanges) {
+    return (
+      <div className="flex min-h-[300px] items-center justify-center">
+        <div className="space-y-2 text-center">
+          <p className="text-muted-foreground text-lg font-medium">
+            {t("diffResults.empty.title")}
+          </p>
+          <p className="text-muted-foreground/70 text-sm">{t("diffResults.empty.description")}</p>
+        </div>
+      </div>
+    );
+  }
+
+  return (
+    <div className="space-y-12">
+      {addedOrRemoved.length > 0 && (
+        <div className="space-y-6">
+          <SectionDivider>{t("diffResults.addedRemovedHeader")}</SectionDivider>
+          <div className="mx-auto max-w-3xl space-y-6">
+            {addedOrRemoved.map((metricDiff) => (
+              <MetricDiffCard key={metricDiff.name} diff={metricDiff} />
+            ))}
+          </div>
+        </div>
+      )}
+
+      {changed.length > 0 && (
+        <div className="space-y-6">
+          <SectionDivider>{t("diffResults.changedHeader")}</SectionDivider>
+          <div className="mx-auto max-w-3xl space-y-6">
+            {changed.map((metricDiff) => (
+              <MetricDiffCard key={metricDiff.name} diff={metricDiff} />
+            ))}
+          </div>
+        </div>
+      )}
+    </div>
+  );
+}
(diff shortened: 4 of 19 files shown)


Adii-45:
Hey @jaydeluca! PR is ready for a look whenever you get a chance! I’ve addressed the review feedback as well. Let me know if there’s anything else you’d like me to add or update, or if I’ve missed anything... Thanks!

ecosystem-explorer/src/features/collector/utils/telemetry-diff.ts
```diff
@@ -0,0 +1,181 @@
...
+/** Shallow-compares the type-specific descriptor (sum/gauge/histogram) of a metric. */
+function metricDescriptorEqual(from: CollectorMetric, to: CollectorMetric): boolean {
+  const fromType = getMetricType(from);
+  const toType = getMetricType(to);
+  if (fromType !== toType) return false;
+  if (fromType === null) return true;
+
+  const fromDescriptor = from[fromType];
+  const toDescriptor = to[toType as "sum" | "gauge" | "histogram"];
+  return JSON.stringify(fromDescriptor) === JSON.stringify(toDescriptor);
```
jaydeluca: `metricDescriptorEqual` detects any difference inside the descriptor, but only the instrument kind is recorded (lines 141-143). A `sum.monotonic: true → false` flip renders as `Metric type changed sum→sum` and `monotonic` appears nowhere. Same for `value_type`, `aggregation_temporality` and `bucket_boundaries` — `telemetry-diff.test.ts:175-198` shows the boundaries being dropped.

`JSON.stringify` is also key-order sensitive: `{monotonic, value_type}` vs `{value_type, monotonic}` reports as changed.

Fix: emit `metricType` only when `getMetricType(from) !== getMetricType(to)`, and add a field-level descriptor change entry compared field-by-field instead of by `stringify`. `telemetry-diff.test.ts:159-172` asserts the `{before: "sum", after: "sum"}` shape and needs updating with it.

ecosystem-explorer/src/features/collector/utils/telemetry-diff.ts
```diff
@@ -0,0 +1,181 @@
...
+  const attributeChanges = compareAttributes(fromMetric, toMetric, fromAttributes, toAttributes);
+  const attributesChanged =
+    attributeChanges.added.length > 0 ||
+    attributeChanges.removed.length > 0 ||
+    attributeChanges.changed.length > 0;
+
+  const descriptionChanged = fromMetric.description !== toMetric.description;
+  const unitChanged = fromMetric.unit !== toMetric.unit;
+  const enabledChanged = fromMetric.enabled !== toMetric.enabled;
+  const stabilityChanged = fromMetric.stability !== toMetric.stability;
```
jaydeluca, line 177: These six fields are all that gets compared. Anything else falls through to `status: "unchanged"` on line 124 and disappears from the diff with no warning.

Occurrences across `ecosystem-registry/collector/**` (excluding `meta/`), 1984 telemetry metric entries:

| field | occurrences | on `CollectorMetric`? | compared? |
| --- | --- | --- | --- |
| `prefix` | 301 | no | no |
| `extended_documentation` | 160 | yes | no |
| `optional` | 88 | yes | no |
| `warnings` | 48 | no | no |
| `deprecated` | 14 | no | no |

`collector_transformer.py:133` assigns `component["telemetry"] = telemetry`, so all five reach the frontend regardless of the TS type.

Fix: compare `extended_documentation` and `optional` — both already on `CollectorMetric`. For `prefix` and `deprecated`, add them to the type and re-run `bun run generate-schemas`. `prefix` participates in the exported metric name, so a `prefix` change is a rename this currently reports as "No differences found".

ecosystem-explorer/public/locales/es/collector.json
```diff
@@ -280,6 +285,68 @@
       "reference": "Referencia"
     }
   },
+  "telemetryComparison": {
```
jaydeluca, line 292: This block and the four that follow (`versionSelectorPanel`, `diffResults`, `diffCard`, `diffAttributeTable`), plus `detail.view` and `detail.tabs.internalTelemetry`, are English — 37 keys, all identical to `en/collector.json`. `locale-parity.test.ts` passes because it compares key sets, not values.

ecosystem-explorer/src/features/collector/hooks/use-telemetry-comparison.ts
```diff
@@ -0,0 +1,141 @@
...
+        if (cancelled) return;
+
+        const fromComponent = fromResult.status === "fulfilled" ? fromResult.value : null;
+        const fromLoadFailed = fromResult.status === "rejected";
+
+        const toComponent = toResult.status === "fulfilled" ? toResult.value : null;
+        const toLoadFailed = toResult.status === "rejected";
+
+        if (fromLoadFailed && toLoadFailed) {
+          setError(new Error("Failed to load both versions for comparison"));
```
jaydeluca: Rendered to users at `telemetry-comparison-section.tsx:87` via `{error.message}`, so it needs a locale key. Java uses `t("telemetryComparison.error.bothVersionsFailed")` (`java-agent/hooks/use-telemetry-comparison.ts:107`, key in both `en`/`es` `java-agent.json:325`). This port added `telemetryComparison.error.title` but no message key.

ecosystem-explorer/src/features/collector/components/telemetry-comparison/metric-diff-card.tsx
```diff
@@ -0,0 +1,162 @@
...
+              </code>
+              <span className="text-muted-foreground">→</span>
+              <code className="rounded border border-green-500/30 bg-green-500/10 px-2 py-1 text-sm text-green-700 dark:text-green-400">
+                {changes.metricType.after ?? "—"}
+              </code>
+            </div>
+          </div>
+        )}
+
+        {status === "changed" && changes?.attributes && (
```
jaydeluca: `changes.attributes` is non-optional and always constructed (`telemetry-diff.ts:127`), so this is always true for a changed metric, while `AttributeDiffList` returns `null` when all three buckets are empty. A unit-only change renders the "Attribute Changes" heading with no table under it. Guard on `added.length || removed.length || changed.length`.

ecosystem-explorer/src/features/collector/components/telemetry-comparison/attribute-diff-list.tsx
```diff
@@ -0,0 +1,139 @@
...
+                <span className="text-xs font-medium text-orange-700 dark:text-orange-400">
+                  {t("diffAttributeTable.status.modified")}
+                </span>
+              </td>
+              <td className="p-4 font-mono text-sm md:text-[12px]">
+                {change.after?.name_override ?? change.before?.name_override ?? change.key}
+              </td>
+              <td className="p-4">
+                <div className="flex items-center gap-2">
+                  <span className="bg-muted/50 text-foreground/70 inline-block w-fit rounded px-2 py-1 text-xs font-bold uppercase line-through opacity-60">
```
jaydeluca: `attributesEqual` compares `type`, `description`, `name_override` and `enum` (`telemetry-diff.ts:29-37`), but this row renders only `type`, so a description- or enum-only change shows `Modified · string → string`. Line 120 also prefers `change.after?.name_override`, so a rename displays only the new name.

ecosystem-explorer/src/features/collector/collector-detail-page.tsx
```diff
@@ -538,6 +551,62 @@ export function CollectorDetailPage() {
...
+                    </div>
+
+                    {!showInternalTelemetryComparison ? (
+                      <CollectorTelemetryTab
+                        metrics={component.telemetry.metrics}
+                        attributes={component.attributes}
+                        resourceAttributes={component.resource_attributes}
+                      />
+                    ) : (
+                      versionData && (
```
jaydeluca: If `useCollectorVersions()` errors while `?version=` is in the URL, `version` is non-empty so the page renders, but `versionData` is `null` — "Version Comparison" then renders nothing, no loader and no error. Same shape as the Java page (`instrumentation-detail-page.tsx:604-612`). No test clicks "Version Comparison", so this branch is uncovered.

jaydeluca (commented):
Three things to fix, two of which make the diff tool report false results:

1. A descriptor change within the same instrument kind renders as "Metric type changed: sum → sum"
   and the actual change is dropped. Inline.
2. `extended_documentation`, `optional`, `prefix`, `deprecated` and `warnings` are never compared, so
   metrics that changed report as `unchanged` and vanish from the diff.

3. All 37 new keys in `public/locales/es/collector.json` are verbatim English.

Two notes on files not in the diff, so they can't be inline:

- `ecosystem-explorer/src/i18n/locale-parity.test.ts` checks key parity only, so it cannot catch #3.
- The PR description lists "Pass through Collector `telemetry` metadata into the Explorer database
  output" and "Add Collector transformer tests covering telemetry passthrough", but there are no
  `ecosystem-automation/` changes here — that landed as #945 (`0d4bba06`, already on `main`). Worth
  confirming nothing was meant to be in this PR and got dropped.

ecosystem-explorer/src/features/collector/utils/telemetry-diff.ts
Adii-45: Addressed this by comparing the descriptor fields individually and only reporting `metricType` when the instrument type actually changes. Added regression coverage as well.

ecosystem-explorer/public/locales/es/collector.json
Adii-45: Fixed the Spanish translation so the new telemetry comparison strings are properly localized and consistent with the existing UI labels.

Adii-45:
@jaydeluca All requested changes have been addressed!

- Updated the telemetry diff logic to avoid the false-positive descriptor changes and ignore the non-comparable fields you called out
- Corrected the Spanish telemetry comparison translations
- Verified the related tests and formatting checks pass
- Updated the PR description to remove the DB passthrough and transformer-test items already covered by #945

Thanks for the detailed review! Please take another look whenever you get a chance..

ecosystem-explorer/src/features/collector/collector-detail-page.tsx
```diff
@@ -538,6 +551,83 @@ export function CollectorDetailPage() {
...
+                      <CollectorTelemetryTab
+                        metrics={component.telemetry.metrics}
+                        attributes={component.attributes}
+                        resourceAttributes={component.resource_attributes}
+                      />
+                    ) : versionData ? (
+                      <TelemetryComparisonSection
+                        distribution={component.distribution}
+                        name={component.name}
+                        versions={versionData.versions}
```
jaydeluca, line 600: **Version dropdowns list every release, not the ones this component exists in**

This passes the **global** release index. `collector-data.ts:133-146` documents exactly this hazard and exists to solve it:

> The global versions index lists every Collector release, but a component is only present from the release that introduced it onward […] Callers that build per-version links need this list rather than the global one — otherwise a link resolves to a version whose manifest has no entry for the component and `loadComponent` throws.

`loadComponentVersions` is currently referenced only by `collector-data.test.ts` — no app code uses it.

This isn't only a hand-selection edge case, it's the **default** state: `telemetry-comparison-section.tsx:42-45` picks `versions[currentIndex + 1]` off the global list, so for a component introduced in the latest release the default `From` is a version where it doesn't exist. Measured against the shipped `public/data/collector/`, 3 of the 34 components carrying internal telemetry are absent from at least one of the 7 published releases:

| Component | Present in |
| --- | --- |
| `contrib/adaptivetailsamplingprocessor` | 0.160.0 only |
| `contrib/dynamicsamplingprocessor` | 0.156.0–0.159.0 |
| `contrib/rollingspanlatencyprocessor` | registry only (0.160.1-SNAPSHOT) |

Reproducible with data already in the repo — `/collector/components/contrib/adaptivetailsamplingprocessor` → Internal Telemetry → Version Comparison:

```
$ python3 -c "import json,glob; [print(f.split('/')[-1], 'contrib-adaptivetailsamplingprocessor' in json.load(open(f))['components']) for f in sorted(glob.glob('ecosystem-explorer/public/data/collector/versions/*-index.json'))]"
0.154.0-index.json False
...
0.159.0-index.json False
0.160.0-index.json True
```

`From` defaults to `0.159.0`, whose manifest has no entry, so `loadComponent` throws (`collector-data.ts:166`), `fromNotFound` flips, and the user's first view of the feature is "Could not load internal telemetry for version 0.159.0" above all 10 of the component's metrics listed as **Added**.

Fix: wrap `loadComponentVersions(distribution, name)` in a hook and feed the comparison section that list instead. The Java page has the same shape (`instrumentation-detail-page.tsx:610` also passes the global list), so scoping the fix to Collector is reasonable — Collector is where the helper already exists and where the data demonstrably diverges.

ecosystem-explorer/src/features/collector/components/telemetry-comparison/telemetry-comparison-section.tsx
```diff
@@ -0,0 +1,146 @@
...
+            <AlertCircle
+              className="mt-0.5 h-5 w-5 flex-shrink-0 text-orange-600 dark:text-orange-400"
+              aria-hidden="true"
+            />
+            <div className="space-y-1">
+              <p className="font-medium text-orange-700 dark:text-orange-400">
+                {t("telemetryComparison.warnings.availability.title")}
+              </p>
+              {fromNotFound && (
+                <p className="text-sm text-orange-700/80 dark:text-orange-400/80">
```
jaydeluca: **The `/80` on the alert body text misses WCAG AA in both themes**

The `/80` opacity on each alert's secondary line drops it below 4.5:1. Computed against the `themes.ts` surfaces (light `background: 0 0% 100%`, dark `232 38% 15%`), composited through the `/10` tint:

| Class | Surface | Ratio | AA (4.5:1) |
| --- | --- | --- | --- |
| `text-orange-700/80` | `bg-orange-500/10` light | **3.44** | fail |
| `text-red-700/80` | `bg-red-500/10` light | **4.21** | fail |
| `text-red-400/80` | `bg-red-500/10` dark | **4.06** | fail |
| `text-orange-400/80` | `bg-orange-500/10` dark | 4.71 | pass |

Dropping `/80` clears all four (orange-700 4.68, red-700 5.68, orange-400 6.54, red-400 5.54). It's all `text-sm` body copy, so the large-text 3:1 allowance doesn't apply. The title lines above each of these, at full opacity, already pass — it's only the `/80` variants.

Five occurrences, all new in this PR: this line, `:111`, `:132`, `:87` (`text-red-700/80`), and `collector-detail-page.tsx:615`.

Worth noting the Java original this is ported from (`java-agent/.../telemetry-comparison-section.tsx:87,104,109,128`) is dark-only — `text-red-400/80`, `text-yellow-400/80`, no light variant — so it isn't precedent for the light half of the pair introduced here. `glow-badge.tsx:27-32` already carries the repo's rule for this decision:

> Light-mode text shades must clear WCAG AA (4.5:1) against the /10 tint composited over the app's tinted light surfaces (slate list rows, not just white cards)

ecosystem-explorer/src/features/collector/components/telemetry-comparison/metric-diff-card.tsx
```diff
@@ -0,0 +1,208 @@
...
+                      {format(fieldChange.after)}
+                    </code>
+                  </div>
+                );
+              })}
+            </div>
+          </div>
+        )}
+
+        {status === "changed" &&
```
jaydeluca, line 331: **Four of the newly compared fields are never rendered — a "Changed" card with nothing changed**

The fix for my previous comment added `extendedDocumentation`, `optional`, `prefix` and `deprecated` to `CollectorMetricChanges` (`telemetry-diff.ts:216-230`), so those metrics now correctly leave the `unchanged` bucket. But this card renders only `description`, `unit`, `enabled`, `stability`, `metricType`, `descriptor` and `attributes`:

```
$ grep -rn "prefix\|deprecated\|extendedDocumentation\|optional" src/features/collector/components/telemetry-comparison/*.tsx
metric-diff-card.test.tsx:33:    // Regression guard: `changes.attributes` is always constructed (non-optional), so a
```

One hit, in a test comment. There are no `diffCard.*` locale keys for them either.

So a metric whose only change is one of those four now renders under **Changed** as a card showing its name, a "Changed" badge, its description and its unit — with no indication of what changed. My previous comment traded "the change vanishes" for "the change appears but is unexplained"; that's on me for asking for the comparison without asking for the rendering.

`prefix` is the one that matters most: it participates in the exported metric name, so a `prefix` change is a rename presented as an unexplained "Changed".

Not reachable with current data — across all 28 version pairs of the 8 registry releases, the only metric fields that ever differ are `description` (16), `unit` (15) and `histogram` (15). So this is latent and fine as a follow-up. Either render the four (4 keys × 2 locales, same before/after row shape as the `stability` block at `:111`), or drop them from the diff until the card can show them — computing them without rendering them is the one combination that produces a blank card.

ecosystem-explorer/src/features/collector/utils/telemetry-diff.ts
```diff
@@ -0,0 +1,268 @@
...
+  const extendedDocumentationChanged =
+    fromMetric.extended_documentation !== toMetric.extended_documentation;
+  const optionalChanged = fromMetric.optional !== toMetric.optional;
+  const prefixChanged = fromMetric.prefix !== toMetric.prefix;
+  const deprecatedChanged = !deprecatedEqual(fromMetric.deprecated, toMetric.deprecated);
```
jaydeluca, line 213: **`warnings` is still uncompared, untyped, and missing from the published schema**

Four of the five fields from my previous table are now compared — `warnings` is the one left. My own fix paragraph named only the other four, so this is a gap in my comment rather than something you skipped; flagging it so it's a decision rather than an oversight.

48 internal-telemetry metric entries across `ecosystem-registry/collector/**` carry it, and it holds user-facing deprecation guidance:

```yaml
# contrib/v0.157.0/exporter.yaml — kafkaexporter kafka_broker_throttling_duration
warnings:
  if_enabled: This metric is deprecated in favor of kafka_broker_throttling_latency and will be removed soon.
```

It's not on `CollectorMetric`, not in `collector-component.schema.json`, and not compared here — so a metric whose only change is a new or reworded `warnings` entry still falls through to `status: "unchanged"` on line 199 and disappears from the diff. `collector_transformer.py:133` assigns `component["telemetry"] = telemetry` wholesale, so the field does reach the frontend regardless of the TS type.

Two parts, both low urgency:

- **The diff gap.** Same silent-drop class as the fix you just applied. No `warnings`-only change occurs in the 8 available registry versions, so latent.
- **The published schema.** `public/schemas/collector-component.schema.json` is served to agent consumers (`generate-agent-docs.mjs:52`, `generate-seo.mjs:73`) rather than used for validation, so an omitted field means the published contract under-describes the JSON those consumers actually receive.

Shape is `{ if_enabled?: string, if_configured?: string, if_enabled_not_set?: string }` upstream — worth checking the full set before adding it to the type, since the registry only exercises `if_enabled`.

ecosystem-explorer/src/features/collector/components/telemetry-comparison/metric-diff-card.tsx
```diff
@@ -0,0 +1,208 @@
...
+          <div className="border-border/30 flex flex-wrap items-center gap-3 border-b pb-6">
+            <span className="text-muted-foreground text-xs font-bold tracking-widest uppercase">
+              {t("diffCard.unit")}
+            </span>
+            <code className="border-border/30 text-foreground/80 rounded border bg-black/[0.02] px-2 py-1 text-sm dark:bg-white/[0.03]">
+              {metric.unit || "1"}
+            </code>
+            {status === "changed" && changes?.unit && (
+              <div className="flex items-center gap-2">
+                <span className="text-muted-foreground text-xs">(was:</span>
```
jaydeluca: **Hardcoded English `(was:` (suggestion)**

Minor: `(was:` here and `)` on line 84 are hardcoded English in a file that otherwise routes everything through `t()`. Two separate problems — the literal bypasses extraction (AGENTS.md: "Never hardcode user-visible strings"), and splitting a parenthetical across two JSX nodes isn't localizable even after extraction, since the whole grammatical unit has to be one key:

```json
"unitWas": "(was: {{unit}})"
```

Ported verbatim from `java-agent/.../metric-diff-card.tsx:106`, which is the only other occurrence in `src/` — fixing both is the tidier change if you want it in scope, or leave it and I'll file it separately.

ecosystem-explorer/src/features/collector/components/telemetry-comparison/attribute-diff-list.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+  if (!hasChanges) {
+    return null;
+  }
+
+  return (
+    <div className="border-border/30 overflow-hidden rounded-lg border">
+      <table aria-label={t("diffAttributeTable.ariaLabel")} className="w-full border-collapse">
+        <thead>
+          <tr className="bg-muted/30">
```
jaydeluca: **Striped-table shades deviate from DESIGN.md (suggestion)**

The header is `bg-muted/30` and odd rows are `bg-muted/20` (line 82). DESIGN.md §"Striped Tables" specifies `bg-white/5` for the header and `bg-muted/40` for odd rows, which `java-agent/.../attribute-diff-table.tsx:38,62` follows.

Worth saying that `bg-white/5` is effectively invisible in light mode, so `bg-muted/30` may well be the deliberate improvement here — in which case it's DESIGN.md that's stale and the doc is the thing to update rather than this file. Either way the two implementations now disagree, and neither is wrong by the doc as written, so it's worth settling which one is the pattern.

jaydeluca (commented):
**`public/locales/{en,es}/collector.json` need 4 new keys if you take the
`metric-diff-card.tsx:187` finding.** `extendedDocumentation`, `optional`, `prefix` and `deprecated`
are now computed as changes but have no `diffCard.*` label keys in either locale, which is why the
card can't render them.

ecosystem-explorer/src/features/collector/collector-detail-page.tsx
Adii-45: Addressed - the version comparison dropdown now uses the versions where the component is actually present, rather than the global Collector release list

ecosystem-explorer/src/features/collector/components/telemetry-comparison/telemetry-comparison-section.tsx
Adii-45: Addressed - updated the alert text opacity classes to meet the WCAG contrast requirement in both themes

ecosystem-explorer/src/features/collector/components/telemetry-comparison/metric-diff-card.tsx
Adii-45: Addressed - the comparison now renders all changed fields, including prefix/deprecated/extended documentation/optional changes, so a field-level change is no longer shown as an unexplained “Changed” card

ecosystem-explorer/src/features/collector/utils/telemetry-diff.ts
Adii-45: Addressed - added `warnings` to the Collector metric types/schema and included it in the comparison logic and UI rendering. Added test coverage for warnings-only changes as well

ecosystem-explorer/src/features/collector/components/telemetry-comparison/metric-diff-card.tsx
Adii-45: Addressed - replaced the hardcoded `was:` text with the localized `diffCard.unitWas` translation key

ecosystem-explorer/src/features/collector/components/telemetry-comparison/attribute-diff-list.tsx
Adii-45: Addressed - updated the attribute diff table row/header styling to follow the existing striped-table pattern from `DESIGN.md`

Adii-45:
Hey @jaydeluca! Addressed all the review feedback from this round!

- Fixed component-specific version selection for comparisons
- Added the missing `warnings` field to the types/schema/diff/rendering
- Fixed changed-field rendering so all supported field changes are surfaced
- Replaced the hardcoded `was:` text with i18n
- Updated attribute diff table styling to match the existing pattern
- Fixed the alert text contrast issue
- Added/updated the relevant test coverage

Everything is pushed and the checks are passing! PTAL whenever you get a chance..

ecosystem-explorer/src/features/collector/components/telemetry-comparison/telemetry-comparison-section.tsx
```diff
@@ -0,0 +1,193 @@
...
+          <div className="flex items-start gap-3">
+            <AlertCircle
+              className="mt-0.5 h-5 w-5 flex-shrink-0 text-red-600 dark:text-red-400"
+              aria-hidden="true"
+            />
+            <div className="space-y-1">
+              <p className="font-medium text-red-700 dark:text-red-400">
+                {t("telemetryComparison.error.title")}
+              </p>
+              <p className="text-sm text-red-700/80 dark:text-red-400/80">
```
jaydeluca: **The `/80` contrast finding is unaddressed, and this block added a sixth occurrence**

The scoping fix introduced a new `componentVersionsError` block that reuses the exact pattern from my
previous comment, so the count went 5 → 6:

```
$ grep -rn "text-\(red\|orange\|yellow\|green\)-[0-9]*/80" src/features/collector/
collector-detail-page.tsx:615                     text-red-700/80 dark:text-red-400/80
telemetry-comparison-section.tsx:96               text-red-700/80 dark:text-red-400/80   <-- new
telemetry-comparison-section.tsx:134              text-red-700/80 dark:text-red-400/80
telemetry-comparison-section.tsx:153              text-orange-700/80 dark:text-orange-400/80
telemetry-comparison-section.tsx:158              text-orange-700/80 dark:text-orange-400/80
telemetry-comparison-section.tsx:179              text-orange-700/80 dark:text-orange-400/80
```

Ratios from before stand: `text-orange-700/80` **3.44**, `text-red-700/80` **4.21**,
`text-red-400/80` **4.06** against the `/10` tint composited over the `themes.ts` surfaces. All six
are `text-sm` body copy, so the large-text 3:1 allowance doesn't apply. Dropping `/80` clears AA in
both themes (orange-700 4.68, red-700 5.68, orange-400 6.54, red-400 5.54). The full-opacity title
lines above each already pass — it's only the `/80` variants.

ecosystem-explorer/src/features/collector/components/telemetry-comparison/telemetry-comparison-section.tsx
```diff
@@ -0,0 +1,193 @@
...
+                {componentVersionsError.message}
```
jaydeluca: **Raw internal English rendered to users**

These messages come from `collector-data.ts` internals and are hardcoded English:

- `"Collector versions index returned null unexpectedly"` (`collector-data.ts:92`)
- `"Collector manifest for version ${version} returned null unexpectedly"` (`collector-data.ts:125`)

The sibling branch for the same class of failure already does this right —
`collector-detail-page.tsx:613` renders the localized `detail.view.comparisonUnavailable.message`
("Could not load the list of versions needed for comparison."). Reuse that key, or add
`telemetryComparison.error.componentVersionsFailed`. Same class as my first-round comment on
`use-telemetry-comparison.ts:92`, which you fixed with `bothVersionsFailed` — note line 134 just
below here renders that localized message correctly, so the two branches disagree.

`telemetry-comparison-section.test.tsx:222-232` currently pins the raw string via
`getByText("could not load component versions")`, so that assertion moves with the fix.

ecosystem-explorer/src/features/collector/components/telemetry-comparison/telemetry-comparison-section.tsx
```diff
@@ -0,0 +1,193 @@
...
+    ? versions.filter((v) => componentVersionList.includes(v.version))
+    : [];
+
+  // "To" defaults to the version being viewed. "From" defaults to the previous release,
+  // or falls back to currentVersion (triggering a same-version warning) if viewing the oldest version.
+  const currentIndex = scopedVersions.findIndex((v) => v.version === currentVersion);
+  const defaultFromVersion =
+    currentIndex >= 0 && currentIndex < scopedVersions.length - 1
+      ? scopedVersions[currentIndex + 1].version
+      : currentVersion;
```
jaydeluca, line 61: **Correct scoping creates an instruction the user can't follow (new, and my ask caused it)**

Scoping to releases the component exists in means a component present in exactly one release gets one
option in both selects, and the only reachable state is:

> **Same version selected** — Choose two different versions to see a comparison.

Reachable today: `/collector/components/contrib/adaptivetailsamplingprocessor` → Internal Telemetry →
Version Comparison. Sized against the published data:

```
components with internal telemetry: 33
release-count distribution: {1 release: 1, 4 releases: 1, 7 releases: 31}
single-release: ['contrib-adaptivetailsamplingprocessor']
```

One component now, but this is the permanent first-release state of every new telemetry-carrying
component, so it recurs every cycle. Either a distinct message ("only one release includes this
component") or suppressing the Current/Comparison toggle when `scopedVersions.length < 2` beats
telling the user to pick a second version that doesn't exist.

ecosystem-explorer/src/features/collector/utils/telemetry-diff.ts
```diff
@@ -0,0 +1,283 @@
...
+}
+
+function warningsEqual(a?: CollectorMetricWarnings, b?: CollectorMetricWarnings): boolean {
+  if (!a || !b) return a === b;
+  return (
+    a.if_enabled === b.if_enabled &&
+    a.if_enabled_not_set === b.if_enabled_not_set &&
+    a.if_configured === b.if_configured
+  );
+}
```
jaydeluca, line 189: **`warnings` comparison is keyed to the three known fields (suggestion, not reachable today)**

This and the render loop at `metric-diff-card.tsx:199-225` both enumerate `if_enabled` /
`if_enabled_not_set` / `if_configured` explicitly. Two latent consequences:

- `warningsEqual(undefined, {})` returns `false`, while every per-field comparison in the render loop
  compares equal — so an empty `warnings: {}` renders the "Warnings changed" heading above an empty
  container. Same blank-section class as the `changes.attributes` guard you fixed last round.
- A fourth upstream condition field would compare equal and fall through to `status: "unchanged"` —
  the silent-drop class again.

Checked the shipped data before raising it: 27 metrics carry `warnings`, all `if_enabled` only, zero
empty objects. So this is latent and fine as-is. Comparing over the union of both objects' keys would
close both at once if you want it cheap.

ecosystem-explorer/src/features/collector/components/telemetry-comparison/diff-results-section.tsx
```diff
@@ -0,0 +1,84 @@
...
+import type { ReactNode } from "react";
+import { useTranslation } from "react-i18next";
+import type { CollectorTelemetryDiffResult } from "@/types/collector";
+import { MetricDiffCard } from "./metric-diff-card";
+
+interface DiffResultsSectionProps {
+  diffResult: CollectorTelemetryDiffResult;
+}
+
+function SectionDivider({ children }: { children: ReactNode }) {
```
jaydeluca: **Third copy of `SectionDivider` (suggestion)**

Byte-identical to `java-agent/.../diff-results-section.tsx:28-38`, while
`@/components/ui/section-divider` — whose import `collector-telemetry-tab.tsx` just dropped in this
PR — is a visually different third variant.

jaydeluca (commented):
just a few more things and this should be good

ecosystem-explorer/src/features/collector/utils/telemetry-diff.ts
```diff
@@ -0,0 +1,290 @@
...
+ * Collector internal telemetry is metrics-only and unconditional.
+ */
+export function compareCollectorTelemetry(
+  fromComponent: CollectorComponent | null,
+  toComponent: CollectorComponent | null
+): CollectorTelemetryDiffResult {
+  const fromMetrics = fromComponent?.telemetry?.metrics ?? {};
+  const toMetrics = toComponent?.telemetry?.metrics ?? {};
+  const fromAttributes = fromComponent?.attributes;
+  const toAttributes = toComponent?.attributes;
```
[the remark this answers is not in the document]
Adii-45: Fixed in the latest commit!

ecosystem-explorer/src/features/collector/components/telemetry-comparison/metric-diff-card.tsx
```diff
@@ -0,0 +1,322 @@
...
+              {t("diffCard.warningsChanged")}
+            </span>
+            <div className="space-y-2">
+              {(
+                [
+                  ["if_enabled", "ifEnabled"],
+                  ["if_enabled_not_set", "ifEnabledNotSet"],
+                  ["if_configured", "ifConfigured"],
+                ] as const
+              ).map(([field, labelKey]) => {
```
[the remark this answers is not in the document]
Adii-45: Fixed in the latest commit!

Adii-45:
@jaydeluca All reviews have been addressed... Ready for a final look whenever you get a chance!

jaydeluca (approved):

PR #1066, opened 2026-09-05
test(js-watcher): add coverage for instrumentation sync

MeloveGupta:
## What changed

I have added a new file ```test_instrumentation_sync.py```. This is the second and last PR for the watcher test coverage, ```instrumentation_sync.py``` was the third untested module after ```main.py``` and ```repository_manager.py``` in #1049 .

The tests cover the branching in ```sync()```, so a package that isn't in the registry yet gets saved, one that is gets skipped, and the three ways a package can end up in ```failed``` (parser throws, parser returns None, or there's no version in the parsed data). Also covers that bundle membership and component owners get loaded once and not once per package, and that they get passed through to each ```PackageParser```.

## Why

Part of #990 

One of the acceptance criteria there is test coverage on the watcher orchestration. With this the three untested modules are done, ```instrumentation_sync.py``` goes from 22% to 100% and the watcher goes from 75% to 89%.

## How it was tested

I ran ```uv run pytest tests/ --cov=js_instrumentation_watcher --cov-report=term-missing``` from ```ecosystem-automation/js-instrumentation-watcher```

Screenshot from 2026-09-05 22-00-06

I also ran ```uv run pytest ecosystem-automation -q``` for the whole Python suite too, 935 passed.

Screenshot from 2026-09-05 22-02-50

## Is this a breaking change?

No

## Special notes for your reviewer

NA

## Checklist

- [x] Tests added or updated for new behavior
- [x] Screenshots attached for any UI changes
- [x] I wrote this PR description myself (AI tools must not check this box on behalf of the author)

Issue #990, opened 2026-08-19
[JS Instrumentation] Wire the registry into explorer-db-builder and the frontend (continuing #9)


MeloveGupta:
### Component / area

JavaScript instrumentation ecosystem - js-instrumentation-watcher, explorer-db-builder, frontend

### Problem this solves

The JS instrumentation watcher (#9 , #577 ) runs nightly and tracks 48 packages in ecosystem-registry/javascript/, but that data never reaches a user. explorer-db-builder's --ecosystem flag only knows javaagent | configuration | collector, and the frontend has no javascript/ feature, API layer, or route. Data is collected in production every night and goes nowhere.

### Proposed solution

Opening this as its own issue since #9 thread has gone quiet since May, but this is meant as a direct continuation of it, not a separate effort.

Closing the loop first: you'd asked for a deeper pass on telemetry data coverage across the READMEs before deciding next steps. I did that in May and never posted it back, sorry for the delay.

| Coverage level | Count | Example |
|---|---|---|
| Structured table (parseable) | 1/47 | aws-sdk |
| Structured headings, inconsistent format | 7/47 | cassandra-driver, kafkajs, pg, others |
| Prose only, not parseable | 39/47 | most packages |

(Counts are from the May audit, projects/9-javascript-instrumentation/02-telemetry-coverage.md - the registry's grown to 48 packages since, so today's total is one higher than the /47 above.)
Bottom line: telemetry data isn't reliably extractable from upstream today without either fragile README parsing or upstream adoption of structured metadata files, which lines up with what you mentioned about maybe proposing metadata.yaml files to js-contrib down the line.

Given that, my rough take is telemetry parity shouldn't block getting JS onto the Explorer at all. I'd like to work on closing the registry-to-frontend gap, roughly in this shape, though I'm open to a different breakdown if you see it differently:

- Watcher hardening - instrumentation_sync.py, repository_manager.py, and main.py are currently the only untested parts of the watcher.
- README extraction into content-addressed storage, the same pattern used for collector's READMEs (#775  /#785 /#819 ), so packages can get a rendered README.
- A JS-specific path through explorer-db-builder and a browse/detail feature on the frontend, following the same pattern collector and java-agent already use rather than a shared abstraction. I can shape the schema around whatever comes out of the Phase 2 schema-design work too.

[...]

PR #1097, opened 2026-09-11
test(js-watcher): cover remaining package_parser branches

MeloveGupta:
## What changed

I have added 9 tests to the existing ```test_package_parser.py```. ```package_parser.py``` already had tests but it still had 23 uncovered statements.

Most of it is the ```.tav.yml``` parsing. The main gap was "Structure 2", where a package maps to a dict with a jobs key instead of a list. That branch runs every night and had nothing on it, it's what produces the exclude and mode fields that are already in the registry for aws-sdk, ioredis, knex, mongodb, mysql2, redis and tedious. The exclude field wasn't covered either. The rest are the failure paths, malformed ```.tav.yml```, unreadable files, yaml that isn't a mapping, and READMEs with no Supported Versions section.

## Why

Part of #990 

I said in #1066 that was the last one for watcher coverage. I meant the three untested modules and those are done, but ```package_parser.py``` was still sitting at 76% so the watcher wasn't actually finished. This takes ```package_parser.py``` from 76% to 100% and the watcher from 89% to 98%.

## How it was tested

I ran `uv run pytest tests/ --cov=js_instrumentation_watcher --cov-report=term-missing` from ecosystem-automation/js-instrumentation-watcher

Screenshot from 2026-09-11 17-18-40

I also ran `uv run pytest ecosystem-automation -q` for the whole Python suite too...

## Is this a breaking change?

No

NA

## Checklist

### DIFF
--- a/ecosystem-automation/js-instrumentation-watcher/tests/test_package_parser.py
+++ b/ecosystem-automation/js-instrumentation-watcher/tests/test_package_parser.py
@@ -249,3 +249,256 @@ def test_not_in_bundle_when_absent(tmp_package):
 
     assert result is not None
     assert result["in_auto_instrumentations_node"] is False
+
+
+def test_supported_versions_empty_when_readme_has_no_section(tmp_package):
+    write_package_json(
+        tmp_package,
+        {
+            "name": "@opentelemetry/instrumentation-express",
+            "version": "0.66.0",
+            "description": "test",
+        },
+    )
+    # A README with no "Supported Versions" heading at all - 39/47 packages
+    # in the May audit were prose-only like this.
+    (tmp_package / "README.md").write_text("# Express Instrumentation\n\nSome prose.\n")
+
+    parser = PackageParser(
+        package_path=tmp_package,
+        bundle_membership=set(),
+        component_owners={},
+    )
+    result = parser.parse()
+
+    assert result is not None
+    assert result["supported_versions"] == []
+
+
+def test_supported_versions_empty_when_readme_unreadable(tmp_package):
+    write_package_json(
+        tmp_package,
+        {
+            "name": "@opentelemetry/instrumentation-express",
+            "version": "0.66.0",
+            "description": "test",
+        },
+    )
+    # A directory named README.md passes exists() but raises IsADirectoryError
+    # (an OSError) on read. The package must still parse rather than be dropped.
+    (tmp_package / "README.md").mkdir()
+
+    parser = PackageParser(
+        package_path=tmp_package,
+        bundle_membership=set(),
+        component_owners={},
+    )
+    result = parser.parse()
+
+    assert result is not None
+    assert result["supported_versions"] == []
+
+
+def test_tav_entry_includes_exclude(tmp_package):
+    write_package_json(
+        tmp_package,
+        {
+            "name": "@opentelemetry/instrumentation-aws-sdk",
+            "version": "0.77.0",
+            "description": "test",
+        },
+    )
+    tav = textwrap.dedent("""
+        "@aws-sdk/client-s3":
+          - versions:
+              include: "^3.6.1"
+              exclude: "3.529.0 || >=3.363.0 <=3.377.0"
+              mode: max-7
+            commands: npm test
+    """)
+    (tmp_package / ".tav.yml").write_text(tav)
+
+    parser = PackageParser(
+        package_path=tmp_package,
+        bundle_membership=set(),
+        component_owners={},
+    )
+    result = parser.parse()
+
+    assert result is not None
+    assert result["tested_versions"][0]["exclude"] == "3.529.0 || >=3.363.0 <=3.377.0"
+    assert result["tested_versions"][0]["mode"] == "max-7"
+
+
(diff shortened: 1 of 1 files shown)


ecosystem-automation/js-instrumentation-watcher/tests/test_package_parser.py
```diff
@@ -249,3 +249,256 @@ def test_not_in_bundle_when_absent(tmp_package):
...
+        package_path=tmp_package,
+        bundle_membership=set(),
+        component_owners={},
+    )
+    result = parser.parse()
+
+    assert result is not None
+    assert len(result["tested_versions"]) == 2
+    ranges = [entry["range"] for entry in result["tested_versions"]]
+    assert ranges == sorted(ranges)
```
jaydeluca, line 405: This assertion can't fail. The fixture lists `^3.587.0` before `^3.600.0`, which is already sorted order, so it passes even if `_parse_tav_yml` did no sorting at all:

```
fixture YAML order: ['^3.587.0', '^3.600.0']
sorted():           ['^3.587.0', '^3.600.0']
assertion passes even with NO sorting: True
```

Either flip the two jobs in the fixture so the YAML order is genuinely unsorted, or just drop the assertion — `test_tested_versions_sorted_deterministically` above already covers sorting non-vacuously, with an out-of-order fixture and a comment saying so.

The rest of this test (two entries parsed from a top-level `jobs` list, `exclude` present on the first) is the valuable part and is correct — this is only about the one line.

jaydeluca (commented):
Following the "validate against real upstream, not only fixtures" rule, I pulled all 31 `.tav.yml`
files from `opentelemetry-js-contrib@main` and classified every package-config shape against the
branches the parser has:

| Shape                             | Upstream occurrences | Covered by                                      |
|-----------------------------------|----------------------|-------------------------------------------------|
| list → `versions` mapping         | 47                   | pre-existing tests                              |
| dict → flat `versions` mapping    | 12                   | **new** `test_tav_flat_versions_on_dict_config` |
| dict → `jobs`                     | 6                    | **new** `test_tav_top_level_jobs_key`           |
| `versions:` as a plain **string** | 4                    | **nothing — silently dropped**                  |
| list → `jobs`                     | 0                    | **new** `test_tav_jobs_key_inside_list_entry`   |

That fourth row is the interesting one, and it's why I'd say the watcher still isn't quite finished
even at 100%.

### 1. `.tav.yml` `versions:` string shorthand is silently dropped (pre-existing, not from this PR)

`tav` accepts `versions` as either a mapping or a bare semver string. Both guards in
`_parse_tav_yml` are `isinstance(versions, dict) and versions` (`package_parser.py:228` and `:249`),
so the string form falls through with no log line and no entry. Four entries across three packages
upstream use it today, and the loss is visible in the committed registry:

| Package | Upstream | Registry today | Lost |
|---|---|---|---|
| `instrumentation-oracledb` | `versions: "6.7.0"`, `versions: ">=6.10.0 <8"` | `tested_versions: []` | **everything** |
| `instrumentation-socket.io` | 3 entries, one is `versions: "^3.1.2"` | 2 entries | `^3.1.2` |
| `instrumentation-nestjs-core` | 7 entries, one is `versions: "6.11.11"` | 5 entries | `6.11.11` |

(`nestjs-core` is also missing `>=12.0.0 <13`, but that's ordinary version lag — the snapshot is
0.68.0 and upstream added NestJS 12 later. Not a bug.)

Running the parser directly on the real upstream oracledb file:

```
tested_versions from REAL upstream oracledb .tav.yml: []
```

Nothing in the code or tests asserts this is deliberate, and `_build_tav_entry`'s docstring
describes `versions` only as a dict, so it reads as an unhandled shape. One extra branch feeding the
existing builder would cover it:

```python
if isinstance(versions, str) and versions:
    results.append({"package": pkg_name, "range": versions, "source": ".tav.yml"})
```

MeloveGupta:
@jaydeluca Sorry for the late reply on this, and thanks for going through all 31. I ran the same check on my end and got the same 47/12/6/4.

Youre right about the sort assertion. I pulled the sorting out of `_parse_tav_yml` completely and my line still passed, only `test_tested_versions_sorted_deterministically` failed. So I removed it out...

The jobs test comment said that shape is behind the exclude data for 7 packages, it's actually just aws-sdk and knex. The no-heading README test said 39/47, which was the telemetry audit number, it's really 5. And I added a note on the list -> jobs test that nothing upstream uses it.

The string shorthand I'm opening as a separate PR so this one stays test only, will link it here. I also ran old vs new parser on all 31 upstream files, 28 come out the same and the other 3 get exactly the 4 entries from your table.

One question on that I had was: right now sync skips versions already in the registry, so the oracledb, socket.io and nestjs-core files already committed won't get fixed until those packages release again. So what do you say? Do we leave it at it, or should I add a backfill mode to the JS watcher like collector and configuration have?

Also tried your check on the README side and it has the same problem. 10 packages have a Supported Versions heading but the regex misses the format, aws-sdk, tedious and langchain lose real ranges. I can pick that up as a follow up if you think it's worth it...

Issue #1088, opened 2026-09-09
[Python Instrumentation] Implement Python instrumentation watcher and initial registry


Adii-45:
## Proposed approach

Implement a Python instrumentation watcher following the existing watcher patterns and infrastructure in the repository!

The watcher should:

- Clone and inspect `opentelemetry-python-contrib`
- Discover Python instrumentation packages from the upstream repository
- Parse the relevant structured metadata from `pyproject.toml` and `package.py`
- Resolve package versions according to the package-oriented versioning model established in #1064
- Preserve the relevant upstream metadata needed by the Python registry
- Generate/update the corresponding entries under `ecosystem-registry/python/`
- Detect and report disagreements between `pyproject.toml` and `package.py` where applicable, so we can understand why such disagreements occur and whether they are actually possible
- Follow the repository's existing watcher, inventory-manager, sync, testing, and automation patterns

The implementation should establish a maintainable automated path:

`opentelemetry-python-contrib → Python watcher → Python registry`

## Scope

- Add the Python instrumentation watcher under `ecosystem-automation/`
- Reuse existing repository-manager and watcher infrastructure where appropriate
- Implement Python-specific scanning, parsing, version resolution, and registry inventory generation
- Add appropriate unit/fixture-based tests
- Add the initial Python registry population through the watcher/inventory flow
- Integrate the watcher with the repository's existing registry automation/nightly workflow
- Add failure reporting consistent with the existing watcher automation
- Update relevant registry structure documentation where required

## Testing

The watcher should follow the existing repository testing conventions, including fixture/synthetic metadata tests where appropriate.

Tests should cover at minimum:

- package discovery
- metadata parsing
- version resolution
- registry entry generation
- relevant dependency/package metadata
- metadata-source disagreement detection/reporting
- expected handling of packages with different release/versioning behavior
- repository manager and sync behavior using the existing watcher test patterns

Tests should avoid relying on live network calls where the repository's existing watcher conventions provide a fixture-based alternative..

## CI / automation

The watcher should be integrated with the existing registry automation patterns, including:

- nightly registry updates
- manual workflow dispatch where appropriate
- failure notification
- registry path configuration
- watcher command registration

The implementation should follow the existing reusable registry-sync workflow rather than introducing a separate automation mechanism!

Part of #1029

Adii-45:
I’m planning to implement this phase in three focused PRs:

[...]

Issue #1064, opened 2026-09-05
[Python Instrumentation] Define registry schema and versioning model


Adii-45:
## Context

This is the second phase of the Python instrumentation effort tracked in #1029!

The metadata audit confirmed that `opentelemetry-python-contrib` provides enough structured metadata to support a dedicated Python instrumentation registry.. The next step is defining how that metadata should be represented before implementing the watcher and downstream integration!

## Goal

Define the registry schema and versioning model for Python instrumentation

This issue focuses on documenting:

- registry directory/file structure
- YAML schema and field semantics
- package versioning strategy
- handling lockstep vs independently-versioned packages
- metadata sources and source-of-truth decisions

## Scope

This is **documentation/design only** ~

**Included**
- Registry layout and schema design
- Versioning model
- Field definitions and metadata sources
- Maintainer decisions and trade-offs

**Not included**
- Watcher or registry implementation
- DB builder or CI changes
- Frontend integration

## Acceptance Criteria

- [ ] Proposed registry structure and schema are documented
- [ ] Versioning approach for lockstep and independently-versioned packages is defined
- [ ] Metadata sources and source-of-truth decisions are documented
- [ ] Remaining design decisions are clearly identified

Part of #1029

Issue #1041, opened 2026-08-31
[Python Instrumentation] Add Python instrumentation metadata audit


Adii-45:
## Component / area

Python instrumentation ecosystem -`opentelemetry-python-contrib` metadata and Ecosystem Explorer integration

## Goal

Add a Phase 1 research artifact documenting how Python instrumentation metadata can be represented and integrated into the OpenTelemetry Ecosystem Explorer..

## Scope

The audit should:

- Document the available metadata in `pyproject.toml`, `package.py`, and generated `instrumentation/README.md`
- Identify the metadata fields relevant to the Explorer, including duplication and any gaps
- Document the current Python package versioning model and its implications for a future registry
- Follow the existing JavaScript ecosystem research structure and ecosystem-mapping process

## Expected outcome

A repository-native Python instrumentation metadata audit that provides a clear foundation for the subsequent schema and implementation work!

This issue is limited to research and documentation; registry, watcher, DB-builder, and frontend implementation will be handled separately..!

Part of #1029

PR #1099, opened 2026-09-11
feat(python): add instrumentation watcher

Adii-45:
## What changed

Implements the first part of #1088 by adding the Python instrumentation watcher and initial registry generation for the Ecosystem Explorer!

This establishes an automated path from `opentelemetry-python-contrib` to the Python instrumentation registry, following the existing watcher architecture used by other ecosystems!

### Changes

- Add a Python instrumentation watcher under `ecosystem-automation`
- Add repository management and package discovery for `opentelemetry-python-contrib`
- Parse instrumentation metadata from `pyproject.toml` and `package.py`
- Resolve package and release versions using the hybrid versioning model defined in #1064
- Generate per-package-version registry entries under `ecosystem-registry/python/`
- Preserve the relevant upstream metadata required by the Python registry schema
- Detect and report metadata-source disagreements where applicable
- Integrate with the repository's existing watcher and workspace infrastructure
- Add unit/fixture tests covering package discovery, metadata parsing, version resolution, registry generation, disagreement handling, and sync behavior

## Why

This is the implementation phase of #1088, following the metadata audit in #1041 and registry schema design in #1064! The watcher provides the automated path needed to keep Python instrumentation metadata synchronized with the Ecosystem Explorer registry...!

## How it was tested

- Full repository test suite passes
- Ruff and formatting checks pass
- Watcher-specific tests pass
- End-to-end smoke testing was performed against a synthetic `opentelemetry-python-contrib` repository

## Is this a breaking change?

No

This PR intentionally covers only the watcher and initial registry integration. CI/nightly automation, DB builder integration, and related documentation updates are left for the subsequent parts of #1088

### DIFF
--- a/ecosystem-automation/python-instrumentation-watcher/README.md
+++ b/ecosystem-automation/python-instrumentation-watcher/README.md
@@ -1,0 +1,117 @@
+# Python Instrumentation Watcher
+
+Automation tool for synchronizing OpenTelemetry Python instrumentation metadata to the ecosystem
+registry.
+
+The metadata source is the contrib repository:
+<https://github.com/open-telemetry/opentelemetry-python-contrib>
+
+Background and design rationale for this watcher live in
+[`projects/135-python-instrumentation/`](../../projects/135-python-instrumentation/), specifically
+the [metadata audit](../../projects/135-python-instrumentation/01-metadata-audit.md) and the
+[registry schema design](../../projects/135-python-instrumentation/02-schema-design.md).
+
+## Methodology
+
+On a scheduled basis, the tool clones (or pulls) `opentelemetry-python-contrib`, discovers every
+instrumentation package under `instrumentation/`, and records per-package metadata snapshots in the
+registry.
+
+Process:
+
+- Clone or pull `opentelemetry-python-contrib` (override the checkout with the
+  `PYTHON_CONTRIB_REPO_PATH` env var; the clone target dir is configurable with
+  `PYTHON_CONTRIB_REPOS_DIR`, default `tmp_repos`).
+- Discover packages under `instrumentation/opentelemetry-instrumentation-*` that ship a
+  `pyproject.toml`. The separate `instrumentation-genai/` area is explicitly out of scope (tracked
+  instead by [#154](https://github.com/open-telemetry/opentelemetry-ecosystem-explorer/issues/154)).
+- For each package, parse `pyproject.toml` (authoritative) and `package.py` (cross-check only):
+  - `pyproject.toml`'s `[project]` table for name, description, `requires-python`, `urls.Homepage`.
+  - `pyproject.toml`'s `[project.optional-dependencies]` `instruments` / `instruments-any` keys for
+    instrumented library/version-range pairs (source key preserved per entry).
+  - `pyproject.toml`'s `[project.entry-points.opentelemetry_instrumentor]` for auto-instrumentation
+    entry points.
+  - The package's own version: read directly from `[project].version` if static, otherwise resolved
+    from the file named by `[tool.hatch.version].path` (reading `__version__`) when the version is
+    `dynamic`.
+  - `package.py`'s `_instruments`, `_supports_metrics`, and `_semconv_status`, parsed with
+    `ast.literal_eval` rather than executed (`package.py` is untrusted upstream code).
+- Compare `package.py`'s `_instruments` against `pyproject.toml`'s `instruments`/`instruments-any`;
+  log and report (but do not block on) any disagreement. `pyproject.toml` remains authoritative for
+  the `instruments` field written to the registry.
+- For each package, skip it if its current version is already tracked; otherwise write a versioned
+  YAML snapshot.
+
+Like the JS watcher — and unlike the Java agent, which has a single release version covering all
+instrumentations — Python packages are resolved and stored at their own, independent version. This
+also holds for packages that happen to release in lockstep with others; see schema design §6 for why
+no "release group" field is recorded.
+
+## Registry layout
+
+The watcher maintains a per-package, per-version inventory under `ecosystem-registry/python/`:
+
+```text
+python/
+└── {package-name}/                 # e.g. opentelemetry-instrumentation-flask
+    └── v{version}.yaml             # e.g. v0.48b0.yaml
+```
+
+### File format
+
+**Example**: `python/opentelemetry-instrumentation-example/v0.48b0.yaml`
+
+```yaml
+description: OpenTelemetry instrumentation for the Example framework
+entry_points:
+  - name: example
+    value: opentelemetry.instrumentation.example:ExampleInstrumentor
+homepage: https://github.com/open-telemetry/opentelemetry-python-contrib/tree/main/instrumentation/opentelemetry-instrumentation-example
+instruments:
+  - library: example
+    source_key: instruments
+    version_range: ">=1.0,<3.0"
+name: opentelemetry-instrumentation-example
+repository: open-telemetry/opentelemetry-python-contrib
+requires_python: ">=3.9"
+semantic_convention_status: null
+source_path: instrumentation/opentelemetry-instrumentation-example
+supports_metrics: null
+version: 0.48b0
+```
+
+Output is deterministic: keys are sorted and the `instruments` / `entry_points` arrays are sorted by
+a stable key so upstream reordering does not churn the registry. `version_range` preserves the raw
+specifier text from `pyproject.toml`, not a renormalized round-trip.
+
+## Usage
+
+From the repository root:
+
+```bash
+uv run python-instrumentation-watcher
+```
+
+## Development
+
+From the repository root:
+
+```bash
+# Install dependencies
+uv sync
+
+# Run tests
+uv run pytest ecosystem-automation/python-instrumentation-watcher/tests
+
+# Run tests with coverage
+uv run pytest ecosystem-automation/python-instrumentation-watcher/tests --cov=python_instrumentation_watcher
+
+# Run the module
+uv run python -m python_instrumentation_watcher
+```
+
+## Adding Dependencies
+
+```bash
+uv add --package python-instrumentation-watcher <package-name>
+```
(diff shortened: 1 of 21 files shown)


jaydeluca (commented):
Atomic registry write, the `_Unresolved` sentinel and the `AnnAssign` handling all look good. Two
findings from the last round are still open — inline on `repository_manager.py` and
`instrumentation_sync.py`. Both reproduce against a real clone of python-contrib but not against the
fixtures.

Not in the diff, so no inline comment possible:

- `projects/135-python-instrumentation/NEXT-STEPS.md` — Phase 3 still "(Future)" with unchecked
  boxes (line 27), `last_updated: "2026-09-05"` (line 6), §7 decisions this PR answers unrecorded.

Minor: `_parse_instruments` and `has_metadata_disagreement()` both drop unparseable requirements, so
a malformed spec on either side can make the two sets agree by omission.

ecosystem-automation/python-instrumentation-watcher/src/python_instrumentation_watcher/repository_manager.py
```diff
@@ -0,0 +1,56 @@
...
+            return existing
+
+        repo_path = self.base_dir / REPO_NAME
+
+        if repo_path.exists():
+            logger.info("Pulling latest changes for python-contrib...")
+            self._pull_latest(repo_path)
+        else:
+            logger.info("Cloning opentelemetry-python-contrib...")
+            self._clone_repository(REPO_URL, repo_path)
```
jaydeluca, line 67: `setup()` only clones/pulls the default branch, where python-contrib's `version.py` only ever holds
the next `.dev` — release versions land on the tag. So 51 of 52 packages resolve to `0.66b0.dev`.
And since that string is stable all cycle, `version_exists()` freezes the registry at the first
nightly of the cycle and skips every later change.

Suggest checking out a release tag — `BaseRepositoryManager._checkout_version` already exists and is
unused here, and tags give clean versions (`v0.65b0` → `__version__ = "0.65b0"`). 65 `v*` tags, so
backfilling the initial population is cheap too.

Mapping `.dev` → `SNAPSHOT` instead isn't really available: `semantic_version` can't parse PEP 440
betas at all — `Version("0.65b0")` raises, not just the `.dev` one — so a Python SNAPSHOT couldn't
be constructed or compared with existing tooling. Worth settling in schema-design §7 #5.

Note python-contrib tags two release lines (`v1.16.0` alongside `v0.65b0`), so tag selection needs
care; `_resolve_version` reads `__version__` from the checkout, so you only pick the tag, not parse it.

ecosystem-automation/python-instrumentation-watcher/src/python_instrumentation_watcher/instrumentation_sync.py
```diff
@@ -0,0 +1,129 @@
...
+        - If the current version already exists in the registry, skip it
+        - Otherwise parse and write the metadata
+        - Separately, report (but do not block on) a pyproject.toml/package.py
+          `instruments` disagreement, per schema design §7 open decision #2
+
+        Returns:
+            Summary dict with counts of new, skipped, failed, and
+            metadata-disagreeing packages
+        """
+        summary: dict[str, Any] = {
```
jaydeluca, line 76: `has_unresolved_metadata()` has no caller outside tests, so the state the sentinel preserves never
reaches a reader: over real upstream this reports `disagreements: 1` while botocore, cassandra,
httpx and psycopg2 were skipped by the cross-check entirely.

Add an `unresolved` list here, append next to the `disagreements` check at line 104, and include the
count in the `logger.info` at line 121.

ecosystem-automation/python-instrumentation-watcher/src/python_instrumentation_watcher/package_parser.py
```diff
@@ -0,0 +1,524 @@
...
+                    # never does. That's a distinct state from "not assigned at all" —
+                    # collapsing it into the default `None` would let
+                    # has_metadata_disagreement() silently treat an unresolvable
+                    # value as an intentional empty list. Mark it unresolved instead.
+                    logger.warning(
+                        "Could not statically resolve %s in %s; treating as unresolved rather than absent",
+                        name,
+                        path,
+                    )
+                    result["instruments"][source_key] = _UNRESOLVED
```
jaydeluca: All 4 upstream packages defining `_instruments_any` hit this branch, so that half of the cross-check
is dead against real data. All four forms resolve statically:

```python
_instruments_any = (*_instruments_botocore, *_instruments_aiobotocore)  # botocore, httpx
_instruments_any = (_instruments_psycopg2, _instruments_psycopg2_binary)  # psycopg2, cassandra
```

`_iter_field_assignments` already walks every top-level assignment — build a `name → literal` map in
the same pass and resolve `Name` / `Starred(Name)` against it. Sentinel stays as the fallback.

ecosystem-automation/python-instrumentation-watcher/src/python_instrumentation_watcher/package_parser.py
```diff
@@ -0,0 +1,524 @@
...
+
+        Returns:
+            Dict with 'instruments' (dict mapping each of INSTRUMENTS_SOURCE_KEYS to
+            a _InstrumentValue — list[str] if resolved, None if not assigned at all,
+            or _UNRESOLVED if assigned but not statically resolvable),
+            'semantic_convention_status' (str | None), and 'supports_metrics'
+            (bool | None). All None if no package.py was found.
+        """
+        candidates = sorted(
+            p for p in self.package_path.rglob("package.py") if "tests" not in p.parts and "test" not in p.parts
```
jaydeluca: `rglob` yields absolute paths, so this matches ancestor directories too: with the checkout under any
segment named `test`/`tests`, every `package.py` is excluded silently. httpx under `…/test/contrib`
parses `semantic_convention_status: None` instead of `"migration"`.

Use `p.relative_to(self.package_path).parts`.

ecosystem-automation/python-instrumentation-watcher/README.md
```diff
@@ -0,0 +1,117 @@
...
+  - `pyproject.toml`'s `[project.optional-dependencies]` `instruments` / `instruments-any` keys for
+    instrumented library/version-range pairs (source key preserved per entry).
+  - `pyproject.toml`'s `[project.entry-points.opentelemetry_instrumentor]` for auto-instrumentation
+    entry points.
+  - The package's own version: read directly from `[project].version` if static, otherwise resolved
+    from the file named by `[tool.hatch.version].path` (reading `__version__`) when the version is
+    `dynamic`.
+  - `package.py`'s `_instruments`, `_supports_metrics`, and `_semconv_status`, parsed with
+    `ast.literal_eval` rather than executed (`package.py` is untrusted upstream code).
+- Compare `package.py`'s `_instruments` against `pyproject.toml`'s `instruments`/`instruments-any`;
```
jaydeluca: `has_metadata_disagreement()`'s docstring says it compares each field only against its own matching
key, never the combined set — this says the opposite. Line 37 also omits `_instruments_any` and says
`ast.literal_eval` where the code now uses `_safe_eval_container`.

ecosystem-automation/python-instrumentation-watcher/src/python_instrumentation_watcher/inventory_manager.py
```diff
@@ -0,0 +1,107 @@
...
+            data: Metadata dict to serialize as YAML
+        """
+        path = self._version_path(package_name, version)
+        path.parent.mkdir(parents=True, exist_ok=True)
+
+        fd, tmp_name = tempfile.mkstemp(dir=path.parent, prefix=f".{path.name}.", suffix=".tmp")
+        tmp_path = Path(tmp_name)
+        try:
+            with os.fdopen(fd, "w") as f:
+                yaml.dump(
```
jaydeluca: `_parse_package_py` stores `ast.literal_eval` output for `_semconv_status` / `_supports_metrics`
with no type check, unlike the instruments path's `isinstance` guard. With `yaml.dump` rather than
`safe_dump`, `_semconv_status = ("migration",)` writes `!!python/tuple`, which `yaml.safe_load`
downstream can't read.

`safe_dump` (as in `watcher_common/inventory_manager.py:193`) makes that a write-time error;
`isinstance(value, (str, bool))` makes it a warning.

ecosystem-automation/python-instrumentation-watcher/src/python_instrumentation_watcher/main.py
```diff
@@ -0,0 +1,58 @@
...
+    repo_path = repo_manager.setup()
+
+    inventory_manager = InventoryManager(registry_dir=REGISTRY_DIR)
+
+    sync = InstrumentationSync(
+        repo_path=repo_path,
+        inventory_manager=inventory_manager,
+    )
+
+    summary = sync.sync()
```
jaydeluca, line 63: `summary["failed"]` is never read, so `main()` exits 0 even if every package fails to parse
(`test_main.py:42` pins `assert main() is None`) and the nightly `notify` job would report success.
Same shape as the JS watcher, so not a regression — just cheaper to fix before the nightly lands.

ecosystem-automation/python-instrumentation-watcher/src/python_instrumentation_watcher/package_parser.py
```diff
@@ -0,0 +1,524 @@
...
+        """
+        Safely extract `_instruments`, `_instruments_any`, `_supports_metrics`, and
+        `_semconv_status` from a package.py file via `ast.literal_eval` (plus the
+        narrow empty-container-call allowlist in `_safe_eval_container`). The file is
+        never executed. Handles both plain and annotated assignments.
+        """
+        result = self._empty_package_py_result()
+
+        try:
+            tree = ast.parse(path.read_text())
```
jaydeluca: Pin `encoding="utf-8"` — also line 363 and `inventory_manager.py:81` (`os.fdopen(fd, "w")`).

ecosystem-automation/python-instrumentation-watcher/src/python_instrumentation_watcher/package_parser.py
Adii-45: Addressed in the latest commit!

Adii-45:
@jaydeluca I’ve addressed the review feedback from the latest round, including the repository/tag handling, unresolved metadata tracking, metadata disagreement handling, test-fixture coverage, and the related documentation updates!

The PR is ready for another look whenever you get a chance... Thanks!

Issue #1123, opened 2026-09-18
Semantic Convention Explorer


jaydeluca:
<placeholder, will flesh out more soon>

Now that we have the [semantic convention conformance](https://github.com/open-telemetry/semantic-conventions-conformance) project picking up speed, we should incorporate this data into the explorer.

We should add a new top level section of the website that provides insights like a high level heat map of some sort

Related to https://github.com/open-telemetry/semantic-conventions-conformance/issues/117 which is a similar initiative to build some of this functionality right alongside the project as well.

Some questions we'd want to be able to answer:

- Adoption progress for various semantic convention versions/domains across languages and instrumentation

Plan:

- We should create a watcher for the data in that project and ingest it into the registry
- Update the explorer-db-builder to cross reference components, and augment their existing metadata with conformance information
- For individual component views, we should add a new "Semantic conventions" tab and display information about their conformance reports
- We should come up with high level visualizations of aggregated data that can provide insights at a macro level to adoption and conformance

Adii-45:
This looks interesting! I’d definitely be down to explore the source project and how we can bring its data into the Explorer!

ThyTran1402:
Hi @jaydeluca,

I'm interested in working on this issue. I'm happy to help to improve, monitor, and make the data available in Explorer.

Thank you!

jaydeluca:
@Adii-45 @ThyTran1402 thank you both for your interest in working on this, I have several tasks I will create issues for in the next day or so that we can start splitting up. I'll tag you both when they are ready.

PR #1138, opened 2026-09-21
Add new semantic conventions ecosystem with timeline

jaydeluca:
Adding a new semantic conventions section of the site. Eventually we will start including conformance information from https://github.com/open-telemetry/semantic-conventions-conformance

For now, it will just provide a history of the semantic conventions themselves. I will create an issue to handle automating keeping this up to date, and expanding to handle the federated projects, but this is a starting point.

Related to #1123

image


vitorvasc (approved):
Awesome! 🤩

PR #1139, opened 2026-09-21
fix(collector): handle out-of-sync distribution versions on collector pages

### LINKED ISSUE #1116: Fix collector pages when distro versions are out of sync
It's not guaranteed that the core collector and contrib will release at the same time, and this often leads to a situation where if only one of them updates in the registry, when we try and show "latest" in the UI, we see a much smaller list of components, for example: 

Image

I'm not sure the best way for us to address this. Perhaps we don't list the latest version or component counts at all and just provide links that just add the filter for the component type.

This is also a problem on the list page

Image

For these two pages we need to figure out:

- What we should display by default
- How we want to provide a way for users to filter/navigate by version, accounting for the fact that the contents might have several different versioning strategies. For example, if we start pulling in components from other distributions aside from core/contrib

Also another feature we should address when touching these pages: when I go to the `/collector/components` page it gives me links to go by component type, but often i just want to get to the overall list without the type filtered, but we don't have a link for that yet. We should add another card on the page that allows me to go to the list with no filters applied

SurbhiAgarwal1:
## What changed

Updated the database builder to build an active catalog in `index.json` combining the components from the latest release of each distribution (`core` and `contrib`), and added per-distribution release metadata to `versions-index.json`. On the frontend, updated `/collector` to display per-distribution version info if out of sync, added an "All Components" navigation card, and updated `/collector/components` to default to the active catalog. The version dropdown is hidden when Distribution is "All" and filtered to relevant releases when a specific distribution is selected, while unversioned detail page links resolve to the component's own distribution's latest release.

## Why

Fixes #1116

When Core and Contrib do not release simultaneously, tagging a single global release version causes component counts on `/collector` and `/collector/components` to drop significantly, showing only whichever distribution published last.

## How it was tested

- **Database Builder Tests**: Updated `test_collector_builder.py` and `test_collector_database_writer.py` to verify that when Core releases ahead of Contrib, `index.json` includes components from both latest releases, and `versions-index.json` records per-distribution latest versions and available distributions per release. Ran `pytest` across all builder tests (455 passed).
- **Frontend Unit Tests**: Updated `collector-explore-landing.test.tsx` and `collector-components-page.test.tsx` to cover the "All Components" card, per-distribution version labels, conditional version dropdown visibility, and unpinned active catalog detail links. Ran `bun run test` (16 test files, 212 tests passed).
- **Typecheck & Linting**: Ran `bun run typecheck` (`tsc -b`, 0 errors) and `bun run lint` (`eslint .`, 0 errors).
- **Manual verification**: Verified landing page counts and out-of-sync version text, confirmed the version dropdown is hidden when Distribution is "All" and appears with distribution-specific releases when filtering by a distribution, and confirmed detail links resolve correctly without 404s.

## Is this a breaking change?

No

Hiding the version dropdown when Distribution is "All" implements the suggestion discussed in #1116 to prevent confusion around global release versions across independent distributions.

## Checklist


ecosystem-explorer/src/hooks/use-collector-data.ts
```diff
@@ -149,15 +149,17 @@ export function useCollectorComponents(version: string): DataState<IndexComponen
     let cancelled = false;
 
     async function loadData() {
-      if (!version) {
+      if (version === null) {
```
jaydeluca, line 152: `""`/`undefined` used to mean "stay idle" and now mean "load the active catalog", but the three callers still passing `""` to disable the hook weren't updated:

- `src/features/collector/collector-components-page.tsx:162`
- `src/v1/features/list/list-page.tsx:220`
- `src/v1/features/detail/detail-page.tsx:166` (not in this diff)

Each now fetches `index.json` in the deprecated view and discards it. Harmless today, but `deprecatedView ? "" : …` reads as if it still disables the hook. Pass `null` at those sites and keep `""` for the active catalog. The hook has no tests of its own — worth one covering all three inputs.

ecosystem-explorer/src/features/collector/collector-detail-page.tsx
```diff
@@ -80,7 +80,10 @@ export function CollectorDetailPage() {
   );
   const version = deprecatedView
     ? (deprecatedEntry?.last_version ?? "")
-    : rawVersion || versionData?.versions.find((v) => v.is_latest)?.version || "";
+    : rawVersion ||
+      (distribution && versionData?.distributions?.[distribution]?.latest) ||
```
jaydeluca, line 84: Right fix, but the v1 twin didn't get it — `src/v1/features/detail/detail-page.tsx:161` still resolves through the global `is_latest`, so with core ahead of contrib, `/collector/components/contrib/<name>` with no `?version=` lands on core's release, whose manifest has no contrib entry. That's the bug this PR fixes, still live under `V1_REDESIGN`. You did update the v1 list page, so I assume this one was just missed.

(Commenting here since that file isn't in the diff.)

ecosystem-explorer/src/features/collector/collector-components-page.tsx
```diff
@@ -244,20 +266,51 @@ function CollectorComponentsContent({ urlVersion }: { urlVersion?: string }) {
...
-      search: `?${params.toString()}`,
+      search: params.size > 0 ? `?${params.toString()}` : "",
     };
   };
 
   const handleDistributionFilterChange = (newDistribution: string) => {
     const params = new URLSearchParams(searchParams);
     if (newDistribution === "all") {
       params.delete("distribution");
+      params.delete("version");
```
jaydeluca, line 287: Unconditional, so switching distribution to "All" drops the deprecated view — the `else` branch at line 297 guards `currentVersion !== "deprecated"` but this one doesn't. And since the dropdown is hidden at "All", there's no control left to get back; only the landing page's Deprecated card or the Back button. The same guard here would keep the branches symmetric.

jaydeluca (commented):
Four things I'd like addressed, all inline below:

1. **The `useCollectorComponents` sentinel change left three call sites behind.** `""` now means "load the active
   catalog", but all three callers still pass it to *disable* the hook.
  2. **`src/v1/features/detail/detail-page.tsx:161` didn't get the same fix as its legacy twin** — still resolves an
     unpinned version through the global `is_latest`, so it reproduces the bug this PR fixes. Not in the diff, so I
     couldn't comment inline.
3. **Switching distribution to "All" silently drops the deprecated view**, and the dropdown is hidden at "All", so
   there's no control left to get back.
4. **The new landing-page test pins an ordering the builder doesn't emit** — it passes only because of the mock's key
   order.

Also:

- **Needs a rebase** — content conflict in `ecosystem-explorer/src/v1/features/list/list-page.tsx` against #1083. Worth
  doing first, since that file also holds finding 1's unfixed sentinel.
- `collector_builder.py:291` — `for d in sorted(DISTRIBUTIONS)` is redundant given the `sort(key=...)` two lines below.
- `collector-components-page.tsx:241-254` — the `handleVersionChange` restructuring is behaviour-neutral and duplicates
  the `navigate` call.
- The "All Components" card duplicates ~25 lines of the `COMPONENT_TYPES` card markup rather than prepending an entry to
  that list.
- `src/v1/features/list/list-page.tsx:210-218` — `allVersions` returns `undefined` whenever
  `filters.distributions.length !== 1`, and `facet-panel.tsx:225` hides the whole facet on `undefined`. So the version
  facet and its "deprecated" option vanish in the default view, and with two distributions selected where core and
  contrib still share releases. Recoverable via the filter chip, so your call whether it's worth a follow-up.

ecosystem-explorer/src/features/collector/components/collector-explore-landing.test.tsx
```diff
@@ -247,11 +251,40 @@ describe("CollectorExploreLanding", () => {
...
+      error: null,
+    });
+
+    render(
+      <MemoryRouter>
+        <CollectorExploreLanding />
+      </MemoryRouter>
+    );
+
+    expect(screen.getByText("Core: v0.151.0 • Contrib: v0.150.0")).toBeInTheDocument();
```
jaydeluca, line 288: Passes only because the mock lists `core` first. `collector_database_writer.py:294` writes `sorted(distribution_latest.items())`, so real data is `{"contrib": …, "core": …}` and `distributionVersionsText` — iterating `Object.keys` — renders `Contrib: v0.160.0 • Core: v0.161.0`. That also disagrees with the distribution cards below, which use the file's own `DISTRIBUTIONS` const. Iterating that const instead fixes the order, puts core first as in #1116, and makes this assertion mean something.

ecosystem-explorer/src/features/collector/collector-components-page.tsx
```diff
@@ -183,6 +183,25 @@ function CollectorComponentsContent({ urlVersion }: { urlVersion?: string }) {
     return Array.from(set).sort();
   }, [components]);
 
+  const availableVersions = useMemo(() => {
+    if (!versionData?.versions) return [];
+    if (distributionFilter === "all") return versionData.versions;
+    return versionData.versions.filter(
+      (v) => !v.distributions || v.distributions.includes(distributionFilter)
+    );
+  }, [versionData, distributionFilter]);
```
jaydeluca, line 192: Minor: with a distribution selected and no path version the data is the active catalog, but this resolves the select to a concrete version — so the control claims a pin that isn't there, and selecting that same option fires no change event. An explicit `value=""` "Latest" option would match the "Latest (Active)" idea from #1116.

Also, the `useCollectorVersions` mock in the test file has no `distributions` block, so this branch and `availableVersions`' filtering are uncovered — the two new dropdown-visibility tests pass regardless of the new fields.

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/collector_builder.py
```diff
@@ -305,7 +321,7 @@ def run_collector_builder(
                 raise ValueError(f"audit report path {report_path} must be outside the database directory {db_dir}")
 
             # Latest release only: that's the version fixable upstream today.
```
jaydeluca, line 323: This comment and the `processed_versions[0]` stamp below it are both stale — `missing` now spans distributions sitting at different versions, so the report's `version` field mislabels anything not from the leading release.

ecosystem-automation/explorer-db-builder/src/explorer_db_builder/collector_database_writer.py
```diff
@@ -260,6 +266,10 @@ def write_version_list(self, versions: list[Version], bundle_hashes: dict[Versio
                 the frontend uses to fetch the single per-version bundle. The
                 field is omitted for versions without a hash so old clients and
                 missing bundles degrade gracefully to the per-component fan-out.
+            version_distributions: Optional map of version to the list of distributions
+                it contains.
+            distribution_latest: Optional map of distribution name to its latest
```
jaydeluca, line 271: Outside the hunks but related: `write_index` (line 305) still takes `latest_components`, documented as "from the latest release version" — it now receives a cross-version active catalog. Worth renaming the parameter and updating the docstring.

